Issue #16, opened 2019-06-13
Add web3-react support


crisgarner:
Web3-React allows the integration of multiple providers, developers can configure the connectors they want to support. for example portis, fortmatic, trezor or ledger

PR #18, opened 2019-06-19
Add Fortmatic

pedrouid:


crisgarner:
removes the need for web3-react of issue #16

PR #24, opened 2019-06-21
Expose core logic separate from UI

pedrouid:


pedrouid:
Well technically it would allow them to choose between what do they wanna use.
1. Button + Modal + Logic
2. Modal + Logic
3. Logic

PR #27, opened 2019-06-25
Adding Squarelink into the mix (squarelink@0.3.9)

theAlexPatin:
Squarelink is a new non-custodial, browser-agnostic, web3 provider. It offers enhanced web2-style UX (including new non-custodial private key recovery methods) and doesn't compromise on any security features. Integrations are popping up daily, so it should be included in the mix with web3connect.


crisgarner:
I see two issues. the first one is when a user closes the squarelink window it crashes the app.
 
Screenshot from 2019-06-25 20-43-57

The second issue is that once you hit confirm and connects to the dapp, nothing happens. I see that I get an object but it doesn't trigger:

```js
   sqlk.getProvider().enable(() => {
          this.onConnect(sqlk.getProvider());
    });
```

theAlexPatin:
Thanks Chris, I'll take a look at this in the morning!

theAlexPatin:
@crisgarner Those issues are fixed now!

PR #34, opened 2019-07-05
select network param for Fortmatic

crisgarner:
you can now provide an optional parameter for Fortmatic which allows developers to select the network if no param is provided mainnet is selected

### DIFF
--- a/src/core/connectors/fortmatic.ts
+++ b/src/core/connectors/fortmatic.ts
@@ -3,13 +3,15 @@ import Fortmatic from "fortmatic";
 
 export interface IFortmaticConnectorOptions {
   key: string;
+  network: string;
 }
 
 const ConnectToFortmatic = async (opts: IFortmaticConnectorOptions) => {
   if (opts && opts.key) {
     try {
       const key = opts.key;
-      const fm = new Fortmatic(key);
+      const fortmaticNetwork = opts.network ? opts.network : "mainnet";
+      const fm = new Fortmatic(key, opts.network, fortmaticNetwork);
       const provider = await fm.getProvider();
       await fm.user.login();
       const isLoggedIn = await fm.user.isLoggedIn();
--- a/README.md
+++ b/README.md
@@ -172,6 +172,7 @@ interface IInjectedProvidersMap {
   - Fortmatic:
 
     - key: the secret key (required)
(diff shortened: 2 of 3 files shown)


src/core/connectors/fortmatic.ts
```diff
@@ -3,13 +3,15 @@ import Fortmatic from "fortmatic";
...
   key: string;
+  network: string;
 }
 
 const ConnectToFortmatic = async (opts: IFortmaticConnectorOptions) => {
   if (opts && opts.key) {
     try {
       const key = opts.key;
-      const fm = new Fortmatic(key);
+      const fortmaticNetwork = opts.network ? opts.network : "mainnet";
```
pedrouid: simpler this way
```
const fortmaticNetwork = opts.network || "mainnet";
```

pedrouid:
Also we need to look into our prettier configs because this PR was supposed to be like 3-4 lines and its over 70

crisgarner:
> Also we need to look into our prettier configs because this PR was supposed to be like 3-4 lines and its over 70

Yes, I noticed that. I'm pretty much just using the default Prettier configuration for VS Code. Maybe you can you send me your config to add it.

pedrouid:
I have this setting on my VS Code that adapts the Prettier rules to the project's Eslint rules
```json
// settings.json
{
    ...
    "prettier.eslintIntegration": true,
}
```
This is a Typescript project however so it uses Tslint instead. Do you know if it's possible to add `.prettierrc` config file that simply follows the Eslint or Tslint rules instead of the default?

pedrouid:
Hmm it doesn't seem like it's possible according to this issue.

https://github.com/prettier/prettier-vscode/issues/292

I will investigate a bit further

pedrouid:
It looks like the solution might be to use `tslint-config-prettier`:
https://github.com/prettier/tslint-config-prettier

pedrouid:
Just tested a prettier configuration to fix this @crisgarner, could you test the `prettier-config` branch on your side?

pedrouid:
Fixed on commit [fdcdc69](https://github.com/web3connect/web3connect/commit/fdcdc6996af2bfaa53c46ce54b96ae21d8fe01d3)

Issue #39, opened 2019-07-08
Provide a way to configure not to show the WalletConnect provider button


dkent600:
There seems to be no way to configure not to show the WalletConnect provider button.  Can we have this?

Come to think of it, I don't see a way to avoid seeing injected providers either.

crisgarner:
I feel like as a default it should work with wallet connect and injected providers. We could add props for not showing these, but can't find a proper reason. I mean these options are the ones that most users have?

dkent600:
Suppose I only want injected plus Portis? Or Webconnect plus Fortmatic?

PR #41, opened 2019-07-12
Disable defaults

crisgarner:
Some developers may need to not show Injected providers or WalletConnect by Default (#39). By default, those are going to show but I've added props that if included it will not show the Injected Provider or WalletConnect


pedrouid:
Why is this PR showing changes from #35? I thought that was already merged.

pedrouid:
Nevermind just fixed it

PR #45, opened 2019-07-18
Add support for iframe based providers

moodysalem:
Adds support for ethvault.xyz

Fixes https://github.com/web3connect/web3connect/issues/43#issue-467760949

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -926,6 +926,14 @@
         "@ethersproject/errors": ">5.0.0-beta.0"
       }
     },
+    "@ethvault/iframe-provider": {
+      "version": "0.1.6",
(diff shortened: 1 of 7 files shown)


moodysalem:
I was surprised the export for this is a react component, I would suggest the interface for this module being 1 method  `getProvider(reload: boolean = false): Promise<EthereumProvider>` (where reload is whether the user should be presented the option to pick again) and using a much lighter component library than React to render the UI elements. Something that compiles to web components like https://lit-element.polymer-project.org/ would be really great. And the modal can be styled with CSS.

Really like the idea of this project. Need to make it an easy decision for developers to choose how they connect to Ethereum, and it's great if they can just drop something in to support all the providers. Thanks!

PR #52, opened 2019-08-02
Add support for Gnosis Safe extension

dkent600:
Adds support for [Gnosis Safe extension](https://safe.gnosis.io/)


crisgarner:
Just tested with Gnosis Safe, it works but it shows the default provider instead of the Safe logo. Is this happening only to me?

dkent600:
@crisgarner I'm not sure what you mean by "default provider"?  You mean MetaMask?  If you have more than one extension active I think it is rather random which is displayed.  If it shows MetaMask then that should actually work for Safe too.

If you are running this PR with only the Safe extension active, then you should see a Safe icon.

PR #71, opened 2019-08-27
Bundle size and build time optimizations

sergejmueller:
**Before optimizations**
* Bundle size: 3.91 MiB
* Build time: 1m17.612s

**After optimizations**
* Bundle size: 440 KiB
* Build time: 0m8.434s

**Changes**
* Add webpack-external-react
* Add vendors to externals
* Add vendors to noParse
* Optimize assets


pedrouid:
Hi @sergejmueller, I just tested your PR changes and I was able to build the bundle and use it in a plain HTML website and I got the following error:

Screenshot 2019-08-28 12 17 03

You can test the plain HTML example: https://github.com/pedrouid/test-web3connect

I also ran the example we provide on this repo which has a React environment but I didn't see neither the button or the modal getting rendered 

Screenshot 2019-08-28 12 17 53

You can test this by running `npm run start` from the repo's root directory

sergejmueller:
web3connect is an npm package which is imported into an app. This is what it says in the Readme. 

My solution removes third-party modules by marking as external. These third-party modules do not have to be part of the package, because they are installed via `npm install`.

If web3connect becomes a standalone JS file (there is nothing in the Readme), all third-party modules have to be included in the package (js file). But this is a completely different scenario. I would store the third-party modules in a separate chunk (e.g. vendors.js) to be load the external modules in an single JS file. You want me to adjust my code for this?

pedrouid:
True, web3connect is an npm package that is usually imported into an app but it also works as a standalone JS since it needs to be supported for other frontend environments like Angular, Vue and plain HTML.

I'm unsure about the pros and cons of putting third-party modules in a separate chunk but if this would allow to reduce the bundle size and maintain the current compatibility than it would be ideal.

sergejmueller:
I split third-party modules into _vendor_ chunk and replace some PNG logos by SVG.
But that doesn't make sense because third-party scripts are over 3 MB.

`index.js` 145 KiB
`vendors.js` 3.39 MiB

Sorry, there's not much I can do. The app code design should be completely revised and optimized for performance (e.g. Lazy-loading components). There is no further optimizations with Webpack possible. It is a React issue.

pedrouid:
Thanks! This was very helpful. I'm going to keep this PR open but I will mark the bounty as completed!

PR #79, opened 2019-09-03
Use dependency injection

pedrouid:


src/helpers/utils.ts
```diff
@@ -1,6 +1,14 @@
 import providers, { fallbackProvider } from "../providers";
 import { IProviderInfo, IInjectedProvidersMap } from "./types";
 
+export function getProviderPackage(name: string) {
+  const ProviderPackage = require(name);
+  if (!ProviderPackage) {
+    throw new Error(`Package ${name} is not available`);
```
dkent600: Are you sure you want to throw an error here, and not just return null?

src/helpers/utils.ts
```diff
@@ -1,6 +1,14 @@
 import providers, { fallbackProvider } from "../providers";
 import { IProviderInfo, IInjectedProvidersMap } from "./types";
 
+export function getProviderPackage(name: string) {
```
dkent600: Maybe call this `importProviderPackage`, to be clear what it does and to avoid confusion with providers/index/providerPackages ?

src/helpers/utils.ts
pedrouid: The error here is intentional because there is two scenarios where you call getProviderPackage:
- when you initiate Web3Connect and you pass provider options
- when you call one of the individual connectors
In both scenarios, I want to notify the developer that intended functionality won't work without the required provider package

PR #82, opened 2019-09-16
Add Authereum provider

miguelmota:
Add [Authereum](https://authereum.org) web3 provider


src/providers/index.ts
```diff
@@ -48,6 +50,11 @@ export const providerPackages = {
     option: "portis",
     required: ["id"]
   },
+  authereum: {
+    name: "@authereum",
```
pedrouid: no need for `@`

pedrouid:
Looking good, something wrong with the build, going to check it out

src/providers/index.ts
miguelmota: whoops; fixed

miguelmota:
Regarding the build, looks like it's running out of memory. You can use `GENERATE_SOURCEMAP=false` for the build since generating sourcemaps is memory intensive. More info [here](https://github.com/facebook/create-react-app/issues/6193#issuecomment-453864524)

pedrouid:
BTW @miguelmota I just tested the example with Authereum and I got the following error:

```
Uncaught DOMException: Failed to read the 'localStorage' property from 'Window': Access is denied for this document.
```

pedrouid:
Also I tested the production build for the example locally too and disabling `GENERATE_SOURCEMAP` worked but I honestly never had to do this for any app I've built before and this is a very small one.

I also noticed that the app bundle size increased considerably with the `authereum` dependency from `799.35kb` (master branch) to `1.19mb` (authereum branch).

You should look into your dependencies and most importantly your webpack config because otherwise you will have to ask all apps to disable `GENERATE_SOURCEMAP` to integrate Authereum

miguelmota:
@pedrouid yeah you're right it's absurdly large. We'll work on removing dependencies to get it to a normal size. Agree that disabling sourcemaps shouldn't be required because of this. Thanks for trying it out and providing suggestions and I'll get back to you when it's improved

miguelmota:
Deployment looks good now with removed dependencies

pedrouid:
Nice 👍 However I just tested again and I still get the missing `localStorage` issue running locally in development.

miguelmota:
I'll test it out; what browser were you using? and were you using incognito mode?

miguelmota:
I wonder if you were using an old version because the latest version doesn't use localStorage at all. You should try pulling the latest changes and wiping node_modules before trying it again

pedrouid:
Hi @miguelmota, I'm happy to merge this PR but I keep getting this issue. I cloned again from authereum/web3connect repo and I got the same with localstorage. The version installed was 0.0.4-beta.8.

Screenshot 2019-09-27 14 44 53

Can you try it yourself on your machine? Fresh clone to try to replicate it.

miguelmota:
@pedrouid ah I see now, it must be because your browser has cookies and localStorage disabled in the settings. I'll try to figure out a way to work around that.

miguelmota:
Hey @pedrouid I'm curious to how you're able to interact with other providers that rely on localStorage (like Portis) with the recommended browser setting turned off

nausettings

We aren't able to maintain sessions or store the ephemeral keys if there's no way to store them in the browser by blocking all site data. The best we can do is to alert the user to enable site data to access the provider (like portis)

miguelmota:
@pedrouid the authereum provider should no longer crash like that and it displays an error toastr if persistent storage is not available which is required. Many other providers won't also work with 3rd party storage disabled which is default in brave

miguelmota:
Hey @pedrouid is there anything we can do to help get this merged? The error you were seeing no longer exists (tested on brave with shield enabled). Would really like to start promoting web3connect to dapps but first need it merged!

pedrouid:
Hey @miguelmota, I just resolved merge conflicts and run the example again but now Im getting a compile error for some reason.

Screenshot 2019-10-16 13 32 52

Any idea what this might be?

miguelmota:
@pedrouid thanks for reporting. On macOS some files seems to be lowercased in the dist directory although the source files are capitalized. Debugging issue..

crisgarner:
@miguelmota I'm trying to test the Authereum integration and I'm getting the following error:

Screenshot from 2019-10-21 23-50-53

miguelmota:
@crisgarner I filed an issue with the dependency provider causing the error https://github.com/blocknative/notify/issues/22

Might need to strip them out if they're not responsive soon enough

miguelmota:
@pedrouid @crisgarner thanks for the patience; the dependency issues have been resolved and the latest build is working

crisgarner:
Tested and merged @miguelmota :+1:

PR #84, opened 2019-09-17
Validate signature for contract-based accounts using ERC-1271

miguelmota:
This pull request checks adds code to check if the signer address is a contract, and if it is, then use the the [is-valid-signature](https://www.npmjs.com/package/is-valid-signature) library by @cwhinfrey to validate the signature using the [ERC-1271](https://github.com/ethereum/EIPs/issues/1271) method. ERC-1271 is adopted by 0x and recently kickback to check signatures are valid for contract-based-accounts.

### DIFF
--- a/example/src/App.tsx
+++ b/example/src/App.tsx
@@ -2,6 +2,7 @@ import * as React from 'react'
 import styled from 'styled-components'
 import Web3 from 'web3'
 import Web3Connect from 'web3connect'
+import validContractSignature from 'is-valid-signature'
 // @ts-ignore
 import WalletConnectProvider from '@walletconnect/web3-provider'
 import Portis from '@portis/web3'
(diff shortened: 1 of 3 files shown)


example/src/App.tsx
```diff
@@ -265,8 +266,18 @@ class App extends React.Component<any, any> {
       const result = await web3.eth.sign(hash, address)
 
       // verify signature
-      const signer = recoverPublicKey(result, hash)
-      const verified = signer.toLowerCase() === address.toLowerCase()
+      let signer = null
+      let verified = false
+      const bytecode = await web3.eth.getCode(address)
+      const isContract = !/^(0x)?0+$/.test(bytecode)
```
pedrouid: (nit) Isn't easier to just test if length is greater than 2?

miguelmota: Depending on version of web3, it'll return `0x`, `0x0`, or `0x00`

miguelmota:
Btw the hard coded gas limit of [21000](https://github.com/web3connect/web3connect/blob/c1f8d5c5221a11ad78bcfe7c1389a73413bf1e0d/example/src/helpers/utilities.ts#L180) in the example won't work for contract-based. MetaMask has the [same problem](https://github.com/MetaMask/metamask-extension/issues/7115). Suggestion is to estimate gas or add additional gas it the sender is a contract.

pedrouid:
Hi @miguelmota, I've added a separate PR for estimating gas limit but we need to parse the data payload for contract-based wallets, currently hard coded to 0x. Could you take a look at #89 to suggest how we handle that for Authereum for example?

PR #86, opened 2019-09-25
add github action for building project

davyvanroy:
This contains an initial build file that builds both the web3-connect library as the example. 

I added some extras for the build (like max_old_space, skip_preflight_check) until it was able to run the build successfully.

If github actions is not enabled yet for this repo, you can request beta access here:
https://github.com/features/actions


davyvanroy:
see #91

Issue #102, opened 2020-01-16
How to get the `torus` instance


leonprou:
When connecting with Torus, how to get the `torus` instance?
With Portis is `portis` instance is put on the `provider`, can the same be done with Torus?

pedrouid:
I have included a fix for this on #103, keep an eye on it

pedrouid:
This has since been fixed on beta.26. Both `portis` and `torus` instances are exposed in the provider as follows:

```javascript
// portis
provider.portis

// torus
provider.torus
```

PR #68, opened 2019-08-22
Updated webpack config to exclude dependencies

eswarasai:
### Description
- [x] Reduced bundle size to ~700KB

web3Connect

Fixes: #66

PR #103, opened 2020-01-16
v1.0.0-beta.26

pedrouid:
- deprecate individual connectors
- updated README for readability
- expose instantiated libraries
- cache preferred provider in localStorage
- update dependencies in example

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,48 +98,6 @@ web3Connect.on("close", () => {
 web3Connect.toggleModal(); // open modal on button click
 ```
 
-## Utils
-
-```typescript
-function checkInjectedProviders(): IInjectedProvidersMap;
-function getInjectedProviderName(): string | null;
-function getProviderInfoByName(name: string | null): IProviderInfo;
-function getProviderInfo(provider: any): IProviderInfo;
-function isMobile(): boolean;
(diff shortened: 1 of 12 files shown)


pedrouid:
Fix #68 
Fix #102

pedrouid:
the build error is probably related to some dependency upgrade on the example app

pedrouid:
We need to discuss the cache preferred provider cc @crisgarner 

Currently all versions until 1.0.0-beta.25 did NOT cache the user's preferred provider by default

So we should probably disable it by default to make it backwards compatible and document it for developers to enable themselves

crisgarner:
> We need to discuss the cache preferred provider cc @crisgarner
> 
> Currently all versions until 1.0.0-beta.25 did NOT cache the user's preferred provider by default
> 
> So we should probably disable it by default to make it backwards compatible and document it for developers to enable themselves

Sure, I don't think it will be a problem since a lot of devs have been asking for this, but it makes sense for people that just update libraries and don't read the docs.  Do we need a way to delete the session? It might be helpful to create some sort of sign-in / sign out. :thinking:

pedrouid:
Maybe we should cut a release with everything except the preferred provider stuff for now

pedrouid:
Replaced by #106

PR #110, opened 2020-01-27
add cacheProvider option

pedrouid:
### Cache Provider

By default is set to `false` and Web3Connect will always require the user to choose a provider option before triggering the onConnect event. However you can enable caching the last chosen provider. Example:

```javascript
const web3Connect = new Web3Connect.Core({ cacheProvider: true });
```

If you wish to reset the cached provider you can call the following method:

```typescript
web3Connect.clearCachedProvider();
```


README.md
```diff
@@ -327,9 +327,49 @@ interface IProviderCallback {
...
+<Web3Connect.Button cacheProvider />
+```
+
+#### Core Module
+
+```javascript
+const web3Connect = new Web3Connect.Core({ cacheProvider: true });
+```
+
+There are four options for this flag: `false`, `true`, `onclick` and `onload`. Here are the expected behaviors for each one of them:
```
jamesmorgan: Where is the cache stored, is it worth documenting this, at least at a high level? Ignore me if it is somewhere else.

README.md
```diff
@@ -327,9 +327,49 @@ interface IProviderCallback {
...
+
+- If `false`, it will always display the modal to choose a provider.
+- If `true` or `onclick`, it will connect to the last chosen provider on `toggleModal` or button click
+- If `onload`, it will connect to the last chosen provider on page load where Web3Connect is instantiated
+
+If you wish to reset the cached provider you can call the following method:
+
+```typescript
+web3Connect.clearCachedProvider();
+```
```
jamesmorgan: Looks good, I will also be utilising the `reset` feature as well 👍 

My thoughts in general are that the cache should be enabled by default, explicitly disabling it if you want to force selection each time. Reasons being that it is assumed better UX if the user not always prompted and it remembers your selection between new tabs.

README.md
pedrouid: It’s stored in localStorage, I will mention it in the README

README.md
pedrouid: The only reason I didn’t want to enable it by default is to not break backwards-compatibility

pedrouid:
clearCachedProvider is inacessible for React Button but it's not widely used so I will undocument it from README

PR #133, opened 2020-03-26
Torus update

### LINKED ISSUE #130: can't run torus against rinkeby
Using:
* [Torus](https://github.com/torusresearch/torus-embed) v1.2.3.
* WebModal v1.1.0

Given the documentation on the Torus website, I would expect the following to result in a web3 connection to rinkeby.  But what actually results is a connection to main.

```
const web3Modal = new Web3Modal({
  providerOptions: {
   network: "rinkeby",
   torus: {
     package: Torus,
     options: {
       network: {
         host: "rinkeby",
         chainId: 4,
         networkName: "rinkeby",
       },
     },
   }
}
);

web3Modal.toggleModal();
```

The code causing the error appears to be here:

https://github.com/Web3Modal/web3modal/blob/dcb435e8805eb7c587a664bc2ea94f68d5332c4a/src/providers/connectors/torus.ts#L56

`opts.network.nodeUrl`, which is `unassigned`, ends up overwriting `network.host` which contains the value "rinkeby", thus changing "rinkeby" to `unassigned`.

chaitanyapotti:
The following changes have been made in this PR:
- simplify torus network interface to be inline with torus documentation
- fix issue with provider matches
- clean up dependencies
- bump torus in example
- updated readme
- Fixed #130 

@pedrouid I'm not sure of the impact for changing the network interface of torus.

### DIFF
--- a/README.md
+++ b/README.md
@@ -214,13 +214,18 @@ const providerOptions = {
   torus: {
     package: Torus, // required
     options: {
-      enableLogging: false, // optional
-      buttonPosition: "bottom-left", // optional
-      buildEnv: "production", // optional
-      showTorusButton: true, // optional
-      enabledVerifiers: {
-        // optional
-        google: false // optional
+      config: {
+        enableLogging: false, // optional
+        buttonPosition: "bottom-left", // optional
+        buildEnv: "production", // optional
+        showTorusButton: true, // optional
+        enabledVerifiers: {
+          // optional
+          google: false // optional
+        }
+      },
+      network: {
+        host: "rinkeby"
       }
     }
   }
--- a/example/package-lock.json
+++ b/example/package-lock.json
@@ -5,11 +5,39 @@
   "requires": true,
   "dependencies": {
     "@babel/code-frame": {
-      "version": "7.5.5",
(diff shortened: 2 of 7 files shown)


pedrouid:
Going to take some time to review this before publishing but basically from scanning through it the only thing that I would change is to use the network string from Web3Modal instead of configuring it seperately from the Torus providerOptions

pedrouid:
Hey @chaitanyapotti, I've taken a slightly different approach to a few points and included this in master branch.

First is the network option, this is present on the Web3Modal and shared between all providers to make it clearer I've added the IAbstractConnectorOptions interface to all connectors and removed from the Torus providerOptions documentation.

Second is the filterMatches approach to get the last match for the provider info. I'm still flexible about it but worried about backwards-compatible since some integrations will be affected by this change. So I've added a booean to toggle it instead.

Let me know if this clear or any issues with merging conflicts.

chaitanyapotti:
@pedrouid 
Thanks for implementing the common IAbstractConnectorOptions. That would clear the confusion for the users.
I see that you've ported over the changes. Everything looks fine.
Please feel free to close this PR

pedrouid:
Great! I will publish these changes with v1.2.0

cc @chaitanyapotti @dkent600

PR #153, opened 2020-05-04
Added link to WalletConnect options

miohtama:


pedrouid:
Actually we should link all providers to their NPM libraries

Issue #150, opened 2020-04-24
Closing the dialog from web3Modal.connect() does not reject promise


mds1:
Steps to reproduce:

```javascript
import Web3Modal from 'web3modal';
import WalletConnectProvider from '@walletconnect/web3-provider';

const providerOptions = {
  walletconnect: {
    package: WalletConnectProvider,
    options: {
      infuraId: process.env.INFURA_ID,
    },
  },
};

const web3Modal = new Web3Modal({ providerOptions });

console.log('open'); // printed to console
const provider = await web3Modal.connect(); // if you close this dialog
console.log('close'); // then this is never reached
```

I think the problem is here, as you can see `reject` is never called on `CLOSE_EVENT` or `ERROR_EVENT ` https://github.com/Web3Modal/web3modal/blob/96c156858abb824a49eed2c0b4cbcb5302a21feb/src/core/index.tsx#L69-L73

pedrouid:
Fixed on [v1.6.0](https://github.com/Web3Modal/web3modal/releases/tag/1.6.0)

PR #155, opened 2020-05-05
Subscribe to close event on connect and connectTo

mds1:
I don't think #150 is completely fixed by v1.6.0, as it currently only rejects the promise on an error event. Closing the dialog by clicking elsewhere on the page does not reject the promise.

I have not tested these changes locally but believe they should work


pedrouid:
Done! Published as v1.6.1. Thanks for the PR! 🙌

mds1:
Awesome, just tested it out and it's working perfectly. Thanks for the quick release!

PR #175, opened 2020-07-21
WalletConnect provider supports RPC param

dmihal:
I've added support for the WalletConnect RPC parameter so that non-infura chains can be used, such as xDai.

Note that infuraId must be set to null to satisfy the required parameter.

Example usage:
```javascript
const providerOptions = {
  walletconnect: {
    package: WalletConnectProvider,
    options: {
      infuraId: null,
      chainId: 100,
      rpc: {
        100: 'https://xdai.poanetwork.dev',
      },
    }
  },
```


dmihal:
I was looking at an outdated version of the provider 🤦‍♂️

PR #205, opened 2020-11-15
Increase zIndex to 9999

TransmissionsDev:
Chakra-UI and many other libraries use a zIndex higher than 2 for many components, tooltips, etc, which block web3modal. 

Increasing zIndex will allow the modal to be used in more situations and not be blocked by other components.

### DIFF
--- a/example/src/components/Modal.tsx
+++ b/example/src/components/Modal.tsx
@@ -18,7 +18,7 @@ const SLightbox = styled.div<ILightboxStyleProps>`
   margin-left: -50vw;
   top: ${({ offset }) => (offset ? `-${offset}px` : 0)};
   left: 50%;
-  z-index: 2;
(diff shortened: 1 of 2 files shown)


crisgarner:
Wouldn't be cleaner if you change it on the css per project? you just need to add in your project: 

```css
.web3modal-modal-lightbox{      
   z-index: 9999 ;
}
```

TransmissionsDev:
@crisgarner Oh I was not aware that Web3Modal had some human readable id's I could easily modify like that. 

That does solve my issue and will use this as a solution in the meantime but I still stand by my PR because I think a zIndex of 2 is pretty low considering web3modal should always be the top layer of the page.

tomonari-t:
@transmissions11 I agree with you. Often the modal is hidden.

PR #212, opened 2020-12-08
Supports providers using legacy API

aggre:
Some injection providers, such as Opera wallet and Coinbase wallet, continue to use deprecated API: https://docs.metamask.io/guide/ethereum-provider.html#ethereum-enable-deprecated

I have added a fallback to support them.


pedrouid:
Thanks @aggre 🙌 this is really useful. Currently we are refactoring this codebase to split it into multiple packages and once that's complete I will include this fix with the v2.0.0

aggre:
Thank you for checking! I can fix anytime this if there is a problem :)

pedrouid:
This PR is too big unfortunately... sorry for leaving it stale for so long but we will need to handle this differently

PR #219, opened 2020-12-21
Monorepo Refactor

namesty:
This PR aims to provide a new structure for web3modal in the form of a monorepo, and allow for multiple framework support.

The new structure consists of:
* Core package with connectors and providers logic in vanilla Typescript.
* Angular implementation using Core package as dependency
* React package using Core package as dependency


namesty:
I am still doing some tweaking to allow for automated local testing and dependency hoisting with lerna @pedrouid

namesty:
The monorepo has been properly set up with Lerna. It's working fine and good for code review.

I made an angular example app to serve as a playground to test the package, in addition to the automated tests that should be developed for each package. My only pet peeve with the playground is that Angular writes metadata in the package.json, this metadata has to be deleted each time the playground is going to be re-run; this is a known issue: https://github.com/angular/angular/issues/33395

The playground app for Angular specifically is an extra, testing should be covered by automated testing.

PR #221, opened 2020-12-22
Add Liquality wallet

ilanolkies:
Adds Liquality injected provider

Screen Shot 2020-12-22 at 15 08 20

Ref: https://liquality.io/atomic-swap-wallet.html


pmprete:
this will also help our project as it uses liquality as well

YorkSvit (commented):
871932

PR #233, opened 2021-02-01
Remove window.web3

makoto:
Metamask recently stopped injecting [window.web3](https://docs.metamask.io/guide/provider-migration.html#table-of-contents) and some user seems to have a problem accessing our site when the error is thrown while most users see warning on their console.

Screenshot 2021-02-01 at 18 55 01


pedrouid:
question is if this is really the best approach because some other injected providers might still use it and this would break their compatibility

perhaps the solution is to provide better safe guards that check whether web3 is completely undefined or incomplete

makoto:
Good point. Of the provider I included, MEW wallet also accesses window.web3 so I raised github issue https://github.com/MyEtherWallet/MEWconnect-web-client/issues/40

VictorSB:
Would a better approach be detect when metamask is installed and the desired option to use window.ethereum on those cases?

sawirricardo:
Any info about this? Thank you

pedrouid:
@makoto is this still relevant?

PR #237, opened 2021-02-15
Add binance testnet

hshar7:
Important for testing. Binance Smart Chain already supported so its testnet should be too


lerog17 (approved):
Thank you.

PR #241, opened 2021-02-19
Add Asynchrounous provider package loading

CodingKitsune:
Hello everyone,
I have been using web3modal for some time, and I noticed that, because I always need to pass in the packages that I want to use, the modal and its providers tend to bloat my applications quite a bit, adding up to 3mb when almost of all the providers are configured and installed.
I like the library and I don't want to stop using it, so I created this PR.

The idea is that we should be able to load providers dynamically whenever a user clicks an option. This way we can keep the most common ones always in the same bundle, loading synchronously alongside with the web3modal package.
But for uncommon providers, we can delegate their downloading and loading until the user clicks the respective option.

We can achieve this with a "packageFactory" config function, which would allow users to resolve their modules however they want, regardless of bundlers.
This function would be async.


pedrouid:
I've been wanting to do this for... so...long!! 🙌  Let's just hardcode the package names and remove the need for injecting them at all

Join on Discord if you haven't to chat more! https://discord.gg/YGnSX9y

Striffly:
Is this feature now available? I can't find any commit that refers to it @pedrouid

PR #286, opened 2021-06-16
add dynamic import 

LeonmanRolls:
Hey, 

Ive added some changes to allow dynamic imports of the provider packages 

The consumer would use this feature as follows 

```
  providerOptions: {
    walletconnect: {
      package: () => import('@walletconnect/web3-provider'),
      packageFactory: true,
      options: {
        infuraId: INFURA_ID
      }
    },
```

When packageFactory is true package will be treated as a function that returns a promise for the provider package

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,5 @@ lib
 
 yarn.lock
 .eslintcache
+
+.idea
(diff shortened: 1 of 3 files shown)


LeonmanRolls:
Closing as an upgrade is already in the works that will solve this issue

PR #297, opened 2021-07-16
Add optimism-kovan, optimism, arbitrum-rinkeby, arbitrum networks

rahul-kothari:


rahul-kothari:
Hopefully it is okay to call optimism eth as "ETH" itself and same for arbitrum eth.
On their block explorers, they use "ETH" too.

ligi:
> On their block explorers, they use "ETH" too.

you mean as symbol right?

here it is used for "chain" - not sure what this is used for anyway. The symbol should not be "ETH"

rahul-kothari:
@ligi in this JSON, `chain` refers to the native currency of the network. For optimism, arbitrum, this is ETH.

ligi:
then IMHO it should not be ETH as it is not ETH (similar - derived - but not the very same)

PR #299, opened 2021-07-23
Compressed PNG logos without loss of quality (-69KB)

NorthHub:
The images are compressed without loss of quality. This will reduce the final package size by 69KB


vitorgamer58 (commented):
Looks nice

PR #300, opened 2021-07-26
Return error from the provider when it is triggered

jessgusclark:
Currently, if the provider throws an error when attempting to connect, this error is hidden and not passed back to the dapp. By returning this error, the developer can get the specific error and handle it or return something to the user. Right now, it fails silently. 

## steps to reproduce the silent error

0. Browser with Metamask that is disconnected to site/dapp.
1. attempt to connect
2. click on Metamask (or injected)
3. Metamask pops up asking for permission. Click "cancel"
4. Web3modal closes and no response to the user.

## why I am suggesting this change

I'm working on a hardware wallet implementation where the provider package could return multiple reasons for failure. I would like to pass the reason for the failure back to the Dapp so that I can help the user troubleshoot the reason for the connection issue.


holic:
Also seeing this! An error is logged to the console, but `undefined` is thrown rather than the RPC error.

image

johnfrades:
same issue here, 
image
im getting this error but when i catch it its undefined

jessgusclark:
Thank you!

PR #387, opened 2022-01-20
Remove nested array typo from walletlink required packages field

### LINKED ISSUE #386: WalletLink never renders
A typo prevents the built-in WalletLink provider from rendering.

Typo:
https://github.com/Web3Modal/web3modal/blob/master/src/providers/providers/index.ts#L156

Failed condition resulting in `shouldDisplayProvider == false`
https://github.com/Web3Modal/web3modal/blob/master/src/controllers/providers.ts#L108

diligentcodoor:
WalletLink does not render because required options are never recognized as present due to being nested within an extraneous array.

Closes #386

### DIFF
--- a/src/providers/providers/index.ts
+++ b/src/providers/providers/index.ts
@@ -153,6 +153,6 @@ export const WALLETLINK: IProviderInfo = {
   type: "qrcode",
   check: "isWalletLink",
   package: {
-    required: [["appName", "infuraId", "rpc"]]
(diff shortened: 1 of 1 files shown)


jkchuynh829:
Hello @pedrouid! This is currently a blocker for our project. Any chance we can expedite this review and deploy?

cc @jeasterman

pedrouid:
Replaced by #428

PR #398, opened 2022-01-31
add support for BlockWallet

ImanH:
adding support for BlockWallet as an injected provider. BlockWallet is a browser extension crypto wallet with a focus on privacy. https://blockwallet.io/


NODRAJ:
bump

PR #402, opened 2022-02-01
Sequence Wallet web and injected connector support

pkieltyka:
This PR adds support for Sequence Wallet (https://sequence.build), an Ethereum-compatible Web3 contract account wallet.

In addition to the PR, there is a demo-dapp made which uses this web3modal connector: https://github.com/0xsequence/demo-dapp-web3modal -- however in order to try it you will need to yarn link it until this PR is merged+published.

Here is a video demonstration of the integration as well: https://watch.screencastify.com/v/5bQMPPRklk1RNBbUb4vQ


pedrouid:
It seems like this is just an injected provider so you don't need to add a separate connector

pkieltyka:
thanks Pedro :D

PR #415, opened 2022-02-22
added MetaMask as optional Provider for new users to web3 (Fixes #385)

### LINKED ISSUE #385: Metamask does not display
If I don't have MetaMask installed, it's a little illogical to choose not to display it

worldlyjohn:
Problem:
Metamask is one of the most popular non-custodial wallets, but it only appears if a user already has it installed.  I wanted to offer it as an option to new users when first onboarding into web3.

There have been several others asking for this (see #385), however it proved a little more tricky than simply following the post on [how to add new providers](https://github.com/Web3Modal/web3modal/blob/master/docs/ADDING_PROVIDERS.md). The reason is that Metamask was already defined in the "injected" list, so after following the post users that have Metamask installed would now see two. 

This happens because the constructor for `ProviderController` adds the current injected provider into the providers list.  The call to `getInjectedProvider()` then calls `getProviderInfoFromChecksArray(checks)` and when the injected provider `check` matches `isMetaMask` (from providers), this is added twice -- once if it's injected and again as part of the available providers. This issue would also occur should another injected provider also be added to provider list with matching `check` string. 

This PR leaves the behavior unchanged if user already has metamask installed. For users that don't have metamask installed, it offers it as an option (should the developer decide to enable it like the rest of the available providers).

The current behavior when user clicks on metamask and doesn't have it installed is to open a new window to the metamask download page.


pedrouid:
Actually if you test this with a non-metamask injected provider you still get it to pop-up ... here is Safari with Tokenary injected provider present in window.ethereum

Screenshot 2022-03-22 at 19 30 19

I think all we need to do is add a check that specifically says 'metamask' should not be displayed if any injected provider is even present in window

pedrouid:
Replaced by #438

worldlyjohn:
hey @pedrouid , apologies for being late to this response.

> Actually if you test this with a non-metamask injected provider you still get it to pop-up ... here is Safari with Tokenary injected provider present in window.ethereum

But what if you test where the **_user does not have any_** injected provider (ie new to web3)?  That's what I was solving for.

danfinlay:
Hey @pedrouid you closed this but didn’t actually address it.

PR #437, opened 2022-03-23
feat: added a listener triggered when a provider selected

SeanZom:
Added an event trigger represents the period between a provider selected(clicked) and before the event `CONNECT_EVENT` triggered.

The providers like WalletConnect and Portis, take some extra time to load their own widgets for further actions. This added listener is handy for developers to add the loading indicator or other prompts to indicate provider initialization is ongoing in the background.

### Usage:

```typescript
web3Modal.on("onProviderSelect", providerId => {
    console.log(
        `the on provider select: ${providerId}`
    );
});
```

### DIFF
--- a/src/constants/events.ts
+++ b/src/constants/events.ts
@@ -1,3 +1,4 @@
 export const CONNECT_EVENT = "connect";
 export const ERROR_EVENT = "error";
 export const CLOSE_EVENT = "close";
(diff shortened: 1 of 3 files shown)


pedrouid:
That's a good idea, I will merge this and rename the event to `selected`

pedrouid:
Renamed on commit e0de286e47616cbb941c8f44330a89ce087cc267

PR #482, opened 2022-05-15
Added an example for Using with ethers.js.

aayushmahapatra:
- Added an example from [codesandbox](https://codesandbox.io/s/j43b10) for using with ethers.js. 
- This example will help the developers get started easily.


pedrouid:
Really nice example @aayushmahapatra 🙌

PR #568, opened 2022-09-04
Add keyboard navigation.

### LINKED ISSUE #521: UX Improvement: Close modal with escape key
The modal should be closable using the escape key.

### LINKED ISSUE #559: Add keyboard navigation to modal
Hi! At this moment user can not navigate through modal content with keyboard, which makes worse the experience of using the modal and whole project.

mikhin:
Fix #559, fix #521.

Hi! First of all I want to thank you for this package – it does really cool job.

My motivation to create this PR (quote from https://webaim.org/techniques/keyboard/):

> Keyboard accessibility is one of the most important aspects of web accessibility. Many users with motor disabilities rely on a keyboard. Some people have tremors which don't allow for fine muscle control. Others have little or no use of their hands, or no hands at all. In addition to traditional keyboards, some users may use modified keyboards or other hardware that mimics the functionality of a keyboard. Blind users also typically use a keyboard for navigation. Users without disabilities may use a keyboard for navigation because of preference or efficiency.

So, in this PR I have added these features:
- Navigate through providers via keyboard;
- Close modal via ESC.

If you will see something that could be better, please feel free to make a comment.

### DIFF
--- a/example/src/styles.ts
+++ b/example/src/styles.ts
@@ -107,20 +107,18 @@ export const globalStyle = `
   	-webkit-text-size-adjust: 100%;
     -webkit-overflow-scrolling: touch;
     -ms-text-size-adjust: 100%;
-    -webkit-text-size-adjust: 100%;  
+    -webkit-text-size-adjust: 100%;
   }
 
   button {
     border-style: none;
     line-height: 1em;
     background-image: none;
-    outline: 0;
     -webkit-box-shadow: none;
             box-shadow: none;
   }
 
   [tabindex] {
-    outline: none;
     width: 100%;
     height: 100%;
   }
@@ -153,10 +151,9 @@ export const globalStyle = `
 
   a {
     background-color: transparent;
-    -webkit-text-decoration-skip: objects;  
+    -webkit-text-decoration-skip: objects;
     text-decoration: none;
     color: inherit;
-    outline: none;
   }
 
   b,
--- a/src/components/Modal.tsx
+++ b/src/components/Modal.tsx
@@ -1,6 +1,7 @@
 import * as React from "react";
 import * as PropTypes from "prop-types";
 import styled from "styled-components";
+import ReactFocusLock from "react-focus-lock";
 
 import { Provider } from "./Provider";
 import {
@@ -156,6 +157,14 @@ export class Modal extends React.Component<IModalProps, IModalState> {
     ...INITIAL_STATE
   };
 
+  public componentWillMount() {
+    window.addEventListener('keydown', this.handleKeyDown);
+  }
+
+  public componentWillUnmount() {
+    window.removeEventListener('keydown', this.handleKeyDown);
+  }
+
   public componentDidUpdate(prevProps: IModalProps, prevState: IModalState) {
     if (prevState.show && !this.state.show) {
       this.props.resetState();
@@ -173,6 +182,12 @@ export class Modal extends React.Component<IModalProps, IModalState> {
     }
   }
 
+  handleKeyDown = (event: KeyboardEvent) => {
+    if (event.key === 'Escape') {
+      this.props.onClose();
+    }
+  };
+
   public render = () => {
     const { show, lightboxOffset } = this.state;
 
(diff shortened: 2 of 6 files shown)


src/components/Modal.tsx
```diff
@@ -188,25 +203,27 @@ export class Modal extends React.Component<IModalProps, IModalState> {
...
-                  name={provider.name}
-                  logo={provider.logo}
-                  description={provider.description}
-                  themeColors={themeColors}
-                  onClick={provider.onClick}
-                />
-              ) : null
-            )}
-          </SModalCard>
+            <ReactFocusLock>
```
mikhin, line 206: We need this feature here to prevent the focus from leaving the modal window.

src/components/Provider.tsx
```diff
@@ -34,12 +34,36 @@ interface IStyedThemeColorOptions {
   themeColors: ThemeColors;
 }
 
-const SName = styled.div<IStyedThemeColorOptions>`
+const SName = styled.button<IStyedThemeColorOptions>`
```
mikhin, line 37: We need to make one of card's elements really interactive to grant the ability to use it as focusable element.

example/src/styles.ts
```diff
@@ -107,20 +107,18 @@ export const globalStyle = `
...
     -ms-text-size-adjust: 100%;
-    -webkit-text-size-adjust: 100%;  
+    -webkit-text-size-adjust: 100%;
   }
 
   button {
     border-style: none;
     line-height: 1em;
     background-image: none;
-    outline: 0;
```
mikhin, line 117: Removed this to show keyboard navigation in the example

src/components/Modal.tsx
```diff
@@ -188,25 +203,27 @@ export class Modal extends React.Component<IModalProps, IModalState> {
...
+              <SModalCard
```
mikhin, line 207: Nothin was changed here, except the nesting levels

src/components/Provider.tsx
```diff
@@ -34,12 +34,36 @@ interface IStyedThemeColorOptions {
...
+
+  @media (hover: hover) {
+    &:hover, &:focus {
+      &::before {
+        background-color: ${({themeColors}) => themeColors.hover};
+      }
+    }
+  }
+
+  &::before {
```
mikhin, line 56: This is just a trick when we need to make clickable area larger than the element. So we make this pseudo element stretch to the full width of provider card.

src/components/Provider.tsx
```diff
@@ -71,20 +95,15 @@ const SProviderContainer = styled.div<IStyedThemeColorOptions>`
...
   width: 100%;
   padding: 8px;
   display: flex;
   justify-content: center;
   align-items: center;
   flex-direction: column;
-  cursor: pointer;
   border-radius: 0;
   border: ${({ themeColors }) => `1px solid ${themeColors.border}`};
-  @media (hover: hover) {
```
mikhin, line 83: Move this and the cursor rule to the pseudo of `SName` button.

chadyj:
Hey @mikhin Thanks so much for this PR!

Unfortunately it came just before a major refactor and we have since moved to v2.

Would you be interested in adding keyboard navigation to v2?

We have an issue here https://github.com/WalletConnect/web3modal/issues/672

mikhin:
> Hey @mikhin Thanks so much for this PR!
Hi, @chadyj! Thanks for your response. 

> Unfortunately it came just before a major refactor and we have since moved to v2.

OK, I see, no problem! I have a question – will the V1 be available in the future for downloading, even if the V2 is released? I mean, if yes, it seems it will be useful to merge this PR since a lot of sites are using `web3modal V1` for now. If not, it means that we do not need this PR, but it is OK.

> Would you be interested in adding keyboard navigation to v2?

Yes, I am interested, but for now, I have a 2 weeks vacation, and then I will have a project that will need much of my attention. So I will try to look at the V2 keyboard navigation issue, but can't promise that I will start it in the nearest time.

chadyj:
> OK, I see, no problem! I have a question – will the V1 be available in the future for downloading, even if the V2 is released?

Yes, it will continue to be available.  But keep in mind v1 will be sunset in 2023.
https://medium.com/walletconnect/walletconnect-v1-0-sunset-notice-and-migration-schedule-8af9d3720d2e

> So I will try to look at the V2 keyboard navigation issue, but can't promise that I will start it in the nearest time.

Awesome thank you! 🙏 

Thanks again for your contribution and understanding.

I'll close this for now, but feel free to re-open if you want to contribute again.

PR #578, opened 2022-09-16
Move projectId into a .env.local

crypblizz8:
This format follows similar formatting of `web-examples` [repo](https://github.com/WalletConnect/web-examples/blob/main/dapps/react-dapp-v2/README.md) 

- .env.local.sample created. 
- Follows better security pattern for env var exposure.

### DIFF
--- a/examples/react/.env.local.example
+++ b/examples/react/.env.local.example
@@ -1,0 +1,1 @@
+NEXT_PUBLIC_PROJECT_ID="my-project-id" // Get projectID at https://cloud.walletconnect.com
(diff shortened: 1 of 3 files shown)


IljaDaderko (approved):
Good addition 👌 I would not mention this as new step #2 in readme if possible, to keep it more general. Instead I think we can create one inside examples/react that can be more specific to next js / how to get started with example maybe?

IljaDaderko:
@crypblizz8 implemented it in one of my pr's, so closing this for now.

Issue #585, opened 2022-09-19
2.0.0 Enhancements


chadyj:
- [x] Better injected wallet management, add default wallet images and names
until dynamic packages are loaded)
- [x] Add artificial loader while explorer is opening when there is cached data (animation performance improvement)
- [x] Add hover effect for modal header buttons (need to discuss with Maciej)
- [x] Fix edge case and re-generate wallet connect code on cancel / expiry
- [x] ~Create custom jsonrpc providers for unsupported chains if they provide rpc url in config~ this is now handled on wagmi side
- [x] Add more horizontal space to wallet-button label
- [x] Add app details to config so it shows up nicer in wallets
- [x] Correct mobile re-direct handling for chain switching (why does it deep link after first time instead of universal?)
- [x] Check for issues with reconnection on mobile when safari is fully closed (I suspect we should delay rendering of app 
- [x] Look into react hooks / ethereal api implementation, check if they are unsubscribed correctly
- [x] ~Create `getWeb3Provider` helper and `useWeb3Provider` hook (https://github.com/WalletConnect/web3modal/issues/602)~
- [x] Add ethers as peerDependency
- [x] Fix "…" color in explorer screen
- [x] Move wage/core management to ethereal package, expose its options
- [x] Add buffer polyfill by default once we manage @wagmi/core
- [x] Make higher breakpoint for mobile
- [x] Add modal shadows
- [x] More mobile breakpoints (stretch + add spacing latter)
- [x] Dynamic spacing between wallet reel on mobile to ensure there is indication of scroll
- [x] Play with spring animations

tussiharel:
I want to help, what can I do?

PR #588, opened 2022-09-20
Styling fixes

IljaDaderko:
- Addresses some styling related issues in #585 (tackled tasks are ticked off)
- Added accessor function for ethereumClient (without it accessing its state resulted in outdated values)
- Updated deps
- Extended toast component to be used for error messages

### DIFF
--- a/packages/ui/src/components/w3m-modal-header/styles.ts
+++ b/packages/ui/src/components/w3m-modal-header/styles.ts
@@ -10,17 +10,20 @@ export default css`
     position: relative;
   }
 
-  .w3m-back-btn {
-    position: absolute;
-    left: 0;
+  .w3m-modal-header button {
     background-color: transparent;
     padding: 13px 17px;
-    margin-left: -17px;
     transition: opacity 0.2s ease-in-out;
   }
 
-  .w3m-back-btn:hover {
-    opacity: 0.6;
+  .w3m-modal-header button:hover {
+    opacity: 0.5;
+  }
+
+  .w3m-back-btn {
+    position: absolute;
+    left: 0;
+    margin-left: -17px;
   }
 `
 
--- a/packages/ui/src/partials/w3m-mobile-wallet-selection/styles.ts
+++ b/packages/ui/src/partials/w3m-mobile-wallet-selection/styles.ts
@@ -8,6 +8,7 @@ export default css`
     padding: 2px 0;
     display: flex;
     justify-content: space-between;
+    position: relative;
   }
 
   .w3m-view-row::-webkit-scrollbar {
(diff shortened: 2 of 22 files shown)


packages/ui/src/partials/w3m-mobile-wallet-selection/styles.ts
```diff
@@ -16,7 +17,7 @@ export default css`
 
   w3m-wallet-button,
   w3m-view-all-wallets-button {
-    padding: 0 10px;
+    padding: 0 calc((100% - 255px) / 11);
```
devceline, line 20: This seems a bit too "magic number"y. Anyway we can use percentages or add comments explaining the reasoning behind the values

packages/ui/src/components/w3m-modal-header/styles.ts
```diff
@@ -10,17 +10,20 @@ export default css`
...
-  .w3m-back-btn:hover {
-    opacity: 0.6;
+  .w3m-modal-header button:hover {
+    opacity: 0.5;
+  }
+
+  .w3m-back-btn {
+    position: absolute;
+    left: 0;
+    margin-left: -17px;
```
devceline, line 26: Replacing `-17px` with `-5%` results in a nearly identical look but it's a lot more clear and perhaps will be more responsive going forward.

devceline (approved):
LGTM

crypblizz8 (approved):
LGTM 🤝

Checked:
- Rework on ethereal client hooks
- Smoother transition on mobile
- Error toast works great 
- Updated packages.

Small Check:
- Want to double check `600px` is the breakpoint we set since it was `450px` before. Assumption is that `450` was too small and most mobile is around `576px` these days. 

Testing Process: 
- Read through code over PR 
- Test the edge cases locally
- Tested ErrorToast on `w3m-metamask-connector-view` / `w3m-walletconnect-connector-view`

packages/ui/src/partials/w3m-mobile-wallet-selection/styles.ts
IljaDaderko: You are right, I added a comment explaining what this does / is used for. Was not able to achieve it with just percentages since actual wallet icons / buttons have fixed widths, this this calculation.

packages/ui/src/components/w3m-modal-header/styles.ts
IljaDaderko: Most of ui elements in web3modal are fixed, including most paddings, so it is fine here. Main responsiveness in the modal comes from things like `display: flex` with `justify-content: space-between` for example.

devceline: Got it

IljaDaderko:
@crypblizz8 

> Want to double check 600px is the breakpoint we set since it was 450px before. Assumption is that 450 was too small and > most mobile is around 576px these days.

yep exactly, this is following Maciejs feedback when we tested on an iPhone Plus sized phone.

PR #589, opened 2022-09-20
Explorer searchbar

devceline:
# Changes
- Add debounce function
- Add logic and searchbar for filtering wallets
- Handle icon and minor styling


IljaDaderko:
- There is a "shift of content" flick when opening explorer for the first time
- It also looks like content is duplicated
- Going back and then to "view all" page results in "No Results" message
- Height and color of the input don't match figma
- State of searchbar when focused and with content doesn't match figma (there input expands and cancel button is shown)
- Slight .2s ease-in-out transition between normal and focus state would be good to match rest of ui
- There should be a placeholder

https://user-images.githubusercontent.com/3154053/191263602-a1f40b89-e6d6-472d-94c4-eb2721cde970.mov

IljaDaderko:
Also, seeing many weird states in mobile

- Getting stuck in loading some pages
- Double loaders / content dissapearing

https://user-images.githubusercontent.com/3154053/191264791-f55f0ae1-a518-4b0f-a462-e5471ab0f27e.mov

IljaDaderko:
- Loader should be positioned at the center of the screen when no other content is displayed
- Initial "View All" access shows "No Results" message and loader at the same time (No Results should also be in the middle of the screen)
- Looks like images are no longer being pre loaded? Before this, data was only shown on screen when it + associated images were loaded. There is utility for this in `UiHelpers`
- Search-bar needs to match figma designs (colors, hover states, focus states, text color, spacings etc...)

https://user-images.githubusercontent.com/3154053/191433865-da7e5c03-4c46-404a-8fc9-7370e47f4094.mov

IljaDaderko:
- On mobile, every time next page is loaded it jumps to the top for some reason, results in ux that is hard to control
- Mobile search bar has 2 icons

https://user-images.githubusercontent.com/3154053/191434928-822a7870-c5dc-4e6d-8ff5-799037cdeed5.MP4

IljaDaderko (changes requested):
Did further tests on desktop and mobile, left comments + vids

packages/core/src/controllers/ExplorerCtrl.ts
```diff
@@ -33,16 +35,30 @@ export const ExplorerCtrl = {
...
     state.wallets = {
-      listings: [...state.wallets.listings, ...listings],
+      listings,
       total
     }
-    const { page, search } = params
-    if (typeof page !== 'undefined' && state.page !== page) state.page = page
-    if (typeof search !== 'undefined' && state.search !== search) state.search = search
+
+    this.state.isLoading = false
```
IljaDaderko, line 61:
```suggestion
   state.isLoading = false
```

packages/core/src/utils/ExplorerApi.ts
```diff
@@ -10,7 +10,11 @@ export function getExplorerApi() {
 }
 
 function formatParams(params: PageParams) {
-  const stringParams = Object.fromEntries(Object.entries(params).map(([k, v]) => [k, v.toString()]))
+  const stringParams = Object.fromEntries(
+    Object.entries(params)
+      .filter(([_, v]) => typeof v !== 'undefined' && v !== null && v !== '')
```
IljaDaderko, line 15: v => value
k => key

Better to keep them as full names

packages/ui/src/utils/UiHelpers.ts
```diff
@@ -69,3 +69,18 @@ export async function preloadImage(src: string) {
 export function getErrorMessage(err: unknown) {
   return err instanceof Error ? err.message : 'Unknown Error'
 }
+
+/* eslint-disable-next-line */
```
IljaDaderko, line 73: Use eslint disable with specific rule name(s)

packages/ui/src/views/w3m-wallet-explorer-view/index.ts
```diff
@@ -13,26 +15,64 @@ import '../../components/w3m-modal-header'
...
 export class W3mWalletExplorerView extends LitElement {
   public static styles = [global, styles]
 
   // -- state & properties ------------------------------------------- //
-  @state() private loading = !ExplorerCtrl.state.wallets.listings.length
-  @state() private firstFetch = !ExplorerCtrl.state.wallets.listings.length
+  @state() private firstFetch = true
+  @state() private search = ''
   @state() private endReached = false
+  @state() private listingResponse: ListingResponse = { listings: [], total: 0 }
```
IljaDaderko, line 38: Why are these stored here as opposed to `ExplorerCtrl`? If you need them separate, let's just create new entry there like `searchResult`?

packages/ui/src/views/w3m-wallet-explorer-view/index.ts
```diff
@@ -89,34 +170,50 @@ export class W3mWalletExplorerView extends LitElement {
...
+          : html`
+              ${listings.length > 0
+                ? html`
+                    <div class="w3m-content">
+                      ${listings.map(
+                        listing => html`
+                          <w3m-wallet-button
+                            src=${listing.image_url.lg}
+                            name=${listing.name}
+                            .onClick=${async () =>
```
IljaDaderko, line 197: Make this a private function with descriptive name. Let's try to keep markup "lean" and logic separate

IljaDaderko (changes requested):

packages/ui/src/views/w3m-wallet-explorer-view/index.ts
devceline: Since we are calling `getPaginatedWallets` from 2 different places, I've opted to do this so we control when the listings are loaded into UI (eg, after preloading the images, etc)

IljaDaderko (approved):
Great work! Thank you for addressing all of the comments 🔥

PR #597, opened 2022-09-22
Search input component, simplified logic in "View All"

IljaDaderko:
- Made search bar a separate component, added placeholder animation
- Simplified logic in "view all" page
   - No longer subscribing to state
   - Search state is separate from main wallet state
   - loading / search state removed from explorer controller
   - search re-uses intersection observer logic for fetching


packages/ui/src/views/w3m-wallet-explorer-view/index.ts
```diff
@@ -173,49 +108,53 @@ export class W3mWalletExplorerView extends LitElement {
     else await this.onConnect(listing.desktop, listing.name)
   }
 
+  private readonly searchDebaunce = debounce((value: string) => {
```
devceline, line 111:
```suggestion
  private readonly searchDebounce = debounce((value: string) => {
```

packages/ui/src/views/w3m-wallet-explorer-view/index.ts
```diff
@@ -173,49 +108,53 @@ export class W3mWalletExplorerView extends LitElement {
...
-                : html`
-                    ${this.firstFetch
-                      ? null
-                      : html`
-                          <div class="w3m-centered-block">
-                            <w3m-text align="center" color="secondary">No results found</w3m-text>
-                          </div>
-                        `}
-                  `}
-            `}
```
devceline, line 218: Sorry if I'm missing something, but don't we need this?

packages/ui/src/views/w3m-wallet-explorer-view/index.ts
```diff
@@ -84,78 +54,43 @@ export class W3mWalletExplorerView extends LitElement {
 
   private createPaginationObserver() {
     this.intersectionObserver = new IntersectionObserver(([element]) => {
-      if (element.isIntersecting) this.fetchWalletsEndless()
+      if (element.isIntersecting && !(this.search && this.firstFetch)) this.fetchWallets()
```
devceline, line 57: What if, hypothetically, a search  term resulted in a over a page of results

packages/ui/src/views/w3m-wallet-explorer-view/index.ts
IljaDaderko: Noup, it is handled via https://github.com/WalletConnect/web3modal/pull/597/files#diff-3f421d6903c140850f3c5abc03fb4209e31026cd27ba8ea65e3a92a6da1cfdabR56-R58

packages/ui/src/views/w3m-wallet-explorer-view/index.ts
devceline: Oh awesome, got it.

devceline (approved):
LGTM

PR #599, opened 2022-09-22
docs: update text

finessevanes:
I'm pretty sure this should read `providers` and not `chains`.


IljaDaderko:
Yep, good catch! Could you please exclude package-lock from the diff.

IljaDaderko:
@finessevanes to speed this up, added the change in search pr here https://github.com/WalletConnect/web3modal/pull/597

PR #614, opened 2022-09-30
FIX: Metamask doesn't inject in Firefox under strict CSP

mikulas-mrva:
Fixing a known issue with Metamask in Firefox under strict Content-Security-Policy that prevents from injecting the provider js object. 

See https://github.com/MetaMask/metamask-extension/issues/3133

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,14 +1,16 @@
 {
   "name": "web3modal",
-  "version": "1.9.8",
+  "version": "1.9.9",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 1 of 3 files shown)


jmrossy:
In case this helps anyone else, just tested this solution with updated metamask libs and seems to still work :)
```
"@metamask/post-message-stream": "6.1.0",
"@metamask/providers": "10.2.1",
 ```

0xAsimetriq:
This looks like implementation that should be done on dapp level, not web3modal. Perhaps you could pr under web3modal FAQ page instead, explaining what needs to be done as a workaround for CSP?

PR #618, opened 2022-10-05
V2 alpha.5 - readme

IljaDaderko:
⚠️ Merge only when alpha.5 is released

Updated hooks readme to account for latest changes.


IljaDaderko:
Published this and will merge the readme, as we can fix typos etc.. latter on

PR #621, opened 2022-10-06
alpha.6

### LINKED ISSUE #600: MetaMask not connecting iOS browser
**Describe the bug**
Unable to connect to injected wallet MetaMask on iOS (possibly Android?) within the Metamask mobile browser. Clicking MM opens up App store rather than just connecting.

**SDK Version**
 - Version 2

**To Reproduce**
1. Load Metamask iOS App Browser
2. Go to 'web3modal.pages.dev'
3. Click on 'connect wallet'
4. Scroll down to 'metamask'
5. See error

**Expected behavior**

**Screenshots**

 - OS: iOS
 - Browser Metamask

**Smartphone (please complete the following information):**
 - Device: iPhone 11
 - OS: iOS 15.6.1
 - Browser Metamask

IljaDaderko:
⚠️ Don't merge until alpha.6 is published as docs are changed

- Update wagmi to latest release, adjust `signer` implementation to account for changes
- Introduce concept of required args to hooks, which skips fetching data if they are not provided
- Export better `useAccount` defaults, that make composition with other hooks easier
- Fix  #600


packages/react/src/hooks/data/useAccount.ts
```diff
@@ -1,13 +1,21 @@
...
     watchFn: AccountCtrl.watch,
     args: undefined
   })
+  const account = {
+    address: data?.address ?? '',
+    connector: data?.connector,
+    isConnected: data?.isConnected,
+    isReconnecting: data?.isReconnecting,
+    isConnecting: data?.isConnecting,
+    isDisconnected: data?.isConnected,
```
crypblizz8, line 16: 🤝

packages/core/src/controllers/stateless/SignerCtrl.ts
```diff
@@ -1,20 +1,19 @@
...
-  SignerCtrlWatchCallback,
-  SignerCtrlWatchOptions
+  SignerCtrlWatchCallback
 } from '../../../types/statelessCtrlTypes'
 import { ClientCtrl } from '../statefull/ClientCtrl'
 
 export const SignerCtrl = {
-  watch(options: SignerCtrlWatchOptions, callback: SignerCtrlWatchCallback) {
-    const unwatch = ClientCtrl.ethereum().watchSigner(options, callback)
+  watch(_options: undefined, callback: SignerCtrlWatchCallback) {
```
crypblizz8, line 9: This is fine to leave as undefined?

examples/react/src/sections/UseBalance.tsx
```diff
@@ -3,8 +3,7 @@ import { useAccount, useBalance } from '@web3modal/react'
 export default function UseBalance() {
   const { address } = useAccount()
   const { data, error, isLoading, refetch } = useBalance({
-    addressOrName: address ?? '',
-    enabled: Boolean(address)
+    addressOrName: address
```
crypblizz8, line 6: 🤝

packages/core/src/controllers/stateless/SignerCtrl.ts
IljaDaderko: It's a cost of keeping `useAsyncController` simple and as only utility hook for react (other controllers have options, callback) pattern while some only have callback

crypblizz8 (approved):
LGTM, left 1-2 comments on types. 

Went through
- Changes in `useAccount`
- `hasRequiredArgs` check 
- Upgrade to the new package. 
- MobileWallet Injected Fix.

packages/react/readme.md
```diff
@@ -108,7 +108,7 @@ const { address, isConnected } = useAccount()
 
 // Returns
 interface Return {
-  address?: string
+  address: string | ''
```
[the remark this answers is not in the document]
IljaDaderko: Yeh, but I wanted to highlight it in the docs, that address can be an empty string if user is not connected instead of undefined. Reason for it not being undefined is to make composability with hooks like `useEnsName` `useBalance` a bit cleaner (they expect string for some required options, however if this string is empty or there are missing required options, they handle it internally). Get's rid of patterns like this

```ts
const { data, error, isLoading, refetch } = useBalance({
    addressOrName: address ?? '',
    enabled: Boolean(address)
 })
```

packages/react/src/hooks/utils/useAsyncController.ts
```diff
@@ -12,22 +12,25 @@ interface Options<TArgs, TReturn> {
...
   fetchFn,
   watchFn,
-  args
+  args,
+  hasRequiredArgs
 }: Options<TArgs, TReturn>) {
   const { enabled, watch } = args
   const isEnabled = typeof enabled === 'undefined' ? true : enabled
-  const [isLoading, setIsLoading] = useState(isEnabled)
+  const isRequiredArgs = typeof hasRequiredArgs === 'undefined' ? true : hasRequiredArgs
```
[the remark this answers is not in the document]
IljaDaderko: Similar to `enabled` check, essentially if no `hasRequiredArgs` were provided, we assume that everything is good i.e. true, but if they were provided (meaning that hook needs some required options to fetch data) we use whatever value `hasRequiredArgs` is.

With `enabled` its the same i.e. if it was not specified, we fetch on mount, if it was specified we use whatever the value of it is (usually false when it is used), which delays initial fetch until user calls explicit action.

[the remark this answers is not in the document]
devceline: Got it!

PR #622, opened 2022-10-06
alpha.7

IljaDaderko:
This fixes following edge case:

Sometimes wagmi's get / watch functions return null before returning actual value. Since we use `undefined` in react hooks, this triggers un-necessary re-renders and sets `isLoading` to false incorrectly.


packages/react/src/hooks/utils/useController.ts
```diff
@@ -1,10 +1,13 @@
+import { CoreHelpers } from '@web3modal/core'
 import { useEffect, useMemo, useState } from 'react'
 import { useClientInitialized } from '../data/useClientInitialized'
 
 // -- types ----------------------------------------------------- //
+type NullishReturn<TReturn> = TReturn | null
```
devceline, line 6: Name's kinda confusing, maybe we could stick to normal standards and call it Nullable?

devceline (approved):
LGTM

packages/react/src/hooks/utils/useController.ts
IljaDaderko: Totally, no idea why I made it so complex 😂 changed

crypblizz8 (approved):

PR #624, opened 2022-10-07
docs(readme): adds development instructions

bkrem:
Adding some dev setup instructions while I'm seeing it with fresh eyes.

Also:
- removes seemingly dead `dev:html` script

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -1,0 +1,52 @@
+# Development
+
+## Workspace setup
+
+Install dependencies from the repository's root directory (this will also set up each workspace):
+
+```bash
+yarn
+```
+
+Common commands:
+
+- `yarn dev` - Build and watch all packages for changes simultaneously.
+- `yarn build` - Build all packages + examples.
+- `yarn lint` - Run the linter.
+- `yarn prettier` - Run prettier.
+
+## React Example
+
+### Configuration
+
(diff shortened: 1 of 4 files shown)


crypblizz8 (approved):
LGTM, as suggested by @devceline, we could potentially add a sidenote that if `yarn link` fails , to try `yarn cache clean` + `yarn add`

IljaDaderko (approved):
Looks good, maybe we can create `development.md` and link to it from main readme? Mainly to keep initial impression less intimidating for newcomers, people sometimes tend to think they need to run development setups to use the product.

re failing symlinking. I think yarn does that by default when `workspaces` key is present in root package.json, issue might be that there are multiple rollup jobs running `watch` at the same time and they sometimes fail to reflect changes, thus need to run `yarn build`. This was bothering me as well, not sure of an elegant solution besides rebuilding atm

bkrem:
* @IljaDaderko done, separated into `docs/development.md` and linked it under `Contributing` in the main readme instead to make it clear that it's separate from regular usage.
* @crypblizz8 definitely, can you outline how exactly you two are doing this atm with that approach? With `yarn add` it doesn't seem to let me use relative paths (e.g. `yarn add "./../../packages/core/"`), and I'd be surprised if you're specifying absolute paths every time 🤔

IljaDaderko (approved):

PR #630, opened 2022-10-13
alpha.8

IljaDaderko:
⚠️ Only merge once alpha.8 is released

- Update to wagmi .7 and address breaking changes
- Vanilla usage docs and minimal example (just so we have something)
- Other "housekeeping" tasks


docs/development.md
```diff
@@ -29,12 +29,19 @@ cp .env.local.example .env.local
 # 3. Replace the NEXT_PUBLIC_PROJECT_ID placeholder inside `.env.local` with your own projectId
 ```
 
-Run the app:
+Bild packages in dev:
```
devceline, line 32:
```suggestion
Build packages in dev:
```

rollup.config.js
```diff
@@ -3,6 +3,7 @@ import json from '@rollup/plugin-json'
 import resolve from '@rollup/plugin-node-resolve'
 import esbuild from 'rollup-plugin-esbuild'
 import minifyHtml from 'rollup-plugin-minify-html-literals'
+import nodePolyfills from 'rollup-plugin-polyfill-node'
```
devceline, line 6: Since this includes `buffer` according to their [docs](https://github.com/FredKSchott/rollup-plugin-polyfill-node#nodejs-builtin-support-table), why are we also depend on `buffer` when doing `window.Buffer = Buffer` above?

devceline (commented):
Took a glance over the files, like the clearer rename for the modal. Left a question.

packages/react/src/hooks/data/useProvider.ts
```diff
@@ -1,18 +1,18 @@
...
+import { useController } from '../utils/useController'
 
 interface Options {
   chainId?: number
 }
 
 export function useProvider(options?: Options) {
   const chainAgnosticOptions = useChainAgnosticOptions(options ?? {})
-  const { data } = useController({
+  const { data, isReady } = useController({
```
crypblizz8, line 11: 🤝

crypblizz8 (approved):
LGTM, small comment on useAccount.

Crytogenz:
@IljaDaderko There is an error when attempting to run alpha.8 and also seems to happen in alpha.7. You mentioned here https://github.com/WalletConnect/web3modal/issues/628 it was resolved, but looks like its not. 

```error - ../../chains/ethereum/index.ts
Module parse failed: Unexpected token (1:7)
You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
> export type { Chain, ChainProviderFn, Connector, Provider, WebSocketProvider } from '@wagmi/core'
| export { Web3ModalEthereum } from './src/api'
| export { chains, providers } from './src/utils/wagmiTools'

File: rollup.config.js (hunk 2 above)
Comment (IljaDaderko) [reply rollup.config.js:6]: We only use this `nodePolyfills` plugin in umd builds, es onesgo through just esbuild. Needed to add polyfills here for some other missing "vanilla" deps.

File: examples/react/src/pages/index.tsx [hunk 4]
```diff
@@ -23,9 +23,9 @@ import UseToken from '../sections/UseToken'
 import UseTransaction from '../sections/UseTransaction'
 
 export default function HomePage() {
-  const { isConnected } = useAccount()
```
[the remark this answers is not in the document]
IljaDaderko: Underlying utility hook for this changed for this (it also adds `isReady`). I kept same pattern here for consistency. Once we start working on solana, we'll likely deviate from being strictly 1to1 with wagmi anyway.

PR #649, opened 2022-10-24
[fix] react example in dev mode

IljaDaderko:
There was some caching done within `.next` folder of react example that made following error slip by:

When repository is pulled fresh and users try to run react example, they encounter an error like this
```
error - ../../chains/ethereum/index.ts
Module parse failed: Unexpected token (1:7)
You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
> export type { Chain, ChainProviderFn, Connector, Provider, WebSocketProvider } from '@wagmi/core'
| export { Web3ModalEthereum } from './src/api'
| export { chains, providers } from './src/utils/wagmiTools'
``` 

After some investigation, it looks like it was happening because nextjs in `dev` mode does not transpile linked monorepo packages (not an issue during build as all of them are pre-built or when these are used from npm). Sollution is to use [next-transpile-modules](https://www.npmjs.com/package/next-transpile-modules)

closes https://github.com/WalletConnect/web3modal/issues/628

### DIFF
--- a/examples/react/next.config.js
+++ b/examples/react/next.config.js
@@ -1,0 +1,16 @@
+/**
+ * @type {import('next').NextConfig}
+ */
+
+/**
+ * NOTE: This config is only required within this monorepo.
+ * Nextjs does not transpile web3modal packages imported from monorepo and throws an error.
+ */
+
+const nextTranspileModules = require('next-transpile-modules')
+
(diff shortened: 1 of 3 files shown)


s-Shaheen (approved):
LGTM

devceline (approved):
LGTM

PR #655, opened 2022-10-26
support kcc chain

ibwei:


IljaDaderko:
Hi @ibwei thank you for this, but in an effort to keep our bundle sizes smaller, we will be focusing on providing presets for most used / top ~100 marketcap chains.

That said, chain config that you created can be passed directly to web3modal ethereum config options, so you can start using it 👍 

In the future we will document this (how to use custom chains) and provide more info on how to provide images that will be used in the ui.

PR #656, opened 2022-10-26
beta.1

IljaDaderko:
- Add chain selection view, enabled automatically if more than 1 chain were added to configuration object
- Handle user selected chain throughout lifecycle of the modal and update it accordingly based on network switch events
- Improved initial open animation performance by delaying canvas gradient through a neat / almost imperceptible css gradient placeholder
- Preloading of assets happens earlier in the lifecycle now, thus reducing waiting time during first "Connect" click
- Introduced `Web3Button` component that automatically switches between connect and account buttons without requiring users to manually manage these
- Added few adjustment from new designs: No more offset in background gradient, header border
- Updated dependencies
- Update dev commands to include / watch react package. Updated dev readme


packages/ui/src/components/w3m-modal/styles.ts
```diff
@@ -29,10 +29,9 @@ export default css`
   .w3m-modal-card {
     width: 100%;
     position: relative;
-    transform: translateX(5px) translateY(5px);
-    border-radius: 40px;
+    transform: translateX(0) translateY(5px);
```
devceline, line 32: We don't really need `translateX(0)`

devceline (approved):
LGTM, tested on the branch preview and everything works.

packages/ui/src/components/w3m-modal/styles.ts
IljaDaderko: True, good catch

PR #667, opened 2022-10-31
Beta.2

IljaDaderko:


examples/react/package.json
```diff
@@ -4,19 +4,18 @@
...
   "scripts": {
     "dev": "next dev",
-    "build": "next build; next export"
+    "build": "next build && next export"
   },
   "dependencies": {
     "@web3modal/ethereum": "2.0.0-beta.1",
     "@web3modal/react": "2.0.0-beta.1",
-    "next": "12.3.1",
+    "next": "13.0.2",
```
devceline, line 12: Tech stack so fresh

packages/ui/src/components/w3m-modal-toast/styles.ts
```diff
@@ -9,14 +9,16 @@ export default css`
     justify-content: center;
     align-items: center;
     padding: 10px 15px;
-    border-radius: 36px;
     position: absolute;
-    bottom: 60px;
+    top: 12px;
```
devceline, line 13: We should really look into more sturdy styling, absolute position always manages to break at some point

packages/ui/src/components/w3m-network-button/styles.ts
```diff
@@ -16,8 +16,8 @@ export default css`
     flex-direction: column;
     align-items: center;
     justify-content: center;
-    width: 80px;
-    height: 80px;
+    width: 78px;
```
devceline, line 19: Is there no way we can make these `%` or `em`? just be more responsive just in case

devceline (commented):
1. Is this intended?
image

2. Copying to clipboard looks a little weird
image


Overall, looks great and works amazing

packages/ui/src/components/w3m-network-button/styles.ts
IljaDaderko: I'll try to play with these, to see if I can 👍 but they might look a bit off on wider mobiles, as we stretch there. Perhaps `display: grid` could be good to investigate as well

packages/ui/src/components/w3m-modal-toast/styles.ts
IljaDaderko: I agree, but I'm not sure if there is way around this here, as this toast thingy needs to "cover over" main content

packages/core/src/controllers/statefull/ExplorerCtrl.ts
```diff
@@ -6,7 +6,8 @@ import { fetchWallets } from '../../utils/ExplorerApi'
 const state = proxy<ExplorerCtrlState>({
   wallets: { listings: [], total: 0, page: 1 },
   search: { listings: [], total: 0, page: 1 },
-  previewWallets: []
+  previewWallets: [],
+  recomendedWallets: []
```
crypblizz8, line 10: 🤝

packages/ui/src/views/w3m-get-wallet-view/styles.ts
```diff
@@ -38,16 +53,8 @@ export function dynamicStyles() {
   const { background } = color()
 
   return html`<style>
-    .w3m-wallet-content {
+    .w3m-walle-item {
```
crypblizz8, line 56: Nitpick: `.w3m-walle-item` --> .w3m-wallet-item + Line 17 

Sorry...

crypblizz8 (approved):
LGTM. Really impressive work + looking forward to this release.

PR #686, opened 2022-11-07
Feat/implement scss

devceline:
- Add `scss`, `postcss` and `autoprefixer` to rollup
- Move all css into `.scss` files
- add `scss` template literal function to import scss files
- add `global.scss` for `MOBILE_BREAKPOINT` and future global `scss` variables and mixins
- Modify all `styles.ts` files to import new scss


packages/ui/package.json
```diff
@@ -24,7 +24,9 @@
   "devDependencies": {
     "@types/qrcode": "1.5.0",
     "eslint-plugin-lit": "1.6.1",
-    "eslint-plugin-wc": "1.3.2"
+    "eslint-plugin-wc": "1.3.2",
+    "rollup-plugin-scss": "^3.0.0",
```
IljaDaderko, line 28: rollup plugin for scss is also already defined in root package.json, lets keep all rollup lugins there

packages/ui/src/components/w3m-button/styles.ts
```diff
@@ -1,29 +1,11 @@
...
-
-  .w3m-button-icon-left svg {
-    margin-right: 5px;
-  }
-
-  .w3m-button-icon-right svg {
-    margin-left: 5px;
-  }
+export default scss`
+  ${scssStyles}
```
IljaDaderko, line 8: Not a huge fan of this, it seems to be adding complexity rather than moving it away. Why do we need to do this again?

I was under assumption that we could just do following
```
import styles from "./styles.scss"

// Inside LitElement class
styles = style
```

And for now only leave `dynamicStyles` in the .ts file

packages/ui/tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "extends": "../../tsconfig.json",
-  "include": ["index.ts", "src", "types"],
+  "include": ["index.ts", "src", "types", "scssDeclaration.d.ts"],
```
IljaDaderko, line 3: This is already included in root tsconfig, so it should either be just here or just there.

packages/ui/src/components/w3m-button/styles.ts
devceline: We could actually do, with the current structure

```ts

import styles from './styles.scss'

// inside LitElement class
styles = scss`${style}` 
// or
styles = scss(style)
```

I just kept it in `styles.ts` because it made sense to me to have one place where we import static and dynamic styles. In the future we could just have this in the class

PR #689, opened 2022-11-08
Beta.3

### LINKED ISSUE #687: Metamask no longer injects web.
**Describe the bug**
When using the "open" method of the hook useConnectModal it is not possible to connect to Metamask.

The error displayed is as follows:
imagen

The application freezes for a few seconds. Then it shows the menu without the Metamask option.

**SDK Version**
 - Version 2.0.0-beta.1 & 2.0.0-beta.2
 - React APP

**To Reproduce**

const { open } = useConnectModal()

1. Call the function "open"
2. Click on button
3. See error

**Expected behavior**
With <ConnectButton /> the problem is solved.
I understand that the operation should be the same, except for the aesthetics of the button.

 - OS: Windows
 - Browser Firefox and Chrome

### LINKED ISSUE #688: Allow to specify namespaces during standalone modal open
**Is your feature request related to a problem? Please describe.**
In some scenarios configuring namespaces in standalone modal config is not sufficient, as users can open modal with namespace config after it has been configured (i.e. our example dapp)

IljaDaderko:
- closes #688
- closes #687
- Updated deps
- Updated preloading strategy

### DIFF
--- a/examples/react-standalone/src/pages/index.tsx
+++ b/examples/react-standalone/src/pages/index.tsx
@@ -25,8 +25,7 @@ async function configureSignClient() {
 ConfigCtrl.setConfig({
   projectId: process.env.NEXT_PUBLIC_PROJECT_ID,
   theme: 'light' as const,
-  accentColor: 'orange' as const,
-  standaloneChains: namespaces.eip155.chains
+  accentColor: 'orange' as const
 })
 
 // 4. Manage manual connection
@@ -37,7 +36,7 @@ export default function HomePage() {
     if (signClient) {
       const { uri, approval } = await signClient.connect({ requiredNamespaces: namespaces })
       if (uri) {
-        ModalCtrl.open(uri)
+        ModalCtrl.open({ uri, standaloneChains: namespaces.eip155.chains })
         await approval()
         ModalCtrl.close()
       }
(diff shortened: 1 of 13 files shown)


examples/react-standalone/src/pages/index.tsx
```diff
@@ -37,7 +36,7 @@ export default function HomePage() {
     if (signClient) {
       const { uri, approval } = await signClient.connect({ requiredNamespaces: namespaces })
       if (uri) {
-        ModalCtrl.open(uri)
+        ModalCtrl.open({ uri, standaloneChains: namespaces.eip155.chains })
```
crypblizz8, line 39: My understanding is that in Ben's PR, it would not allow standaloneChains to pass + defaulted to line 40?

crypblizz8 (approved):
LGTM, ty legend

examples/react-standalone/src/pages/index.tsx
IljaDaderko: This would be how ben uses it essentially (initialize modal on app start without standalone chains and pass them in the modal). This also triggers different behaviour in preloading of the assets i.e. instead of app start they are preloaded during open

Issue #683, opened 2022-11-07
Vanilla usage and stability improvements


IljaDaderko:
- [ ] Complete example to show usage of all controllers
- [ ] Address github issues / discussions

IljaDaderko:
No longer relevant, as we are using wagmi

Issue #682, opened 2022-11-07
React package stability improvements


IljaDaderko:
- [ ] Address github issues and discussions
- [ ] More tests for hook composition
- [ ] Implement contract abi type completion with wagmi
- [ ] Address incorrect type issues

IljaDaderko:
No longer relevant as we are using wagmi

Issue #669, opened 2022-11-01
Error: underlying network changed when connected to Metamask


doerfli:
**Describe the bug**
The error _underlying network changed_ is shown when i try to call a contract after connecting to the blockchain via metamask. 

**SDK Version**
 - Version 2.0.0-beta.1

**To Reproduce**
1. Check out repository `https://github.com/etherisc/poc-wallet-access/` and switch to branch `bug/web3modal-metamask`
2. Create .env.local file with contents 
```
NEXT_PUBLIC_COUNTER_ADDRESS=0x6A5bEdE3957275C6bF2F8dd8C16B0F099422AA10
```
4. Run `npm install`
5. Run `npm run dev` to start the application
6. Naviate to http://localhost:3000/ with a browser that has the Metamask plugin installed and a connection to the fuji avalanche blockchain configured. 
7. Click 'connect wallet' on top right
8. Follow instructions in Metamask to create connection
9. See error

**Expected behavior**
The call to the contract does not fail as connection to metamask has been successful. 

**Screenshots**
<img width="1439" alt="Counter" src="https://user-images.githubusercontent.com/7255610/199244828-3f8c0af9-4cce-44f3-9100-d4e195cab6d5.png">

**Desktop (please complete the following information):**
 - OS: MacOS 12.6.1
 - Browser Chrome, Firefox
 - Version 107, 106.2

**Additional context**
The relevant code is here https://github.com/etherisc/poc-wallet-access/blob/bug/web3modal-metamask/pages/index.tsx#L15-L25
Up update of the signer (useSigner) via useEffect it connects to a smart contract and calls a read method.

civa:
same issue

IljaDaderko:
@doerfli @civa just an idea, but could you try following:

1. Do not use `useProvider` until user is connected
2. When user connects, ensure avalanche fuji (chain from the error) is added to your wallet (this should happen automatically if you don't have it, just need to approve)
3. Once user is connected and chain is added, do you still see the error?

I looked into this and this issue seems to be related to ethersjs. I will look into this as a workaround https://github.com/Uniswap/web3-react/issues/127 . My hunch is that ethersjs is trying to do some work on non-existent chain before user is connected and chain is added.

JSanchezFDZ:
Same error when:

Config chains -> Mumbai in web3Modal
Connect in your dApp con Mumbai
Change to Ehtereum and back to Mumbai
Now you can no longer interact with the smart contract.
[...]

Issue #661, opened 2022-10-29
Unable to pass signer in useContract.signerOrProvider hook options


crackcodesprogramloads:
**Describe the bug**
When I passed the signer data returned from useSigner hook in useContract.signerOrProvider hook, it threw this error: `Uncaught TypeError: Converting circular structure to JSON
    --> starting at object with constructor 'd'
    |     property '_readableState' -> object with constructor 'b'
    |     property 'pipes' -> object with constructor 'i'
    |     property '_readableState' -> object with constructor 'b'
    --- property 'pipes' closes the circle`

**SDK Version**
 - Version [2.0.0-alpha.2]

**To Reproduce**
```
    const { data: signer } = useSigner({ chainId: 5 })
    const { contract, isReady: isContractReady } = useContract({
        address: process.env.REACT_APP_PUBLIC_DEPLOYED_CONTRACT_ADDRESS,
        abi: Waitlist.abi,
        signerOrProvider: signer
    })
```

**Expected behavior**
Will let me pass signer to useContract

**Screenshots**
<img width="986" alt="Screen Shot 2022-10-29 at 12 54 36 pm" src="https://user-images.githubusercontent.com/59869248/198763316-67c982a2-44b1-45dc-ad48-16b7d758f3ce.png">

**Desktop (please complete the following information):**
 - OS: macOS 12.4
 - Browser: Chrome 
 - Version: 105.0.5195.125 (Official Build) (arm64)

b4rtt:
I have same problem here

`import { useAccount, useContractRead } from "@web3modal/react";

import USDC from "./ABI/USDC.json";
import { chains } from "@web3modal/ethereum";

function Approve() {
  const { account } = useAccount();

  const config = {
    address: "xxx",
    abi: USDC,
    functionName: "allowance",
    chainId: chains.polygonMumbai.id,
    args: [account, "xxx"],
  };

  const { data, error, isLoading, refetch } = useContractRead(config);

  return (
    <div className="Approve">
      {account.isConnected && (
        <>
          <ul>
            <li>
              Contract read config: <span>{JSON.stringify(config)}</span>
            </li>
            <li>
              Returned data:{" "}
              <span>{isLoading ? "Loading..." : JSON.stringify(data)}</span>
            </li>
            <li>
              Error: <span>{error ? error.message : "No Error"}</span>
            </li>
          </ul>
          <button onClick={async () => refetch()}>Refetch data</button>
        </>
      )}
    </div>
  );
}

[...]

PR #691, opened 2022-11-08
Beta.4

### LINKED ISSUE #677: Errors on Vanilla (Beta.2)
**Describe the bug**
There is a clear improvement on beta 2 compared to beta 1, however, we still have some errors: 
- see image (use switchNetwork + querySelector (UIHelpers.js))
- sometimes on Event "**NetworkCtrl.watch((network)...**" -> chain name = 'Chain1' instead of 'Ethereum' 

**FYI:**  We reused the component <w3-modal/> several times  (no reallocation)

So, another question: 
LOBSTR (Stellar) is compatible WalletConnect v2, but why is it not in your list of wallets?

**Envirionement Dev**
- Vue + Nuxt v2
- Use Web3Modal + Vanilla

**SDK Version**
 - Version [2.0.0-beta.2]

**To Reproduce**
```
  <template>
    <div>
      <w3m-modal />
    </div>
  </template>


 import {
  AccountCtrl,
  BalanceCtrl,
  ClientCtrl,
  ConfigCtrl,
  ModalCtrl,
  NetworkCtrl,
  TransactionCtrl
 } from '@web3modal/core'
 import { chains, providers } from '@web3modal/ethereum'
 import '@web3modal/ui'
  ...
  data: { 
    clientConfig: {
          projectId: PROJECT_ID,
          theme: 'dark',
          accentColor: 'green'
     }
    ethereumConfig: {
          appName: 'O21Pay',
          autoConnect: true,
          chains: [],
          providers: []
    },
   watchAccount: undefined,
  }
  
  async connect(chainId) {
      chainId = parseInt(chainId)

      if (chainId === 1) {
        this.ethereumConfig.chains = [chains.mainnet]
      } else if (chainId === 97) {
        this.ethereumConfig.chains = [chains.binanceSmartChainTestnet]
      } else {
        return
      }

      await ConfigCtrl.setConfig(this.clientConfig)
      await ClientCtrl.setEthereumClient(this.ethereumConfig)

      if (this.watchAccount) {
        this.watchAccount()
        this.watchAccount = undefined
      }
      // We force the network switch, because on the beta.1, "ClientCtrl.setEthereumClient(this.ethereumConfig)" was not 
      //   enough -> see chains
      NetworkCtrl.switchNetwork({ chainId })

      this.watchAccount = AccountCtrl.watch(async (account) => {
        console.log('******* Account ')
      })

      const account = AccountCtrl.get()
      if (account.isConnected === false) {
        ModalCtrl.open()
     }
  } 
```

**Expected behavior**
During the call of the method '**connect()**', we have systematically these 2 errors which are displayed...

**Screenshots**
(https://gist.github.com/Droppix/c6411d4c75ee42d89fc74f4fd85a6921)

 - OS: [MacOS]
 - Browser [chrome]
 - Version [22]

Thanks

IljaDaderko:
⚠️ Don't merge until we coordinate with new docs release

- Removed custom modal hooks, we are now encouraging using wagmi directly
- Removed stateless hooks
- Removed hooks from react / react examples
- Renamed `useConnectModal` hook, to `useWeb3Modal`
- `ClientCtrl` now routes requests through generalised `.client()` that will be responsible for routing to ethereum / solana / cosmos etc... implementations
- `@web3modal/ethereum` package was drastically simplified, it now only exports functions required by modal ui package, everything else should be used from source i.e. wagmi or solib
- Added `encodeUriComponent` handling to link formatter helpers, so that it works with standalone mode
- Updated examples

closes #677 #661 #669 #682 #683 #697 #698 #699


chains/ethereum/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@web3modal/ethereum",
-  "version": "2.0.0-beta.3",
+  "version": "2.0.0-a38e60a",
```
devceline: Do we want to stick with this version?

chains/ethereum/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@web3modal/ethereum",
-  "version": "2.0.0-beta.3",
+  "version": "2.0.0-c2564bd",
```
devceline, line 3: Are we sure we want to stick with this version

examples/html/index.html
```diff
@@ -2,22 +2,14 @@
 <html lang="en">
   <head>
     <meta charset="UTF-8" />
-    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <title>HTML Example</title>
-    <link rel="stylesheet" href="style.css" />
   </head>
```
devceline, line 6: Why remove these

packages/ui/src/partials/w3m-core-button/index.ts
```diff
@@ -8,14 +8,15 @@ import '../w3m-connect-button'
 @customElement('w3m-core-button')
 export class W3mCoreButton extends LitElement {
   // -- state & properties ------------------------------------------- //
-  @state() public isConnected = AccountCtrl.get().isConnected
+  @state() public isConnected = false
```
devceline, line 11: What are the implications of this change?

devceline (commented):
Looks great, looked over every file.  It works. However, now the demo/example app does not have any sections to show off the app working. Maybe we can show Wagmi hooks working, or something along those lines because the example app works as implicit docs.

examples/html/index.html
IljaDaderko: All these removals were intentional to simplify the examples. I had same idea about wagmi section examples, but it is unnecessary chore for us to keep them updated as opposed to linking to wagmi. Instead, I am planning on full blown dapp example in our `web-examples` repo that showcases them at some point. But for examples within this repo, I think simplest possible ones do the trick.

packages/ui/src/partials/w3m-core-button/index.ts
IljaDaderko: I moved this down into constructor (as it is now coming from client, there was ssr error for undefined ethereum client in next)

PR #696, opened 2022-11-10
Fix: Updated React Examples

crypblizz8:
### Overview
- Explores #669 
- This seems to behave as per our library if configured correctly.

### Tested & Works
- Have tested the userflow of network switching of networks (ETH + Goerli) + reading + writing to different contracts with `useContractWrite` and `useContractRead`
- Making calls using `useContract` with `signerOrProvider: provider` / nothing passes makes the calls for data.

The recommended approach is:
1. import the correct `chains` in your modalConfig with` import { chains, providers } from '@web3modal/ethereum'`
2. Wait for users to Connect before calling hooks such as `useSigner` or `useProvider`.
3. Make sure you pass your `chainId` if you are using `useProvider` 
```
  const { network } = useNetwork()
  const chainId = network?.chain?.id
  const { provider } = useProvider({ chainId })
```
4. `useContractRead` and `useContractWrite` as opposed to `useContract`.

### To Try
1. On the interactive example / CloudFlare link below, Connect Wallet on Mainnet
2. Check `useContract` section + call the getHunger endpoint. Data should populate
3. Switch networks to Goerli in `Network Swtich` section.
4. Check `useContract` section again and the contractAddress / contract should be for Goerli USDC. (Will populate and flag check for the functions after the below debugging is solved)

### Bug / Debugging
- However while testing this ran into another issue.
When passing in signer into `useContract`, faces an issue of JSON flattening. Similarly found in [Wagmi repo in #367 ](https://github.com/wagmi-dev/wagmi/issues/367)
image

### Other Notes
- Can remove the goerli network / mumbai testnetwork + USDC contract at end of the debugging but might be useful for developers to see the contract switch

### DIFF
--- a/examples/react/src/data/goerliUSDC.json
+++ b/examples/react/src/data/goerliUSDC.json
@@ -1,0 +1,63 @@
+[
+  {
+    "inputs": [{ "internalType": "address", "name": "implementationContract", "type": "address" }],
+    "stateMutability": "nonpayable",
+    "type": "constructor"
+  },
+  {
+    "anonymous": false,
+    "inputs": [
+      { "indexed": false, "internalType": "address", "name": "previousAdmin", "type": "address" },
+      { "indexed": false, "internalType": "address", "name": "newAdmin", "type": "address" }
+    ],
+    "name": "AdminChanged",
+    "type": "event"
+  },
+  {
+    "anonymous": false,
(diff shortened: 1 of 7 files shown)


IljaDaderko:
Closing in favour of #691

PR #708, opened 2022-11-16
Beta.5

IljaDaderko:
- Hotfix for wagmi and wagmi/core version compatibility
- Dependency updates / cleanup

### DIFF
--- a/chains/ethereum/src/utils.ts
+++ b/chains/ethereum/src/utils.ts
@@ -18,16 +18,13 @@ export function walletConnectProvider({ projectId }: WalletConnectProviderOpts) 
 }
 
 // -- connectors ------------------------------------------------------ //
-export function modalConnectors({ appName, chains: connectorChains }: ModalConnectorsOpts) {
+export function modalConnectors({ appName, chains }: ModalConnectorsOpts) {
   return [
-    new WalletConnectConnector({ chains: connectorChains, options: { qrcode: false } }),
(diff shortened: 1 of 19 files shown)


chains/ethereum/src/utils.ts
```diff
@@ -18,16 +18,13 @@ export function walletConnectProvider({ projectId }: WalletConnectProviderOpts)
 }
 
 // -- connectors ------------------------------------------------------ //
-export function modalConnectors({ appName, chains: connectorChains }: ModalConnectorsOpts) {
```
crypblizz8, line 21: Could you explain why this type was removed? Isn't it a better pattern to keep ? Unless it was causing issues

packages/core/src/controllers/ConfigCtrl.ts
```diff
@@ -24,10 +24,6 @@ export const ConfigCtrl = {
   },
 
   setConfig(config: ConfigOptions) {
-    if (!config.projectId)
```
crypblizz8, line 27: This is part of this beta?

crypblizz8 (commented):
LGTM. Two comments left. 

Thanks for fast dependency upgrade / removal.

chains/ethereum/src/utils.ts
IljaDaderko: Not the type, change here is basically `chains: connectorChains` to just `chains` so its a bit cleaner

PR #731, opened 2022-11-24
Beta.7

### LINKED ISSUE #732: Injected connector shown on mobile incorrectly
**Describe the bug**
Injected connector that should only be shown within mobile wallet browsers, is show incorrectly in safari.

**SDK Version**
 - Version 2.0.0-alpha.6

IljaDaderko:
## Related PR's
https://github.com/WalletConnect/walletconnect-docs/pull/342

## Changes
- [x] Move to vanilla css and css variable based theming (@devceline )
- [x] Decouple ui imports into central `index` files
- [x] Renamed `theme` -> `themeMode`, `accentColor` -> `themeColor` 
- [x] Introduced new config option `themeBackground: 'gradient' | 'accentColor'`
- [x] Add individual gradient colors for each accentColor as presets
- [x] Add `useWeb3ModalTheme` hook and associated controller
- [x] Documentation (new config options, theme hook, ConfigCtr.setThemeConfig)

closes #686 
closes #732


packages/core/readme.md
```diff
@@ -0,0 +1,20 @@
+### 📚 [Documentation](https://docs.walletconnect.com/2.0/introduction/web3modal/about)
```
devceline, line 1: Why are we repeating the same readme across multiple sub projects, wouldn't it be more useful to have dedicated readmes that describe what this package does?

packages/ui/tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "extends": "../../tsconfig.json",
-  "include": ["index.ts", "src", "types"],
+  "include": ["index.ts", "src", "../../global.d.ts"],
```
devceline, line 3: If the `global.d.ts` file is only used for `*.css`, wouldn't make it sense to have it in `packages/ui` and have it be called `uiGlobals.d.ts` or something. Doesn't make sense for that file to be at the root dir if it's only used in one package

devceline (dismissed):
LGTM, comments are extremely minor and feel free to ignore

packages/core/readme.md
IljaDaderko: Mainly for it to look nicer on npm

packages/ui/tsconfig.json
IljaDaderko: Also needed in react package, it errored during builds because `.css` was not recognised for some imported bits there

PR #738, opened 2022-11-28
Add html and standalone packages

IljaDaderko:
## Changes
- Add `@web3modal/html`
- Add `@web3modal/standalone`

Each package exposes `Web3Modal` client that makes working in those targets much easier

## Todo
- [x] Update documentation

### DIFF
--- a/packages/html/index.ts
+++ b/packages/html/index.ts
@@ -1,0 +1,1 @@
+export { Web3Modal } from './src/client'
--- a/packages/html/src/client.ts
+++ b/packages/html/src/client.ts
@@ -1,0 +1,35 @@
+import type { ConfigCtrlState } from '@web3modal/core'
+import { ClientCtrl, ConfigCtrl, ModalCtrl } from '@web3modal/core'
+import type { EthereumClient } from '@web3modal/ethereum'
+
+/**
+ * Types
+ */
+type Web3ModalConfig = Omit<ConfigCtrlState, 'enableStandaloneMode' | 'standaloneChains'>
+
+/**
+ * Client
+ */
+export class Web3Modal {
+  public constructor(config: Web3ModalConfig, client: EthereumClient) {
+    ConfigCtrl.setConfig(config)
+    ClientCtrl.setEthereumClient(client)
+    this.initUi()
+  }
+
+  private async initUi() {
+    if (typeof window !== 'undefined') {
+      await import('@web3modal/ui')
+      const modal = document.createElement('w3m-modal')
+      document.body.insertAdjacentElement('beforeend', modal)
+    }
+  }
+
+  public openModal = ModalCtrl.open
+
+  public closeModal = ModalCtrl.close
+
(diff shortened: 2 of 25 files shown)


packages/html/src/client.ts
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * Types
+ */
+type Web3ModalConfig = Omit<ConfigCtrlState, 'enableStandaloneMode' | 'standaloneChains'>
+
+/**
+ * Client
+ */
+export class Web3Modal {
+  public constructor(config: Web3ModalConfig, client: EthereumClient) {
```
devceline, line 14: I don't quite understand the vision with this.

Are we going to make a generic `Client` type? With the imminent integration of Solib, how do we plan to make this more generic?

This is absolutely not a blocker for this PR, just want to know

packages/standalone/index.ts
```diff
@@ -0,0 +1 @@
+export { Web3Modal } from './src/client'
```
devceline, line 1: Since this is the index for the package, maybe it'd be nice to also add

```suggestion
export { Web3Modal } from './src/client'
export { Web3Modal as default } from './src/client'
```

packages/html/index.ts
```diff
@@ -0,0 +1 @@
+export { Web3Modal } from './src/client'
```
devceline, line 1: Since this is the index for the package, it'd be nice to export Web3Modal as default to have a nicer import

```suggestion
export { Web3Modal } from './src/client'
export { Web3Modal as default} from './src/client'
```

Very minor, but it's what's typically expected of npm packages

devceline (approved):
LGTM, minor comments/question

crypblizz8 (commented):
LGTM. 

This is my understanding of this PR from reading the new packages + examples. Would like confirmation on it to be on the same page. 

### High Level 

HTML: Still dependant on @web3modal/ethereum + /core + /ui
Standlone: Just @web3modal/core + /ui + people will add in sign-client

### More Granular

Standalone: 
- Separated into @web3modal/standalone
- Config controls are set from the above package with Standalone Omittance
- Copies the normal config. Can devs still adjust them if they like through ConfigCtrlState (theme etc) ? I assume under ConfigCtrl.setConfig({ enableStandaloneMode: true, ...config }) this is possible?
- Only projectID necessary

HTML:
- Separated into @web3modal/html
- Similar to Standalone structure but also standaloneChains omittance


SideNote: Since we are starting to separate the Standalone, we should figure out the Vercel deployment of that repo.

packages/html/src/client.ts
IljaDaderko: As you said, this will become a generalised client type in the future that all clients should satisfy.

packages/standalone/index.ts
IljaDaderko: Keeping it consistent / one way to import for now.

crypblizz8 (approved):

PR #748, opened 2022-12-04
update logic to handle http(s) native urls thus support web wallets

WarrickFitz:
### Background
- https://oreid.io/ provides dapps with tooling that allows users to sign up for their dapp using Google and other oAuth providers. 
- Users logging into the dapp are then provided with a web based wallet. 
- OREID would like to provide its dapp customers the option to implement web3modal on their sites. 
- Users logging in would then have the ability to login with metamask etc. or select to login with OREID 

### Current challenge
When configuring web3model if you provide a normal web URL in the "native" property, the formatNativeUrl() method mangles the URL

### What has been done
- Updated the logic in formatNativeUrl() to check if the URL is http or https - if it is, then treat is as a regular web url that needs to be opened in a new tab rather than treating it as a native protocol URI (eg  ledgerlive://)
- Update the React Custom example to provide an example and easy test
- Updated w3m-desktop-connector-view so that if we're opening a regular web URL, it opens in a new tab (target = '_blank')

### How to test
- Run the react example
- Go to http://localhost:3000/custom
- Select Ethereum
- Select OREID 
- This should take you to https://oreid.io/deeplink/wc?uri=wc%253Ab60.....  (Note that this URL does not current function, it is simply for testing purposes, you should expect a 404)

### Todo
- [ ] Once this is approved I will test mobile and standalone etc. Just don't want to spend that time in case there is some fatal flaw in my thinking here. 

### Notes
- Happy to jump on a call and discuss this if that is helpful


IljaDaderko:
@WarrickFitz Good catch! Could you open this pr against `beta.8` branch please, I'll merge it in this Monday and will test it out (as you mentioned, change also touches native mobile apps).

WarrickFitz:
Ya sure, will close this and open another against the beta.8 branch.

WarrickFitz:
Looks like the `beta.8` branch currently has some other challenges. Could be something silly I'm doing (it's late), but I'd rather not mess with that side of things. Could you comment here when that branch is functional and I'll make the same pull request against that branch + test. 

image

IljaDaderko:
I'll port your changes over to beta.8 branch and will @ you on that PR once it is done 👍

WarrickFitz:
Fantastic. Thanks @IljaDaderko !

PR #755, opened 2022-12-07
Beta.9

IljaDaderko:
- [ ] Re-adjust connectors

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
   "devDependencies": {
     "@typescript-eslint/eslint-plugin": "5.45.1",
     "@typescript-eslint/parser": "5.45.1",
-    "esbuild": "0.15.18",
+    "esbuild": "0.16.1",
     "eslint": "8.29.0",
     "eslint-config-prettier": "8.5.0",
     "lerna": "6.1.0",
--- a/packages/ui/package.json
+++ b/packages/ui/package.json
@@ -18,7 +18,7 @@
   "dependencies": {
     "@web3modal/core": "2.0.0-beta.8",
     "lit": "2.4.1",
-    "motion": "10.15.0",
+    "motion": "10.15.3",
     "qrcode": "1.5.1"
   },
   "devDependencies": {
--- a/packages/ui/src/components/w3m-box-button/styles.css
+++ b/packages/ui/src/components/w3m-box-button/styles.css
@@ -23,3 +23,7 @@ button > div {
   border-radius: 50%;
   margin-bottom: 4px;
 }
+
+button path {
+  fill: var(--color-fg-inverse);
+}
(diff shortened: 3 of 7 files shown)


chains/ethereum/src/utils.ts
```diff
@@ -6,12 +6,13 @@ import type { ModalConnectorsOpts, WalletConnectProviderOpts } from './types'
 
 // -- constants ------------------------------------------------------- //
 export const NAMESPACE = 'eip155'
+const APP_TAG = 'app=2'
```
crypblizz8, line 9: Looks good, no action.

Trying to understand this process. This is just an RPC thing? I read the issue linked but not sure how the RPC identifies between 0 / 1 / 2? Could you explain it a bit more?

examples/react/src/pages/_app.tsx
```diff
@@ -3,7 +3,8 @@ import { Web3Modal } from '@web3modal/react'
 import type { AppProps } from 'next/app'
 import { useRouter } from 'next/router'
 import { useEffect, useState } from 'react'
-import { chain, configureChains, createClient, WagmiConfig } from 'wagmi'
+import { configureChains, createClient, WagmiConfig } from 'wagmi'
+import { arbitrum, avalanche, bsc, fantom, mainnet, optimism, polygon } from 'wagmi/chains'
```
crypblizz8, line 7: No action: cool how it is now `wagmi/chains`

chains/ethereum/src/utils.ts
0xAsimetriq: RPC and Explorer api tag to identify that web3modal is making these requests. 0 and 1 were reserved for other things as per https://github.com/WalletConnect/web3modal/issues/758

packages/ui/src/utils/PresetUtil.ts
```diff
@@ -18,21 +26,23 @@ export const PresetUtil = {
       tokenPocket: 'f3119826-4ef5-4d31-4789-d4ae5c18e400',
       tokenary: '5e481041-dc3c-4a81-373a-76bbde91b800',
       '1inch': 'dce1ee99-403f-44a9-9f94-20de30616500',
-      phantom: '62471a22-33cb-4e65-5b54-c3d9ea24b900'
+      phantom: '62471a22-33cb-4e65-5b54-c3d9ea24b900',
```
crypblizz8, line 29: Minor: Might have to be renamed phantomETH / phantomSOL wallet in near future?

crypblizz8 (dismissed):
LGTM

packages/ui/src/utils/PresetUtil.ts
0xAsimetriq: Yeh, I had similar thought actually. I think that presets will be moved out of `core` package into specific chain package instead, so id's for multichain wallets can stay the same, they'll just come from diff chain packages.

PR #788, opened 2022-12-20
add Zerion wallet support

zerts:


0xAsimetriq:
Injected wallets are now managed through https://cloud.walletconnect.com you can either create a new entry for your project or add your unique injected flag to existing entry.

PR #806, opened 2023-01-02
Rc.2

### LINKED ISSUE #781: When I switch the bsc network, I always get disconnected
**Describe the bug**
When I switch the bsc network, I always get disconnected

**SDK Version**
  "@web3modal/ethereum": "^2.0.0-beta.7",
  "@web3modal/react": "^2.0.0-beta.7",
  "wagmi": "^0.9.2",

### LINKED ISSUE #797: Handle use cases when wallet is not installed on android
3101671848200_ pic

### LINKED ISSUE #799: 'Connection Declined' error after wallet successfully connects in beta.10 / rc.1
When calling the `open` function, the modal displays an error showing 'Connection Declined.', despite the wagmi `useAccount` hook correctly returning the connected wallet's `address` and `isConnected` as true. If you reload the page, the hook continues to return data suggesting the wallet is connected, but upon calling the `open` function, it shows the web3modal as if no wallet is connected. At this point the `useAccount` hook returns `isConnected` as false and no longer returns the connected wallet's address.

**SDK Version**
2.0.0-rc.1

I have a repo that reproduces the bug issue here: https://github.com/kwebster2/web3modal-bug-demo
1. Connect metamask wallet and observe address displayed in place of Connect button (`useAccount` correctly returning `isConnected` and `address`)
2. Click the address to call the `open` function of the `useWeb3Modal` hook
3. An error is displayed
4. Reload the page, and note that the `useAccount` hook is still continuing to return the correct data
5. Click the address once again, and note that the web3modal appears as if the wallet is not connected. Attempting to connect again results in the same 'Connection Declined' error

**Expected behavior**
When the wallet is connected and you call the `open` function, the modal should display the screen with the wallet details and show that the wallet is connected

**Screenshots**
Screen Shot 2022-12-25 at 12 53 16

 - OS: macOS Monterey 12
 - Browser: Chrome
 - Version: Version 108.0.5359.124 

**Additional context**
This error was not happening with 2.0.0-beta.9. I searched the changelog for beta.10 and rc.1 and did not immediately see anything that might cause this. I did see the fallback provider change and tried adding a fallback InfuraProvider when configuring the chains since I am currently using Görli and mumbai testnets, but that did not work.

### LINKED ISSUE #804: Remove duplicate extension entries for custom wagmi connectors
At the moment we are only checking injected connector for duplicate entries agains our explorer api data. But there are other custom connectors like safe / ledger that we should also check agains i.e. expand this check to all connectors, not just injected.

### LINKED ISSUE #805: Add support and test wagmi version with universal-provider and WalletConnect v2
Recent version of wagmi merged support for WalletConnect v2 via universal provider, we need to implement default version and test everything.

Modal needs to update explorer filter version.

### LINKED ISSUE #809: Add ready / initialized flag to modal controller and handle race conditions in open method
We need to asynchronously import ui package under window check to support smooth ux in ssr environments. This requires some additional work within ModalCtrl

- Add readiness flag that gets set to true after everything is initialised and can be subscribed to in order to show spinners and such
- Handle and queue `open` methods that were called before configuration of modal was complete

0xAsimetriq:
## Changes
- closes #799 (fix router logic for undefined options)
- closes #805 (add support for `version: '2'` in wagmi)
- closes #804 (remove duplicate extension listings for custom connectors on best effort / case-insensitive name matches)
- closes #797 (add "I don't have a wallet" option to android view)
- closes #781 (resolved via updated rpc management)
- closes #809 (We need to load modal's web-component ui asynchronously as well as fetch preview wallets from explorer if projectId was provided. These might not be ready before modal's `open` method is called (slow networks, modal is only initialised on a specific user action). To ensure good ux, `open` is now async and will wait for these to become resolved, thus allowing users to show spinners / loaders in rare ocassions where it takes longer to bootstrap things.)
- Similarly to above, on very slow networks it takes some time to generate uri for qr code, thus we added fallback to show loader in those cases.

## Todo
- Document modal's open function arguments
- Document modal's open function async changes
- Add note for create-react-app users to add `defaults and supports es6-module` in browserlist config
- Add `custom` / `v1` / `v2` page toggles for web3modal test examples
- Add version query param to explorer to query wallets that support specific wc version
- Cut and publish `rc.2` release

## V2 (Super experimental)
Overall v2 stuff works, I think it is a bit early to expose it it in documentation, more work is needed to optimise it within universal-provider level and take advantage of some v2 specific functionality (single websocket). Atm it mimics behaviour of v1 which leads to some issues like longer pairing code generation.

https://user-images.githubusercontent.com/3154053/210991335-33d37784-4387-4674-a266-8708be4c15fd.mov


packages/ui/src/views/w3m-desktop-connector-view/index.ts
```diff
@@ -48,7 +48,7 @@ export class W3mDesktopConnectorView extends LitElement {
     }
   }
 
-  private async onConnect() {
+  private async createConnectionAndWait(retry = 0) {
```
devceline, line 51: Just trying to catch up: what is this solving?

chains/ethereum/src/client.ts
```diff
@@ -15,6 +15,8 @@ import type { ConnectorId } from './types'
 
 export class EthereumClient {
   private readonly wagmi = {} as Client
+  public walletConnectUri = ''
+  public walletConnectVersion = 1
```
devceline, line 19: Shouldn't this be `'1'`?

devceline (approved):
LGTM

packages/ui/src/views/w3m-desktop-connector-view/index.ts
0xAsimetriq: retries? There was a bug with sign v2 implementation that resulted in infinite fails / re-attempts. So I added this in as pre-caution for future (this bug was fixed, but kept this here just in case)

chains/ethereum/src/client.ts
0xAsimetriq: Kept is as number internally, but it is string in wagmi.

crypblizz8:
On Custom Example, some of the images are not being picked up. 

image

Also on the case I switch to one with an Image (i.e. Polygon), the balance icon is maybe not being picked up too.
image
image

packages/ui/src/utils/DataFilterUtil.ts
```diff
@@ -54,5 +54,16 @@ export const DataFilterUtil = {
     }
 
     return wallets
+  },
+
+  deduplicateExplorerListingsFromConnectors(listings: Listing[]) {
```
crypblizz8, line 59: NB: Understand this is for #804 but was still confused with this function. Could you please expand.

examples/react/src/pages/_app.tsx
```diff
@@ -1,23 +1,22 @@
...
 }
 
-const projectId = process.env.NEXT_PUBLIC_PROJECT_ID
+export const projectId = process.env.NEXT_PUBLIC_PROJECT_ID
 
 // 2. Configure wagmi client
 const chains = [mainnet, polygon, optimism, arbitrum, avalanche, fantom, bsc]
 const { provider } = configureChains(chains, [walletConnectProvider({ projectId })])
-const wagmiClient = createClient({
+export const wagmiClient = createClient({
```
crypblizz8, line 19: NB: Understand this is being used in `custom.tsx `but would these `exports` be a bit confusing for developers looking at this example? MB if this is a weird nitpick.

crypblizz8 (dismissed):
LGTM + approved. Not sure if the Custom.tsx missing images are blocking though

packages/ui/src/utils/DataFilterUtil.ts
0xAsimetriq: If wagmi connector exists who's name matches entry from our explorer api, connector will remain displayed while explorer listing gets hidden. This name also gave me a headache :D Open to suggestions here tbf

examples/react/src/pages/_app.tsx
0xAsimetriq: Yeh I agree, I think I will move wagmi initialisations etc... and have them defined in each page (need to do it anyway for wc v1 and v2 examples). Will be thinking on how to do this in a clean way on Monday. I think there might be conflicts in wamgi storage if I do it this way

PR #831, opened 2023-01-19
2.0.0

### LINKED ISSUE #817: Display ens data on all chains
Right now we are only showing resolved ens data on mainnet, but this can be also shown across other chains.

0xAsimetriq:
## Changes
- Fix some parent styling leaking into core web components
- closes #817 
- Fix web3-button layout shift when in "connecting" state


packages/ui/src/components/w3m-text/index.ts
```diff
@@ -14,8 +14,6 @@ type Variant =
   | 'xsmall-normal'
   | 'xxsmall-bold'
 
-type Align = 'center' | 'left' | 'right'
-
```
crypblizz8, line 18: NB: Just to check all of this removal is related to the `all:inherit` pattern?

crypblizz8 (approved):
LGTM, mainly package upgrades + CSS + restructure of ENSProfile fetching. GG

packages/ui/src/components/w3m-text/index.ts
0xAsimetriq: Moreso to simplify text props, as this can be managed with css easily where needed.

PR #847, opened 2023-01-28
fixed ledger error

umbecanessa:


0xAsimetriq:
We are no longer maintaining v1 of the modal I'm afraid.

PR #859, opened 2023-02-05
feat(laboratory): adds option to use sentry

arein:
Allows us to capture any errors when doing QA on mobile devices e.g. iOS

# Testing
Not tested


0xAsimetriq (commented):
You can add `bootstrapSentry()` one time in `_app.tsx` file, will be applied to all pages. Other than that, looks good, just need ci to pass. Lets also stick to naming convention in utilities i.e. `SentryUtil`

0xAsimetriq (approved):

PR #867, opened 2023-02-09
basic init of modal in angular

bws9000:


0xAsimetriq:
This seems too verbose for a simple example, could you strip it down to an absolute minimum that's required within angular app and replace usage of  `@web3modal/react` with `@web3modal/html`. You are also using `@web3modal/standalone`, which should likely be html.

bws9000:
> This seems too verbose for a simple example, could you strip it down to an absolute minimum that's required within angular app and replace usage of `@web3modal/react` with `@web3modal/html`. You are also using `@web3modal/standalone`, which should likely be html.
> 
> This should help with html package https://docs.walletconnect.com/2.0/web3modal/html-js/installation and make implementation in anguler much simpler

Hey great, no problem 😎 👍

0xAsimetriq:
Closing this as stale, feel free to re-open once comments left above get addressed.

PR #885, opened 2023-02-15
fix Less than 6 digits remain unchanged

ryosukemori:
If the balance is less than exactly 6 digits, w3m-balance will not show the balance correctly and `_._`

## Reproduction
Create a new wallet account, transfer 0.01 and connect with that wallet.

In the output of the Web3Modal component, the `_._` appears in the output of the Web3Modal component.

image

## Environment
Goerli Test net , Ethereum Main net


0xAsimetriq:
Thank you 🙏

PR #891, opened 2023-02-16
Fix bug that obfuscates/deprioritizes Coinbase Wallet

taycaldwell:
Fixes the following reported issue: https://github.com/WalletConnect/web3modal/issues/890


0xAsimetriq:
@taycaldwell Thank you for this, it will be part of next release https://github.com/WalletConnect/web3modal/pull/889

We removed Coinbase connector from defaults and all this associated logic, thus allowing users to add it in when they need it and show it on the first page of modal.

0xAsimetriq:
There are many other connectors like Safe / Ledger etc. that developers are adding, thus to treat them all properly a unified approach is needed. Coinbase is still part of the injected wallet list, even without connector, so users with browser extension can still use it, however to see full Coinbase sdk view, you just need to add the connector.

PR #906, opened 2023-02-23
Create 1dependabot.yml

HAWKS3CURE:
I do not read this jargon.
I believe in owed retribution. My PC is broken and it's federal when your devices are harmed even in open projects. 
I demand this to be fixed 
Please help to anyone that cares


HAWKS3CURE:
Why am I not authorized with my open source data

PR #907, opened 2023-02-24
Add Kucoin Wallet

574495412:


0xAsimetriq:
We are migrating to managing extension wallets via https://cloud.walletconnect.com to prepare for this you can already create/edit your project and add extension wallet details.

Web3Modal will be soon pulling data from there instead of having these hardcoded in.

PR #911, opened 2023-02-26
stripped angular example with html package

bws9000:
sorry about the delay i can only work on this on the weekend - stripped angular added html package in reference to the last PR [867](https://github.com/WalletConnect/web3modal/pull/867)


0xAsimetriq:
Thank you for taking time to update this! Some notes

Core notes
1. React is still used in the example, should not be the case i.e this should be a pure angular example

Overal notes
1. I can still some extra dependencies in the source i.e.`react-dom`, `@wagmi/chains`, react types etc. (this one is part of wagmi/core)
2. Could you remove any extra un-necessary files like editorconfig / tests etc... ideally leaving minimal ammout of these
3. Files like `package-lock` need to be removed in favour of using setup in this monorepo (yarn and there is only one lockfile at the root)
4. Example should be renamed to just `angular`
5. Similarly to other examples, some commands need to be exposed in root package json workspaces i.e. `dev:angular`
6. Remove readme (as per other examples)
7. Update `.gitignore` to de-duplicate entries from root `.gitignore` and only include rules specific to angular and this example
8. All package.json dependencies should be strict i.e no `^` in versions
9. Remove duplicate packages like typescript that are already present in the root of the monorepo, also is `tslib`/ `tsx` needed?
10. `tsconfig.json` should extend root one and only include extra rules / overrides needed for this example
11.  There are some files that are empty

😅 Sorry for the extensive list, but these should be complete to keep this example inline with practices established in a repo and make it simpler for people to get started with.

bws9000:
> 😅 Sorry for the extensive list, but these should be complete to keep this example inline with practices established in a repo and make it simpler for people to get started with.

ok, no problem. i'll work on this

bws9000:
hi so far:

Core notes
1. React is still used in the example, should not be the case i.e this should be a pure angular example
 +done

Overal notes
1. I can still some extra dependencies in the source i.e.`react-dom`, `@wagmi/chains`, react types etc. (this one is part of wagmi/core)
+done

2. Could you remove any extra un-necessary files like editorconfig / tests etc... ideally leaving minimal ammout of these
+done

3. Files like `package-lock` need to be removed in favour of using setup in this monorepo (yarn and there is only one lockfile at the root)
+done

4. Example should be renamed to just `angular`
+done

bws9000:
I had to add a couple duplicates in tsconfig angular, wouldn't build without this config. Could be missing something possible on your list but from root ... yarn, yarn dev:angular and you're good to go

0xAsimetriq:
This is awesome 🔥 will be reviewing during the week, but can already tell there was a lot of cleanup and simplification done, thank you!

0xAsimetriq:
Sorry for taking too long on this, we were super busy with bringing V2 of web3modal to stable and are now focusing on V3 which will have entirely new architecture. Plan is to add more framework support there, thus I'll be closing this pr as it is no longer portable.

PR #913, opened 2023-02-27
feat: add xdefi integration

hinterlist:


0xAsimetriq:
We are moving towards managing injected extensions from https://cloud.walletconnect.com which means that these manual presets will soon be removed. To add your extension wallet into web3modal, please create a project (or add to your existing one in the cloud dashboard).

There will be an option to select injected wallet and provide your uniquely identifiable flag per chain i.e.  `window.ethereum.xfi` as well as download links and assets.

PR #947, opened 2023-03-18
fix: prevent name and icon mismatch for injected wallets

bfriel:
**Bug**
It's currently possible for an injected wallet's name to not match its icon. This is currently affecting Phantom as shown in the **before** images below. 

This PR changes how names for injected wallets are detected. Instead of detecting off of an optimistic name, it detects based on the wallet's `id`. This `id` detection is currently used for wallet icons, so no behavior there is changed.

**Before**
mm1
mm2

**After**
ph1
ph2

**Additional Notes**
I've tested with the following injected wallets to make sure there was no disruption: MetaMask, Coinbase Wallet, Trust Wallet, Phantom, Taho, Core

### DIFF
--- a/packages/ui/src/components/w3m-wallet-button/index.ts
+++ b/packages/ui/src/components/w3m-wallet-button/index.ts
@@ -32,17 +32,27 @@ export class W3mWalletButton extends LitElement {
     return null
   }
 
+  private nameTemplate() {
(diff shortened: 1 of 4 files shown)


0xAsimetriq:
Fixed and released as part of `v2.2.2`

PR #950, opened 2023-03-20
Release/2.2.2

0xAsimetriq:


0xAsimetriq:
Closing in favour of https://github.com/WalletConnect/web3modal/pull/950

We were actually handling id correctly to begin with, but it appears wagmi was not returning `injected` string correctly. Luckily upgrading to latest version seems to have fixed this. There is test deployment available in pr linked above.

PR #961, opened 2023-03-27
feat: expose selectedChain property

timefordroids:
When a user selects a different network in the network selector without being connected to their wallet, there is no way to determine which network they have selected, leading to a case when network in the network selector is Ethereum, but data is showing for BSC. Exposing the selectedChain property will enable the ability to determine which network is selected.


da748:
A review is required to correspond to bug link

timefordroids:
@da748  My bad. Thank you for reminder

0xAsimetriq:
Closing this as we won't be adding any new functionality to v2. Events (including currently selected chain) are planned for upcoming v3.x releases.

PR #963, opened 2023-03-30
Make default chain configurable

linuslagerhjelm:


0xAsimetriq:
This is configurable via web3modal options https://docs.walletconnect.com/2.0/web3modal/options#defaultchain-optional

However, behaviour for default chain will change slightly in [2.3.0](https://github.com/WalletConnect/web3modal/pull/954) to  where we no longer use it unless explicitly provided (will connect to active wallet chain by default)

PR #967, opened 2023-04-01
Remove metamask id check for injected wallets

UTkzhang:
In order to solve https://github.com/WalletConnect/web3modal/issues/892, we tried introducing the Metamask connector by overriding the w3mConnectors in wagmi client. This does work, visually, however due to this additional check for injectedConnector, it does not work for metamask wallet. Wondering why this check was there?

Sample fix on the UI side:
```
export const wagmiClient = createClient({
  autoConnect: true,
  queryClient,
  connectors: () => {
    return uniqBy(
      (connector) => connector.name,
      [
        ...w3mConnectors({
          projectId: WALLET_CONNECT_PROJECT_ID,
          version: 1,
          chains,
        }),
        new MetaMaskConnector(), // Force override due to window.ethereum conflict with coinbase & other wallets
      ],
    );
  },
  provider,
});
```

### DIFF
--- a/packages/ui/src/partials/w3m-desktop-wallet-selection/index.ts
+++ b/packages/ui/src/partials/w3m-desktop-wallet-selection/index.ts
@@ -1,9 +1,8 @@
 import type { DesktopConnectorData } from '@web3modal/core'
 import { ConfigCtrl, ExplorerCtrl, OptionsCtrl, RouterCtrl } from '@web3modal/core'
-import { html, LitElement } from 'lit'
+import { LitElement, html } from 'lit'
 import { customElement } from 'lit/decorators.js'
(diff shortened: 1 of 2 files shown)


UTkzhang:
Followup, might be related to: https://github.com/WalletConnect/web3modal/issues/879?

0xAsimetriq:
Interesting, this might've been left from older days when mm connector was included. Will give it a test drive on Monday and test everything, ty 👍

UTkzhang:
@0xAsimetriq please and thank you, appreciate the quick look! This can be considered "breaking" behaviour for customers hence the urgency :D

0xAsimetriq:
@UTkzhang thank you for this PR, this will be addressed on 2.3.0, but with a slightly different approach due to refactor around injected wallets.

PR #971, opened 2023-04-03
Sped up ENS reverse resolver lookup. Consolidated synchronous registry + reverse resolver queries to single request.

doplur:
Improved time to render the user's wallet "name".

Utilized immutable helper contract published to:  
[https://etherscan.io/address/0xb3bc20cdc6858bef29db785543aab7436ac86c83#code](https://etherscan.io/address/0xb3bc20cdc6858bef29db785543aab7436ac86c83#code)

### DIFF
--- a/chains/ethereum/src/client.ts
+++ b/chains/ethereum/src/client.ts
@@ -4,9 +4,9 @@ import {
   disconnect,
   fetchBalance,
   fetchEnsAvatar,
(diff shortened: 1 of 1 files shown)


0xAsimetriq:
Thank you for this pr, but we are currently working on a different solution for this as well, plus will not be introducing call's to 3rd party contracts.

PR #984, opened 2023-04-17
Update frame logo

goosewobbler:
Updating the Frame logo in W3M v1 to one which works better across both light and dark themes.


0xAsimetriq:
Thank you for this, but we've deprecated v1 branch and there are no plans to release new versions unless they fix a critical issue. For more info on current web3modal version check out https://web3modal.com

goosewobbler:
Ok, I saw the recent releases for v1 and thought that it was in maintenance mode.

PR #985, opened 2023-04-17
Add support for independent qrcode component

devceline:
# Changes
- Export qrcode web component
- Export independent React QRCode component
- Make a `qrcode` page in the labratory to show it off
- Allow optional override for image in `w3m-wallet-image`, this does not affect the public API as this is an internal component
- Allow passing `theme` to `w3m-qrcode` to allow for websites just using the qrcode to be able to pick the theme


0xAsimetriq (commented):
Glanced over, everything looks good, but I think qrcode is implemented too much like being "it's own library" i.e. initiating ui package / taking in custom theming.

If you need that explicit granular control over this component, I think separate lib would make more sense, but if we can align it more with other components like connect button or network one, can keep it exported.

In essence, users should still init modal as usual and pass options for theme there. Then we just export qrcode component (Let's keep naming aligned and call it `W3mQrcode` or something along those lines). In the end it shouldn't init ui library, not have theme override (instead all of this should be inherited from normal flow).

laboratory/src/pages/index.tsx
```diff
@@ -43,6 +43,12 @@ const cards = [
     description: 'Simple ethereum example that uses WalletConnect Auth SDK',
     link: '/v2Auth',
     color: 'success'
+  },
+  {
```
0xAsimetriq, line 47: Let's remove this to keep examples to full modal ones with differen't configs, but new component and it's props should be documented in

https://docs.walletconnect.com/2.0/web3modal/react/components
and
https://docs.walletconnect.com/2.0/web3modal/html-js/components

packages/react/src/components/QrCode.tsx
```diff
@@ -0,0 +1,30 @@
...
+ */
+export function QrCode(props: JSX.IntrinsicElements['w3m-qrcode']) {
+  const { size } = props
+
+  useEffect(() => {
+    import('@web3modal/ui')
+  })
+
+  return (
+    <div style={{ width: size, height: size }}>
```
0xAsimetriq, line 15: Does display break without the div wrapper, or could we remove it?

packages/ui/src/components/w3m-qrcode/index.ts
```diff
@@ -7,18 +7,20 @@ import { ThemeUtil } from '../../utils/ThemeUtil'
...
+export class W3mQrCode extends LitElement {
   public static styles = [ThemeUtil.globalCss, styles]
 
   // -- state & properties ------------------------------------------- //
   @property() public uri = ''
   @property({ type: Number }) public size = 0
   @property() public imageId? = ''
   @property() public walletId? = ''
+  @property() public imageUrl? = ''
+  @property() public theme?: typeof ThemeCtrl.state.themeMode
```
0xAsimetriq, line 19: Per previous comment, let's inherit theme as before

packages/ui/src/components/w3m-qrcode/index.ts
```diff
@@ -31,11 +33,12 @@ export default class W3mQrCode extends LitElement {
   protected render() {
     return html`
       <div>
-        ${this.walletId
+        ${this.walletId || this.imageUrl
           ? html`
               <w3m-wallet-image
                 walletId=${this.walletId}
                 imageId=${this.imageId}
+                imageUrlOverride=${this.imageUrl}
```
0xAsimetriq, line 41: Maybe this prop on wallet image can remain as `imageUrl` as well?

packages/react/src/components/QrCode.tsx
devceline: It breaks since `w3m-qrcode` takes `width: 100%` and I don't want to break existing usecases for it

0xAsimetriq (approved):
Nice, thank you for the changes!

laboratory/src/pages/index.tsx
devceline: Will do

PR #996, opened 2023-04-20
fix: manualWalletsTemplate not loading MobileWallets

justraman:
Fixes the bug where a manually added mobile wallet doesn't appear on the wallet selection screen.


0xAsimetriq:
Thank you 🙏

PR #1002, opened 2023-04-27
Release/2.4.0

### LINKED ISSUE #723: Auth API support
- `@web3modal/auth-react` and `@web3modal/auth-html` packages
- Add api level filter for sdk's
- pre-bundle standalone and auth-client libraries
- create lab / repo examples
- update docs

### LINKED ISSUE #952: [feature] Indicate chains that are not supported by wallet
### Summary

We need to indicate chains that are not supported by wallets (after connected). We can dissable buttons for these and add footer info component.

### LINKED ISSUE #1001: [feature] Allow mobile wallet browsers to indicate default injected connector

As of now, when users visit dapps on mobile in-wallet browsers they are presented full choice of wallets with injected mobile wallet being highlighted as installed. We can further improve on this by allowing such wallet browsers to append `w3mPreferInjected=true` etc. to url and indicate which injected connector to use by default, thus resulting in a more seamless ux (there is no reason to show other wallets to users in these use case).

- Document `w3mPreferInjected=true`
- Implement `w3mPreferInjected=true` handling in modal

0xAsimetriq:
## Breaking Changes
- feat: wagmi v1 support, under the hood ethers is replaced with [viem](https://viem.sh/) and api on wagmi side is changed slightly, see migration guides for [react](https://wagmi.sh/react/migration-guide#1xx-breaking-changes) and [core](https://wagmi.sh/core/migration-guide#1xx-breaking-changes) for more info

## Changes
- feat: Indicate chains that are not supported by connected wallet
- fix: use `:active` vs `:hover` for mobile effects
- feat: Added `w3mPreferInjected` url flag, which will force web3modal to trigger injected wallet by default (if installed). This will be useful for so called "in-wallet" browsers to set if they wish so.
- feat: Added `@web3modal/auth-html` package for simplified auth only flows with w3m in html
- feat: Added `@web3modal/auth-react` package for simplified auth only flows with w3m in react

## Associated Issues
closes #952
closes #1001
closes #723


packages/auth-html/readme.md
```diff
@@ -0,0 +1,21 @@
+> Looking for v1.x readme? It is available on [V1 Branch](https://github.com/WalletConnect/web3modal/tree/V1)
```
ignaciosantise, line 1:
```suggestion
```

packages/core/src/utils/CoreUtil.ts
```diff
@@ -88,6 +90,10 @@ export const CoreUtil = {
     )
   },
 
+  removeWalletConnectDeepLink() {
+    localStorage.removeItem(CoreUtil.WALLETCONNECT_DEEPLINK_CHOICE)
```
ignaciosantise, line 94: localStorage wont work in react-native, but i guess this function will be used in web, right?

0xAsimetriq: Yeh, this functionality is something you will need to replicate in rn using their storage mechanisms 👍

svenvoskamp (commented):
Looks good to me. Only concern I have is that `:active` only works on iOS devices when there is a touch event attached to the element.

PR #1028, opened 2023-05-19
fix: deeplink format error

xwal:
Fix the bug that when the deep link has a path like `celo://wallet`, the format string will become `celo://walletwc?uri=xxx`.


0xAsimetriq:
Thank you for this, it will be addressed in next release, with slightly different fix in #1050

PR #1037, opened 2023-05-30
Initial Playwright testing

chris13524:
This PR adds some Playwright tests in the laboratory directory.

Some tests test basic clicking around. Another test tests an actual paring with React Wallet. There's also a test that opens Zerion.

These are run in GitHub Actions automatically on pull requests and pushes, and also run daily via a cron job schedule.


chris13524:
Closing as we will work on updated code in a new PR

PR #1047, opened 2023-06-05
Move `@web3modal/ethereum` to dependencies

zhuqingguang:
When use `@web3modal/core` in a typescript project, it will throw `Cannot find module '@web3modal/ethereum' or its corresponding type declarations` error because install `@web3modal/core` won't install `@web3modal/ethereum` as it is a dev dependency.

# Breaking Changes

N/A

# Changes

- feat:
- fix:
- chore:

# Associated Issues

closes #...

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -16,11 +16,9 @@
     "typecheck": "tsc --noEmit"
   },
   "dependencies": {
+    "@web3modal/ethereum": "2.4.2",
     "buffer": "6.0.3",
     "valtio": "1.10.5"
-  },
-  "devDependencies": {
-    "@web3modal/ethereum": "2.4.2"
   },
   "keywords": [
     "web3",


0xAsimetriq:
`@web3modal/ethereum` should not be a dependency of core as it is not used directly. It is kept in `devDependencies` as some types are used from it (will soon be removed). If you are using `core` package, it is likely in dev environment, so you should install devDependencies as well (don't use it in production).

zhuqingguang:
> `@web3modal/ethereum` should not be a dependency of core as it is not used directly. It is kept in `devDependencies` as some types are used from it (will soon be removed). If you are using `core` package, it is likely in dev environment, so you should install devDependencies as well (don't use it in production).
If the types in `@web3modal/ethereum` is used, maybe it's appropriate to put it in `peerDependencies`.

0xAsimetriq:
`peerDependencies` are for required dependencies that need to be installed alongside the package when it's used in prod, for example `react` is a common one.

May I ask, how / why are you using `@web3modal/core` package? It is only meant to be used internally in our user facing libraries documented at https://docs.walletconnect.com/2.0/web/about-web3modal not on its own

zhuqingguang:
I am using `@web3modal/standalone` and when i use `tsc` to build, errors occurs.
image

And I installed `@web3modal/ethereum` but another type error occurs related to `@wagmi/connectors`:https://github.com/wagmi-dev/references/issues/323

My project can not be built now.

0xAsimetriq:
I see, you either need to add [skipLibCheck](https://www.typescriptlang.org/tsconfig#skipLibCheck) option to your tsconfig or add `@web3modal/ethereum` as devDependency in your project to resolve this.

zhuqingguang:
`2.4.2` has a bug:
image

This seems to be a BREAK change as `2.4.1` has export `Chain` and `2.4.2` not . This makes application break down.

zhuqingguang:
Is there any solution for type errors in `@wagmi/connectors` ? I am only using `@web3modal/standalone`  for replcing `QRCodeModal` but encounter many errors.

0xAsimetriq:
Ones mentioned above don't work? When you use `@wagmi/connectors` you con't need to install standalone modal separately, it comes with connecotr.

zhuqingguang:
I don't want to use `skipLibCheck: true`.  I will appreciate it if this type error is fixed. Thanks.

0xAsimetriq:
In that case you should add `@web3modal/ethereum` as a dev dependency in your project. There is no type error, it is required during dev if you want to validate library types as well.

zhuqingguang:
I have created a [repo](https://github.com/zhuqingguang/web3modal-test), And I set `skipLibCheck: false`.
Just clone it and run `pnpm i && pnpm build` and you will get type error.

zhuqingguang:
I move `@web3modal/ethereum` from ependencies to devDependencies  and run `pnpm i`. It still throws errors. What can I do for this?

PR #1067, opened 2023-06-15
Refactor and port feature/web3account code to current release

svenvoskamp:
# Changes

- feat: Added a `W3aCtrl` that will act as the intermediary between UI and logic for email login 
- feat: Added a `W3aUtil` with helper functions & variables to support email login

# Associated Issues

closes https://github.com/WalletConnect/web3modal/issues/1065


packages/core/index.ts
```diff
@@ -8,6 +8,7 @@ export { OptionsCtrl } from './src/controllers/OptionsCtrl'
 export { RouterCtrl } from './src/controllers/RouterCtrl'
 export { ThemeCtrl } from './src/controllers/ThemeCtrl'
 export { ToastCtrl } from './src/controllers/ToastCtrl'
+export { W3aCtrl } from './src/controllers/W3aCtrl'
```
0xAsimetriq, line 11: Let's use full naming for clarity i.e `Web3AccountCtrl`

packages/core/src/controllers/W3aCtrl.ts
```diff
@@ -0,0 +1,105 @@
+import { proxy, ref, subscribe as valtioSub } from 'valtio/vanilla'
+import type { W3aCtrlState } from '../types/controllerTypes'
+import { IFRAME_API, IFRAME_ID, IS_SERVER, call, messageHandler } from '../utils/W3aUtil'
+
+// -- initial state ------------------------------------------------ //
+const state = proxy<W3aCtrlState>({
+  iFrameMsgIndex: 0,
+  iFrameReady: false,
+  providerInited: false,
```
0xAsimetriq, line 9: `providerInitialized`

packages/core/src/controllers/W3aCtrl.ts
```diff
@@ -0,0 +1,105 @@
...
+  iFrame: null as unknown as HTMLIFrameElement,
```
0xAsimetriq, line 10: Ui should be completely different from the controllers i.e. think about controllers being portable to other platforms like react-native.

packages/core/src/controllers/W3aCtrl.ts
```diff
@@ -0,0 +1,105 @@
...
+
+// -- controller --------------------------------------------------- //
+export const W3aCtrl = {
+  state,
+
+  subscribe(callback: (newState: W3aCtrlState) => void) {
+    return valtioSub(state, () => callback(state))
+  },
+
+  setIFrame() {
```
0xAsimetriq, line 22: All of this logic should be taken outside of controller and into web3account sdk. We are doing too much here. Role of controller is to be used by ui and call methods on underlying sdk's / manage state. They should be very simple

0xAsimetriq (changes requested):
Alright, I stopped review at the controller level, as we need to re-do few bits:

Make `Web3AccountUtil` a class with internal private fields to track state like it's iframe, initialisation status / index etc. 

`Web3AccountCtrl` is not needed for this first iteration I think. You would take `Web3AccountUtil` and use it to create a wagmi connector. Once wagmi connector is ready, it can be added to `@web3modal/ethereum` package and expose few methods similar to `connectWalletConnect`. One hard thing here is that we have 2 steps for connection, sending and verifying email, so we might need to add new function here to send email and when it is being verified user actually connects. After this, user's should make requests through wagmi methods as with any other connector. In the future we will also add off boarding.

When we get to v3 refactor, we will move things around and use `Web3AccountUtil` lass with ethers and other libs, so keep this in mind when refactoring. It has to be very portable.

0xAsimetriq:
I think we can close this one for now due to changes in the roadmap. Web3Account will look very different on new arch.

PR #1072, opened 2023-06-16
add number type to --w3m-z-index

siosio34:
# Breaking Changes

work : add number type to --w3m-z-index

When I tried to use the web3 modal within the project, there was confusion because the number type zIndex was not specified. If the number type is added to zIndex, it seems to be more convenient to use without confusion. So I did this.

# Changes

- feat:
- fix:  add number type to --w3m-z-index
- chore:

# Associated Issues

closes #...


0xAsimetriq:
Thank you for this, but all css variable values are used downstream in `setPropery` method which only accepts strings https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleDeclaration/setProperty to avoid complexity of checking and converting everything to string, we instead just default to all of these values being strings.

PR #1073, opened 2023-06-16
fix: add chainImages type WalletConnectModalConfig

siosio34:
# Breaking Changes

The QrModalOptions type uses Pick to get the type in WalletConnectModalConfig, but the chainImages type doesn't seem to exist in WalletConnectModalConfig.

So I modified it as below. If I misunderstood, you can close the pr.

# Changes

- feat:
- fix:add chainImages type WalletConnectModalConfig
- chore:

# Associated Issues

closes #...

### DIFF
--- a/packages/walletconnect-modal/src/client.ts
+++ b/packages/walletconnect-modal/src/client.ts
@@ -6,6 +6,7 @@ import { ConfigCtrl, ModalCtrl, OptionsCtrl, ThemeCtrl } from '@web3modal/core'
  */
 export type WalletConnectModalConfig = Pick<
   ConfigCtrlState,
+  | 'chainImages'
   | 'desktopWallets'
   | 'enableAuthMode'
   | 'enableExplorer'


0xAsimetriq:
WalletConnectModal does not support any chain switching views (that is a feature of Web3Modal), hence this option was not picked.

siosio34:
@0xAsimetriq 

Currently I am using wagmi. EthereumProvider is used when connecting Walletconnect in wagmi, and chainImages type is used by using Pick in Web3ModalConfig type in QrModalOptions type of the package, but chainImages type does not exist in Web3ModalConfig type.

If that's the intention, shouldn't the part that uses Pick in the EthereumProvider package to get the chainImages type also need to be modified?

Add an image of the code for better understanding.

image

0xAsimetriq:
There are 2 types of modal:

WalletConnectModal
https://docs.walletconnect.com/2.0/web/walletConnectModal/installation
This one is very basic modal experience for when you only need to connect via WalletConnect, it is used in ethereum-provider. You can use this with wagmi's `WalletConnectConnector` 

Web3Modal
https://docs.walletconnect.com/2.0/web/about-web3modal
This is a managed offering that includes things like account view, chain switching, extension wallets etc. It uses wagmi, but disables modal from `WalletConnectConnector`.

PR #1089, opened 2023-06-20
Adding loading state when disconnecting 

svenvoskamp:
# Changes

- feat: Adding loading state when disconnecting 

# Associated Issues

closes https://github.com/WalletConnect/web3modal/issues/1084


packages/ui/src/components/w3m-box-button/index.ts
```diff
@@ -11,13 +11,14 @@ export class W3mBoxButton extends LitElement {
...
   @property() public label = ''
+  @property() public loading?: boolean = false
   @property() public onClick: () => void = () => null
 
   // -- render ------------------------------------------------------- //
   protected render() {
     return html`
-      <button @click=${this.onClick}>
-        <div>${this.icon}</div>
+      <button @click=${this.onClick} class="${this.loading ? 'disabled' : ''}">
```
0xAsimetriq, line 20: Let's use buttons disabled attribute instead, i.e. `disabled=${this.loading}` (you might need to do something weird with lit-html i.e. `.disabled=`) where that dot is used to boolean check it.

packages/ui/src/components/w3m-box-button/index.ts
```diff
@@ -11,13 +11,14 @@ export class W3mBoxButton extends LitElement {
...
+        <div>${this.loading ? html`<span class="spinner"></span>` : html`${this.icon}`}</div>
```
0xAsimetriq, line 21: There is existing `w3m-spinner` component that can be re-used here. Might need to adjust it, to take in different size. Will also mean we can remove extra css made for spinner.

0xAsimetriq (changes requested):

packages/ui/src/components/w3m-box-button/styles.css
```diff
@@ -32,3 +36,12 @@ button > div {
 button path {
   fill: var(--w3m-accent-fill-color);
 }
+
+@keyframes rotation {
```
0xAsimetriq, line 40: Are rotation keyframes still needed?

packages/ui/src/components/w3m-spinner/index.ts
```diff
@@ -1,17 +1,28 @@
...
+import { customElement, property } from 'lit/decorators.js'
 import { ThemeUtil } from '../../utils/ThemeUtil'
 import styles from './styles.css'
 
 @customElement('w3m-spinner')
 export class W3mSpinner extends LitElement {
   public static styles = [ThemeUtil.globalCss, styles]
 
+  @property() public color = 'accent'
+  @property({ type: Boolean }) public small = false
```
0xAsimetriq, line 11: Maybe it would be more versatile to make this `@property() public size = 24` and allow more granular overrides with numbers?

Alternatively, it could be something like `@property() public size: 'small' | 'medium' = 'medium'`

Both of these allow us to set extra sizes in the future if needed.

packages/ui/src/components/w3m-spinner/styles.css
```diff
@@ -29,5 +29,12 @@ svg {
...
   stroke-linecap: round;
   animation: dash 1.5s ease infinite;
+}
+
+.accent {
   stroke: var(--w3m-accent-color);
 }
+
+.white {
+  stroke: #fff;
```
0xAsimetriq, line 39: This doesn't really work in terms of light / dark modal, plus what if users override colors of a button or its text? Instead of `white` option here, maybe lets make it as "fill" (or whatever terminology is used for color of a text in a button) and use css variable that is used for button text color. This will make things more consistent.

svenvoskamp:
@0xAsimetriq

* Removed keyframes
* Added size option instead of small/medium
* Added fill color with css variable 

Could you review again?

0xAsimetriq (commented):
Thank you for all the code changes, much neater now 💪 
Tested on my phone, everything works as expected, but loader seems to be off center. Perhaps it needs to be a bit bigger? Also, I don't think we need to go as extreme with button .25 opacity, something like .7 should be fine and still keep loading state visible

https://github.com/WalletConnect/web3modal/assets/3154053/d02c89a6-90e7-4422-9951-423c04fa9616

0xAsimetriq (approved):
Re-tested on mobile, looks good now 👍 ty for adjustments

svenvoskamp:
@0xAsimetriq  Last commit was not a great fix. Did some extra change, it works correctly now with the correct options etc. Can you check again?

0xAsimetriq (approved):

PR #1090, opened 2023-06-20
feat: use new identity API for profile resolution

chris13524:
# Breaking Changes

N/A

# Changes

- chore: use new identity API for faster ENS resolution

# Associated Issues

Depends on https://github.com/WalletConnect/rpc-proxy/pull/183


arein:
In a follow up PR please also add a widget to the dashboard so we can see how it performs (status codes/availability and latency and once you use the cache also the cache hit ratio)

packages/core/src/utils/BlockchainApiUtil.ts
```diff
@@ -0,0 +1,26 @@
...
+  async getIdentity(
+    address: string,
+    chainId: number
+  ): Promise<{ name: string; avatar?: string } | null> {
+    const { projectId } = ConfigCtrl.state
+    const chain_id = `${NAMESPACE}:${chainId}`
+    const endpoint = `${BLOCKCHAIN_API}/v1/identity/${address}?chainId=${chain_id}&projectId=${projectId}`
+
+    const response = await fetch(endpoint)
+    if (response.status === 404) {
```
0xAsimetriq, line 19:
```suggestion
    if (response.status !== 200) {
```

And then we can probably just `throw` to trigger fallback. This way you don't need that `if(profile)` check I commented on above anymore.

packages/core/src/controllers/AccountCtrl.ts
```diff
@@ -35,18 +36,31 @@ export const AccountCtrl = {
...
-          const avatar = await ClientCtrl.client().fetchEnsAvatar({ name, chainId: 1 })
-          if (avatar) {
-            await preloadAvatarFn(avatar)
+        try {
+          const profile = await BlockchainApiUtil.getIdentity(address, mainnetId)
+          if (profile) {
+            state.profileName = profile.name
+            state.profileAvatar = profile.avatar
+          } else {
+            // Profile not found
```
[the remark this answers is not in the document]
chris13524: This is intentional as-is. The API returned successfully, but no profile was found.

packages/core/src/utils/BlockchainApiUtil.ts
chris13524: This is intentional as-is. a 404 is a non-error response saying that the profile was not found. A 500 or 400 would `throw` and trigger the `catch` in the calling function.

0xAsimetriq (dismissed):

packages/core/src/utils/BlockchainApiUtil.ts
```diff
@@ -0,0 +1,27 @@
...
```
chris13524, line 18: Actually, since 404 errors log to the console, I think this should be part of the JSON response not a status code. Also, it enables differentiating the route not being found vs profile not being found.

image

chris13524: Fixed by https://github.com/WalletConnect/rpc-proxy/pull/222

arein (approved):

PR #1096, opened 2023-06-26
[V3] Adding wui-text component

svenvoskamp:
# Changes

- feat: Add `wui-text` component 

# Description

We use the following text styles in V3:

Large
Medium
Small
XXSmall
Medium Numerals

Took them from Figma and implemented in the component. 

Link to Storybook: ...


packages/ui/src/utils/ThemeUtil.ts
```diff
@@ -13,6 +13,32 @@ export const globalStyles = css`
...
+
+    --wui-letter-spacing-sm: -0.02em;
+    --wui-letter-spacing-md: -0.03em;
+    --wui-letter-spacing-lg: -0.05em;
+
+    --wui-line-height-sm: 120%;
+    --wui-line-height-md: 125%;
+    --wui-line-height-lg: 130%;
+
+    --wui-color-accent: #47a1ff;
```
0xAsimetriq, line 35: Color names don't seem to match ones in figma's `Library 3.0` file

packages/ui/src/components/wui-text/styles.ts
```diff
@@ -0,0 +1,94 @@
...
+  .wui-xxs-bold {
+    font-size: var(--wui-font-size-xxs);
+    font-weight: var(--wui-font-weight-bold);
+    line-height: var(--wui-line-height-sm);
+    letter-spacing: var(--wui-letter-spacing-sm);
+  }
+
+  .wui-color-accent {
+    color: var(--wui-color-accent);
+  }
```
0xAsimetriq, line 75: Inconsistent spacing after `}`

packages/ui/src/components/wui-text/index.ts
```diff
@@ -0,0 +1,63 @@
...
+      'wui-color-error': this.color === 'error',
+      'wui-color-success': this.color === 'success',
+      'wui-color-inverse': this.color === 'inverse',
+      'wui-color-primary': this.color === 'primary',
+      'wui-color-secondary': this.color === 'secondary',
+      'wui-color-tertiary': this.color === 'tertiary'
+    }
+
+    return html`
+      <span>
```
0xAsimetriq, line 52: Not sure about this, but is it possible to remove span and style `slot` element directly?

packages/ui/src/components/wui-text/index.ts
```diff
@@ -0,0 +1,63 @@
...
+  | 'md-semibold'
+  | 'sm-medium'
+  | 'sm-semibold'
+  | 'xxs-bold'
+
+type Color = 'accent' | 'error' | 'inverse' | 'primary' | 'secondary' | 'tertiary' | 'success'
+
+@customElement('wui-text')
+export class WuiText extends LitElement {
+  public static styles = [globalStyles, styles]
```
0xAsimetriq, line 23: Something to research in `lit`, do you think we can find a better solution to global styles i.e. instead of defining them in each component, maybe we can define one time somewhere?

0xAsimetriq (changes requested):
Left few comments

svenvoskamp:
@0xAsimetriq 

- Renamed colors
- Inconsistent spaces fixed
- Removed span in `wui-text`
- Did some research on global `css` styles for `lit` but couldn't find a better solution than what we currently have.

0xAsimetriq (approved):
Looks good to me 👍

apps/gallery/.storybook/preview.tsx
```diff
@@ -0,0 +1,32 @@
+import { Controls, Description, Primary, Source, Stories, Subtitle, Title } from '@storybook/blocks'
+import React from 'react'
+
+import {} from '@storybook/blocks'
```
0xAsimetriq, line 4: Is this import needed?

apps/gallery/.storybook/preview.tsx
```diff
@@ -0,0 +1,32 @@
+import { Controls, Description, Primary, Source, Stories, Subtitle, Title } from '@storybook/blocks'
+import React from 'react'
```
0xAsimetriq, line 2: Is react import needed? (might be global like in nextjs)

packages/ui/src/components/wui-text/index.ts
```diff
@@ -0,0 +1,71 @@
...
+      'wui-color-success-100': this.color === 'success-100',
+      'wui-color-inverse-100': this.color === 'inverse-100',
+      'wui-color-inverse-000': this.color === 'inverse-000',
+      'wui-color-fg-300': this.color === 'fg-300',
+      'wui-color-fg-200': this.color === 'fg-200',
+      'wui-color-fg-100': this.color === 'fg-100'
+    }
+
+    return html`
+      <span>
```
0xAsimetriq, line 60: Didn't we remove span? styles file seems to target host now, so is this span element needed?

packages/ui/src/utils/ThemeUtil.ts
```diff
@@ -13,6 +14,33 @@ export const globalStyles = css`
     -moz-osx-font-smoothing: grayscale;
     -webkit-tap-highlight-color: transparent;
     backface-visibility: hidden;
+
+    --wui-font-family: 'SF Pro Text', 'Segoe UI', Roboto, Ubuntu, 'Helvetica Neue', sans-serif;
```
0xAsimetriq, line 18: I think variables should be moved to `:root {}`

apps/gallery/.storybook/preview.tsx
svenvoskamp: Yes it is needed, otherwise you get an `ReferenceError: React is not defined` error

PR #1100, opened 2023-06-27
Add wui-icon & wui-icon-box

svenvoskamp:
# Changes

- feat: Add `wui-icon` & `wui-icon-box`

Storybook link: https://web3modal-gallery-git-v-3-wui-icon-walletconnect1.vercel.app/?path=/docs/components-wui-icon-box--docs


packages/ui/src/components/wui-icon-box/index.ts
```diff
@@ -0,0 +1,68 @@
...
+      default:
+        iconSize = 'xxs'
+    }
+
+    const backgroundClasses = {
+      [`wui-opacity-${this.size === 'lg' ? 'sm' : 'md'}`]: true,
+      [`wui-color-${this.backgroundColor}`]: true
+    }
+
+    const sizeClasses = {
```
0xAsimetriq, line 51: This can be a constant string and used on elements without `classMap`

packages/ui/src/components/wui-icon-box/index.ts
```diff
@@ -0,0 +1,68 @@
...
+      [`wui-size-${this.size}`]: true
+    }
+
+    return html`
+      <div class="${classMap(sizeClasses)}">
+        <div class="${classMap(backgroundClasses)}"></div>
```
0xAsimetriq, line 57: Why do we need extra absolute element here? This component seems like it can be acheived via following composition

```
<div>
  <wui-icon></wui-icon>
</div>
```

0xAsimetriq (changes requested):

packages/ui/src/components/wui-icon/index.ts
```diff
@@ -0,0 +1,50 @@
...
+    const sizeClasses = {
+      [`wui-size-${this.size}`]: true
+    }
+
+    const colorClasses = {
+      [`wui-color-${this.color}`]: true
+    }
+
+    return html`
+      <i class=${classMap(sizeClasses)}>
```
0xAsimetriq, line 39: So we need `i` element here? As slot is just <svg /> we can style it directly? All icons should have same aspect ratio of `1:1` and fit svg within it as per figma. So we prob don't need a wrapper here with display flex?

0xAsimetriq (changes requested):

packages/ui/src/components/wui-icon-box/index.ts
svenvoskamp: We need the extra element, because we have to set an `opacity` on the `background`. There is a way to make it work with `rgba()`, but then we need to change our color config.

svenvoskamp:
@0xAsimetriq 

- docs and centered layout by default
- `PresetUtil` added to /gallery/utils
- `TypesUtil` added to /ui/src/utils
- `svg` formatted and `xmlns` removed
- Code cleanup components & stories 
- Update branch with V3 branch 


Could you review again?

packages/ui/src/components/wui-icon-box/index.ts
```diff
@@ -0,0 +1,55 @@
+import { html, LitElement } from 'lit'
```
0xAsimetriq, line 1: wui-icon-box should be a "composite" as it uses icon component. (we can probably get rid of example composite as well).

0xAsimetriq (changes requested):

0xAsimetriq:
Screenshot 2023-06-27 at 17 01 35
Let's keep naming consistent i.e. all of these have `Default` but box icon says `Copy`

svenvoskamp:
@0xAsimetriq 

- Added `color-mix` solution
- Moved `wui-icon-box` to `composites`
- Removed `composites` example

Could you review again?

apps/gallery/stories/components/wui-icon-box.stories.ts
```diff
@@ -0,0 +1,57 @@
+import type { Meta } from '@storybook/web-components'
```
0xAsimetriq, line 1: This entire file (wui-icon-box.stories.ts) file needs to move to `stories/composites`

apps/gallery/stories/components/wui-icon-box.stories.ts
```diff
@@ -0,0 +1,57 @@
...
+import { clockSvg } from '@web3modal/ui/src/assets/svg/clock'
```
0xAsimetriq, line 2: I think a separate story is needed to display all svg's in a sort of list / grid. Then for icon components let's just use 2 as preview. Updating these whenever you add more icons in each component will be too big of a chore. Same can be applied to other stories like icon component

packages/ui/src/composites/wui-icon-box/index.ts
```diff
@@ -0,0 +1,54 @@
...
+        iconSize = 'xxs'
+    }
+
+    const sizeClass = `wui-size-${this.size}`
+    const opacityValue = this.size === 'lg' ? '12%' : '16%'
+
+    return html`
+      <div
+        class="${sizeClass}"
+        style="background-color: var(--wui-rgba-transparent);
```
0xAsimetriq, line 40: Let's avoid inline styles. Do we really need to differentiate % of opacity between different sizes? Perhaps we just use one middleground value for all? i.e. `14%`. Should confirm with design team first. If they absolutely do need different opacity values here, we can just set default bg color in css for `div` and add 2 classes for diff opacities below that with bg color mix.

packages/ui/src/utils/ThemeUtil.ts
```diff
@@ -41,6 +41,8 @@ export const globalStyles = css`
     --wui-color-fg-100: #e4e7e7;
     --wui-color-fg-200: #949e9e;
     --wui-color-fg-300: #6e7777;
+
+    --wui-rgba-transparent: rgba(255, 255, 255, 0.2);
```
0xAsimetriq, line 45: Not too descriptive, perhaps something like `--wui-color-translucent` ? This one will also likely end up being used in few other components like toast notification perhaps.

svenvoskamp: Renamed it to `wui-overlay-020`, because we will use this for more components.

0xAsimetriq: good naming idea 👍 
`--wui-color-overlay-20` then perhaps? Just to make it clearer

packages/ui/src/composites/wui-icon-box/index.ts
svenvoskamp: I agree that inline styles are not the best option, but even if we have only 1 opacity value, we still need to create a class for every color. 

```
 .wui-bg-blue-100 {
    background-color: color-mix(in srgb, var(--wui-color-blue-100) 14%, transparent);
  }

.wui-bg-fg-100 {
    background-color: color-mix(in srgb, var(--wui-color-fg-100) 14%, transparent);
  }
```

My preferred solution is to use inline styles only to create a css variable:

```
<div
  class="${classMap(classes)}"
  style="--wui-bg-value: var(--wui-color-${this.backgroundColor});"
>
  <wui-icon color=${this.iconColor} size=${iconSize}><slot></slot></wui-icon>
</div>
```

Then we only have to create 2 classes and don't break the design: 

  ```
.wui-overlay-sm {
    background-color: color-mix(in srgb, var(--wui-bg-value) 12%, transparent);
  }

  .wui-overlay-md {
    background-color: color-mix(in srgb, var(--wui-bg-value) 16%, transparent);
  }
```

Thoughts @0xAsimetriq ?

0xAsimetriq (approved):

PR #1105, opened 2023-06-28
V 3/wui button

svenvoskamp:
# Changes

- feat: Add `wui-button` component
- feat: Added `css` to storybook, so background for stories is always dark (better for testing purposes as we are making the components for dark mode). In the future we can make this dynamic with a dark/light mode toggle.


0xAsimetriq:
Instead of adding css to storybook and setting it to dark background, is it possible to make storybook to default to "dark" mode instead?

apps/gallery/.storybook/preview.tsx
```diff
@@ -1,6 +1,8 @@
 import React from 'react'
 import { Controls, Description, Primary, Source, Stories, Subtitle, Title } from '@storybook/blocks'
 
+import '../css/index.css'
```
0xAsimetriq, line 4: Is it possible to set storybook to dark mode by default instead, thus avoiding its customisation (let's try to keep it as out of the box as possible so we don't spend too much time on it atm and can easily upgrade in the future in case they change classes)

packages/ui/src/components/wui-button/index.ts
```diff
@@ -0,0 +1,76 @@
+import type { TemplateResult } from 'lit'
```
0xAsimetriq, line 1: Button uses text and icon components, thus should be a `composite` instead. (same for it's story)

svenvoskamp:
@0xAsimetriq 

- Moved `wui-button` to `composites`
- Removed css from storybook and added dark mode plugin
- Storybook has dark mode by default 

Tried some things to change the preview background color based on the theme toggle without touching css, but couldn't make it work.

apps/gallery/.storybook/main.ts
```diff
@@ -1,7 +1,7 @@
 /** @type { import('@storybook/web-components-vite').StorybookConfig } */
 const config = {
   stories: ['../stories/**/*.mdx', '../stories/**/*.stories.@(js|jsx|ts|tsx)'],
-  addons: ['@storybook/addon-links', '@storybook/addon-essentials'],
+  addons: ['@storybook/addon-links', '@storybook/addon-essentials', 'storybook-dark-mode'],
```
0xAsimetriq, line 4: Instead of using an adon from a third-party, can we utalise default storybook packages / manager api for this? https://storybook.js.org/docs/web-components/configure/theming

0xAsimetriq (changes requested):

apps/gallery/.storybook/main.ts
svenvoskamp: I removed the dark mode addon. Dark mode is enabled by default and we can switch theme in the toolbar above. Would be nice if we can hook the `ThemeUtil` to this switch in the future.

0xAsimetriq (approved):

PR #1119, opened 2023-06-29
Add wui-wallet-image composite

svenvoskamp:
# Changes

- feat: Add `wui-wallet-image` composite


packages/ui/src/composites/wui-wallet-image/index.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  @property() public walletName = ''
+
+  @property() public src?: string
+
+  // -- render ------------------------------------------------------- //
+  public render() {
+    const sizeClass = `wui-size-${this.size}`
+
+    function isImageUrl(url: string): boolean {
```
0xAsimetriq, line 25: In my opinion, un-necessary complication, there are other formats we might use like avif etc.. Just checking for empty `this.src` will be enough here. We actually want to see something going wrong if we provide non image url by accident.

packages/ui/src/composites/wui-wallet-image/styles.ts
```diff
@@ -0,0 +1,56 @@
...
+    position: relative;
+    border-radius: 12px;
+    overflow: hidden;
+    background: var(--wui-overlay-002);
+    display: flex;
+    justify-content: center;
+    align-items: center;
+  }
+
+  div::after {
```
0xAsimetriq, line 14: When using stylistic pseudo elements like this that overlay original, it's good practice to set `pointer-events: none` just in case. There might be cases where it covers something up that is interactive. I'm not sure, but could be that in this case it prevents user from right clicking and saving an image or something?

0xAsimetriq (approved):
Removal of src validation also fixed image not displaying (our api doesn't use file extensions), approved 👍

0xAsimetriq (approved):

PR #1120, opened 2023-06-29
V 3/wui all wallet image

svenvoskamp:
# Changes

- feat: Add `wui-all-wallets-image` composite


apps/gallery/stories/composites/wui-all-wallets-image.stories.ts
```diff
@@ -0,0 +1,40 @@
+import type { Meta } from '@storybook/web-components'
+import type {
+  WalletImage,
+  WuiAllWalletsImage
+} from '@web3modal/ui/src/composites/wui-all-wallets-image'
+import '@web3modal/ui/src/composites/wui-all-wallets-image'
+import '@web3modal/ui/src/composites/wui-wallet-image'
```
0xAsimetriq, line 7: Is standard wallet image import needed here?

apps/gallery/stories/composites/wui-all-wallets-image.stories.ts
```diff
@@ -0,0 +1,40 @@
...
+      },
+      {
+        src: 'https://explorer-api.walletconnect.com/w3m/v1/getWalletImage/a7f416de-aa03-4c5e-3280-ab49269aef00?projectId=f63a70d1bd9dd78b5eb556c25d11cf05',
+        walletName: 'Ledger'
+      },
+      {
+        src: 'https://explorer-api.walletconnect.com/w3m/v1/getWalletImage/7e1514ba-932d-415d-1bdb-bccb6c2cbc00?projectId=f63a70d1bd9dd78b5eb556c25d11cf05',
+        walletName: 'Fireblocks'
+      }
+    ] as WalletImage[]
```
0xAsimetriq, line 33: I think this should be inherited from `as Component`? What was typescript saying here that make you add this assertion? Does `as const` work instead? if this works we don't need to export / import `WalletImage` type, as this is currently un-common along other components.

packages/ui/src/composites/wui-all-wallets-image/index.ts
```diff
@@ -0,0 +1,36 @@
...
+  // -- state & properties ------------------------------------------- //
+  @property({ type: Array }) public walletImages: WalletImage[] = []
+
+  // -- render ------------------------------------------------------- //
+  public render() {
+    const walletImagesHtml = this.walletImages.map(walletImage =>
+      this.renderWalletImage(walletImage.src, walletImage.walletName)
+    )
+
+    return html` <div>${walletImagesHtml}</div> `
```
0xAsimetriq, line 24: Do we need div here, perhaps we can style `:host` element directly? If so, this return can be simplified to

```
return this.walletImages.map(({src, alt})=> html`<wui-wallet-image size="inherit" src=${src} alt=${alt}></wui-wallet-image>`)
```

packages/ui/src/composites/wui-all-wallets-image/index.ts
```diff
@@ -0,0 +1,36 @@
...
+  }
+
+  private renderWalletImage(src: string, alt: string) {
+    return html` <wui-wallet-image size="inherit" src=${src} alt=${alt}></wui-wallet-image> `
```
0xAsimetriq, line 28:
```suggestion
    return html`<wui-wallet-image size="inherit" src=${src} alt=${alt}></wui-wallet-image>`
```

packages/ui/src/composites/wui-all-wallets-image/index.ts
0xAsimetriq, line 24:
```suggestion
    return html`<div>${walletImagesHtml}</div>`
```

0xAsimetriq (changes requested):

apps/gallery/stories/composites/wui-all-wallets-image.stories.ts
svenvoskamp: Yes, you are right, it is inherited from `as Component` so not needed.

0xAsimetriq (approved):

PR #1122, opened 2023-06-30
V 3/ Add wui-flex component

svenvoskamp:
# Changes

- feat: Add `wui-flex` component + `wui-spacing` css variables


packages/ui/src/layout/wui-flex/index.ts
```diff
@@ -0,0 +1,68 @@
...
+  @property() public gap?: Spacing
+
+  @property() public width?: string
+
+  @property() public height?: string
+
+  @property() public padding?: string
+
+  public render() {
+    const inlineStyles = {
```
0xAsimetriq, line 42: Too many options here, flex component should only be responsible for layout / flex functionality, so we need to get rid of: `backgroundColor`, `borderRadius`, `padding`, `width`, `height`

packages/ui/src/layout/wui-flex/index.ts
```diff
@@ -0,0 +1,68 @@
...
+      justifyContent: this.justifyContent,
+      columnGap: this.columnGap && `var(--wui-spacing-${this.columnGap})`,
+      rowGap: this.rowGap && `var(--wui-spacing-${this.rowGap})`,
+      gap: this.gap && `var(--wui-spacing-${this.gap})`,
+      padding: this.padding && `var(--wui-spacing-${this.padding})`,
+      width: this.width,
+      height: this.height
+    }
+
+    return html`<div style="${styleMap(inlineStyles)}"><slot></slot></div>`
```
0xAsimetriq, line 60: I think we can get rid of div here and style `:host` directly in css. You should be able to do `this.styles = styleMap(inlineStyles)` inside render, to apply it directly to host as well. If not we can keep the div 👍

packages/ui/src/layout/wui-flex/index.ts
```diff
@@ -0,0 +1,68 @@
...
+
+@customElement('wui-flex')
+export class WuiFlex extends LitElement {
+  public static styles = [styles]
+
+  @property() public backgroundColor?: Color
+
+  @property() public borderRadius?: BorderRadius
+
+  @property() public flexDirection?: string
```
0xAsimetriq, line 15: Instead of these being just strings, any chance we can hook into actual style types for each, to get correct auto-complete? If not, can we define them in type util and use here.

I saw you did something like this in preset util, so maybe types can be derived from there?

apps/gallery/stories/layout/wui-flex.stories.ts
```diff
@@ -0,0 +1,92 @@
...
+      .columnGap=${args.columnGap}
+      .rowGap=${args.rowGap}
+      .gap=${args.gap}
+      width="400px"
+      height="250px"
+      padding="m"
+      backgroundColor="fg-300"
+      borderRadius="m"
+    >
+      <wui-flex width="40px" height="40px" backgroundColor="blue-100" borderRadius="3xl"></wui-flex>
```
0xAsimetriq, line 83: Let's create a `components` folder in "gallery", for some "dumb" ones like a div which can be a different color / size and use those for preview purposes.

Also, for layout preview, maybe we can make them square (displays layout edges better than circle)

0xAsimetriq (changes requested):

packages/ui/src/layout/wui-flex/index.ts
svenvoskamp: Couldn't get this to work, so I kept the div.

svenvoskamp:
@0xAsimetriq 

* Added test component
* Removed `backgroundColor`, `borderRadius`, `padding`, `width`, `height` from `wui-flex`
* Added types to properties of `wui-flex`

Could you review again?

apps/gallery/components/wui-gallery-div/index.ts
```diff
@@ -0,0 +1,32 @@
+import { html, LitElement } from 'lit'
```
0xAsimetriq, line 1: To avoid confusion with ui package, let's not use `wui-` prefix, nor any imports from web3modal/ui. Storybooks components should be their own, completely detached instances. Similarly how `laboratory` will have its own things as well.

Better name like `gallery-placeholder` could be more useful.

apps/gallery/components/wui-gallery-div/styles.ts
```diff
@@ -0,0 +1,33 @@
+import { css } from 'lit'
```
0xAsimetriq, line 1: Same as above, no reference and completely decouple from wui

packages/ui/src/utils/TypesUtil.ts
```diff
@@ -10,3 +10,27 @@ export type Color =
   | 'success-100'
 
 export type Size = 'inherit' | 'lg' | 'md' | 'sm' | 'xs' | 'xxs'
+
+export type Spacing = '3xl' | '3xs' | '4xs' | 'l' | 'm' | 's' | 'xl' | 'xs' | 'xxl' | 'xxs'
+
+export type BorderRadius = Exclude<Spacing, 'xl' | 'xxl'>
+
+export type FlexDirection = 'column-reverse' | 'column' | 'row-reverse' | 'row'
```
0xAsimetriq, line 18: We can improve all of these and avoid discrepancies with PresetsUtil by doing something like

```
export type FlexWrap = typeof flexWrapOptions[number]
```

P.S for this to work you might need to do something like `flexWrapOptions as const`

Here is related stackoverflow post https://stackoverflow.com/a/54061487/911930

packages/ui/src/layout/wui-flex/styles.ts
```diff
@@ -0,0 +1,7 @@
+import { css } from 'lit'
+
+export default css`
+  div {
+    display: flex;
```
0xAsimetriq, line 5: As we are keeping div in, I think we need to add `width: inherit` and `height: inherit` here, so that we can add styles on `wui-flex` from other places.

0xAsimetriq (changes requested):

0xAsimetriq (approved):
Amazing 🔥

PR #1124, opened 2023-06-30
V 3/Add wui-grid component

svenvoskamp:
# Changes

- feat: Add `wui-grid` component


apps/gallery/utils/PresetUtils.ts
```diff
@@ -61,17 +63,29 @@ export const justifyContentOptions: JustifyContentType[] = [
...
 export const spacingOptions: SpacingType[] = [
-  '3xl',
-  '3xs',
   '4xs',
-  'l',
-  'm',
+  '3xs',
+  'xxs',
+  'xs',
+  'xxs',
```
0xAsimetriq, line 85: `xxs` looks to be repeated

0xAsimetriq (changes requested):
Left one comment, also have you tested all of the properties? I can't seem to get things like column / row gap working for some reason in storybook. Noticed similar on flex element. It could be that it is being overpowered by other styles like gap.

Maybe there is a way in storybook to remove certain properties when others are set to showcase them all?

svenvoskamp:
- When gap is set, it overrules the `column-gap` and `row-gap`. I've added an `undefined` option in Storybook for spacing properties.
- Removed duplicate value

0xAsimetriq (approved):

PR #1127, opened 2023-06-30
V 3/wui text refactor

svenvoskamp:
# Changes

- chore: `wui-text` refactor according to new figma text styles


0xAsimetriq (approved):
Nice, new naming is cleaner as well!

PR #1128, opened 2023-06-30
V 3/wui list select

svenvoskamp:
# Changes

- feat: Add `wui-list-select` component


packages/ui/src/composites/wui-list-select/index.ts
```diff
@@ -0,0 +1,63 @@
...
+@customElement('wui-list-select')
+export class WuiListSelect extends LitElement {
+  public static styles = [globalStyles, styles]
+
+  // -- state & properties ------------------------------------------- //
+  @property({ type: Array }) public walletImages?: IWalletImage[] = []
+
+  @property() public imageSrc? = ''
+
+  @property() public listName = ''
```
0xAsimetriq, line 19: Since component already is named with 'list' in it, we don't need to repeat that in props i.e. this can be just `name`

packages/ui/src/composites/wui-list-select/index.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  @property() public listStatus?: string
```
0xAsimetriq, line 21: same as above, can be just `status`. As we only have a handful of statuses, perhaps we can make this more strict instead of `string` i.e specific status like installed | recent etc.

packages/ui/src/composites/wui-list-select/index.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  @property({ type: Boolean }) public showAllWallets = false
```
0xAsimetriq, line 23: There is no need to for this as we can check if `walletImages` is undefined i.e. further down below:

```
if (this.showAllWallets && this.walletImages) {
 throw new Error('Either showAllWallets or walletImages can be defined, not both')
}

// ...

{this.showAllWallets && <wui-all-wallet....}
{this.imageSrc && <wui-wallet-imag...}
```

packages/ui/src/composites/wui-list-select/index.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  // -- render ------------------------------------------------------- //
+  public render() {
+    const getWalletComponent = () => {
```
0xAsimetriq, line 27: If you adjust as per previous comment, no need to use if/else here and return undefined

packages/ui/src/composites/wui-list-select/index.ts
```diff
@@ -0,0 +1,63 @@
...
+      return undefined
+    }
+
+    const status =
+      this.listStatus &&
+      html` <wui-text variant="micro-700" color="fg-300">${this.listStatus}</wui-text> `
+
+    return html`
+      <button>
+        <div>
```
0xAsimetriq, line 49: Can we avoid using div here, by making `<wui-text>` for name (as these can be the longest) occupy all remaining available space in flex, thus pushing status label to the end?

0xAsimetriq (changes requested):
Could you re-ping me for another review once your previous pr for text changes is merged and applied, I believe it is also included here.

packages/ui/src/composites/wui-list-select/index.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  @property() public name = ''
+
+  @property() public status?: 'installed' | 'qr code' | 'recent'
+
+  @property({ type: Boolean }) public showAllWallets = false
+
+  // -- render ------------------------------------------------------- //
+  public render() {
+    const visual =
```
0xAsimetriq, line 27: When you have sub-components like this, best create separate methods like:

```ts
private visualTemplate() {
  // ...
  return html`...`
}
```

And then use them inside `render()`. Makes things easier to follow and understand when scanning through code after some time when returning to it.

0xAsimetriq: Further down below, you have a relatively complex ternary i.e. `? ... : ... &&`. Those are also hard to follow sometimes, I'd suggest you take out all logic for rendering those into new template functions I suggested above and basically check for all conditions to match before you return `html` or instead just return `null`

This will simplify main render to just

```
// ...
${this.templateAllWallets()}
${this.templateWalletImage()}
// ..
```

packages/ui/src/composites/wui-list-select/styles.ts
```diff
@@ -0,0 +1,37 @@
+import { css } from 'lit'
+
+export default css`
+  button {
+    display: flex;
+    column-gap: var(--wui-spacing-s);
+    align-items: center;
+    width: 336px;
```
0xAsimetriq, line 8: If I understand correctly idea for this component is to occupy full width of its parent in the modal right? So setting static width like this should be avoided.

0xAsimetriq (changes requested):

packages/ui/src/composites/wui-list-select/index.ts
```diff
@@ -0,0 +1,71 @@
...
+      <button>
+        ${this.templateAllWallets()} ${this.templateWalletImage()}
+        <wui-text variant="paragraph-500" color="fg-100">${this.name}</wui-text>
+        ${this.templateStatus()}
+      </button>
+    `
+  }
+
+  private templateAllWallets() {
+    if (this.showAllWallets && this.walletImages) {
```
0xAsimetriq, line 37: We need to account for less that 4 walletImages being passed here. There is some logic in V2 to use images that were given and replace missing ones with wallet placeholder graphics.

0xAsimetriq (changes requested):

svenvoskamp:
- Fallback logic added
- Disabled state added

apps/gallery/stories/composites/wui-list-select.stories.ts
```diff
@@ -0,0 +1,40 @@
...
+import { walletImagesOptions } from '../../utils/PresetUtils'
+
+type Component = Meta<WuiListSelect>
+
+export default {
+  title: 'Composites/wui-list-select',
+  args: {
+    walletImages: walletImagesOptions,
+    imageSrc:
+      'https://explorer-api.walletconnect.com/w3m/v1/getWalletImage/7a33d7f1-3d12-4b5c-f3ee-5cd83cb1b500?projectId=a8d876c6f91c3748db621583fad358f1',
```
0xAsimetriq, line 15:
```suggestion
    imageSrc: walletImagesOptions[0].src
```

packages/ui/src/composites/wui-all-wallets-image/index.ts
```diff
@@ -1,31 +1,39 @@
...
 
 @customElement('wui-all-wallets-image')
 export class WuiAllWalletsImage extends LitElement {
   public static styles = [globalStyles, styles]
 
   // -- state & properties ------------------------------------------- //
-  @property({ type: Array }) public walletImages: WalletImage[] = []
+  @property({ type: Array }) public walletImages: IWalletImage[] = []
+
+  @state() public totalImages = 4
```
0xAsimetriq, line 15: We should try to avoid using state in ui package, you can define constants outside of class via simple `const TOTAL_IMAGES = 4` or if you need to latter update this value, you can make it a private class one i.e. `private totalImages = 4`.

State should be used only when we need to trigger component's re-render from it self, which in case of ui we want to avoid, because we want scaffold package to control this logic through `@properties`

packages/ui/src/composites/wui-all-wallets-image/index.ts
```diff
@@ -1,31 +1,39 @@
...
-          size="inherit"
-          src=${src}
-          walletName=${walletName}
-        ></wui-wallet-image>`
-    )
+    return html`${this.templateWalletImages()}`
+  }
+
+  private templateWalletImages() {
+    const renderedImages = this.walletImages
```
0xAsimetriq, line 23: We can simplify this i.e. loop through existing images and add them, then add placeholders for everything else, so

```
${renderedImages.map( w => html`...`)}
${[...Array(TOTAL_IMAGES - renderedImages.length)].map(w => html`...placeholder`)}
```

0xAsimetriq (changes requested):

0xAsimetriq (approved):

PR #1137, opened 2023-07-03
V3/ Add wui-card-select component

svenvoskamp:
# Changes

- feat: Add `wui-card-select` component

### DIFF
--- a/apps/gallery/stories/composites/wui-card-select.stories.ts
+++ b/apps/gallery/stories/composites/wui-card-select.stories.ts
@@ -3,14 +3,14 @@ import '@web3modal/ui/src/composites/wui-card-select'
 import '../../components/gallery-container'
 import type { WuiCardSelect } from '@web3modal/ui/src/composites/wui-card-select'
 import { html } from 'lit'
+import { walletImageSrc } from '../../utils/PresetUtils'
 
 type Component = Meta<WuiCardSelect>
 
 export default {
   title: 'Composites/wui-card-select',
   args: {
-    imageSrc:
-      'https://explorer-api.walletconnect.com/w3m/v1/getWalletImage/7a33d7f1-3d12-4b5c-f3ee-5cd83cb1b500?projectId=a8d876c6f91c3748db621583fad358f1',
+    imageSrc: walletImageSrc,
     name: 'Rainbow',
     disabled: false
   }
--- a/apps/gallery/stories/composites/wui-card-select.stories.ts
+++ b/apps/gallery/stories/composites/wui-card-select.stories.ts
@@ -1,0 +1,29 @@
+import type { Meta } from '@storybook/web-components'
+import '@web3modal/ui/src/composites/wui-card-select'
+import '../../components/gallery-container'
+import type { WuiCardSelect } from '@web3modal/ui/src/composites/wui-card-select'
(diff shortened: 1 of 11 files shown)


packages/ui/src/composites/wui-card-select/index.ts
```diff
@@ -0,0 +1,43 @@
...
+  public render() {
+    return html`
+      <button ?disabled=${this.disabled}>
+        <wui-wallet-image size="md" src=${this.imageSrc} alt=${this.name}></wui-wallet-image>
+        <wui-text variant="tiny-500" color="inherit">${this.getTruncatedName()}</wui-text>
+      </button>
+    `
+  }
+
+  private getTruncatedName() {
```
0xAsimetriq, line 30: This is a good candidate to make into utility inside ui package i.e. `HelperUtil` or `ToolsUtil`, you can reference how v2 does some of these files. Alternatively, can we acheive this with css? i.e https://css-tricks.com/snippets/css/truncate-string-with-ellipsis/

packages/ui/src/composites/wui-card-select/styles.ts
```diff
@@ -0,0 +1,50 @@
+import { css } from 'lit'
+
+export default css`
+  button {
+    display: flex;
+    flex-direction: column;
+    width: 100%;
```
0xAsimetriq, line 7: Would this component be responsible or does it actually have fixed width? We need to check with design team, either this button expands or stays constant and only spacing / gaps etc around it do

packages/ui/src/composites/wui-card-select/styles.ts
```diff
@@ -0,0 +1,50 @@
+import { css } from 'lit'
+
+export default css`
+  button {
```
0xAsimetriq, line 4: I've noticed that we have quiet a few repeated styles on components that use buttons atm, would it make sense to put them under global styles on a `button`? If not all buttons need this, we could make a global class like `wui-button-base` or something and apply it where needed. 

Similarly for button focus / hover states etc.

apps/gallery/stories/composites/wui-card-select.stories.ts
```diff
@@ -0,0 +1,29 @@
...
+import '../../components/gallery-container'
+import type { WuiCardSelect } from '@web3modal/ui/src/composites/wui-card-select'
+import { html } from 'lit'
+
+type Component = Meta<WuiCardSelect>
+
+export default {
+  title: 'Composites/wui-card-select',
+  args: {
+    imageSrc:
```
0xAsimetriq, line 12: Could reuse that `walletImagePresets[0].src` trick from last pr here as well, maybe there are other places?

0xAsimetriq (changes requested):
Make sure to update with latest V3 as well, there are new dangerjs rules that should catch one other issue here where new components need to be added to `packages/ui/index.ts` file. Would be good to test it before fixing. (just re-run pr check when updated)

packages/ui/src/composites/wui-card-select/index.ts
svenvoskamp: Went for the `css` solution. Cleaner imo.

0xAsimetriq (approved):

PR #1147, opened 2023-07-04
add wui-input

svenvoskamp:
# Changes

- feat: Add `wui-input` component


packages/ui/src/composites/wui-input/index.ts
```diff
@@ -0,0 +1,47 @@
...
+    return html`<input
+        class=${sizeClass}
+        ?disabled=${this.disabled}
+        placeholder=${this.placeholder}
+      />
+      ${this.templateIcon()}`
+  }
+
+  private templateIcon() {
+    const inputSize = this.size === 'md' ? 'lg' : 'md'
```
0xAsimetriq, line 34:
```suggestion
    const iconSize = this.size === 'md' ? 'lg' : 'md'
```

packages/ui/src/composites/wui-input/styles.ts
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  @media (hover: hover) and (pointer: fine) {
+    input:hover:enabled {
+      background-color: var(--wui-overlay-010);
+    }
+  }
+
+  wui-icon {
+    position: absolute;
```
0xAsimetriq, line 51: let's set `pointer-events: none` on the icon, as clicking on it at the moment doesn't focus the input

packages/ui/src/composites/wui-input/styles.ts
```diff
@@ -0,0 +1,72 @@
...
+  input {
+    width: 100%;
+    border-radius: var(--wui-border-radius-xxs);
+    border: 1px solid var(--wui-overlay-005);
+    background: var(--wui-overlay-005);
+    font-size: var(--wui-font-size-paragraph);
+    font-weight: var(--wui-font-weight-regular);
+    letter-spacing: var(--wui-letter-spacing-paragraph);
+    color: var(--wui-color-fg-100);
+    transition: all 200ms ease-in-out 35ms;
```
0xAsimetriq, line 17: Can add `carret-color` property to match current accent color for 🔥 points 😄 
Similarly we could also set global `:selection` background color to be accent and text color to be inverse of that.

0xAsimetriq (changes requested):

packages/ui/src/composites/wui-input/styles.ts
svenvoskamp: Good one! Will set `caret-color` for now. Let's add `:selection` later on.

0xAsimetriq (changes requested):
Just noticed that even if icon is not passed, input will have bigger padding-left as if something was still there, can we adjust this to be normal padding when icon is not provided

0xAsimetriq (approved):

PR #1149, opened 2023-07-04
add wui-input-cancel component

svenvoskamp:
# Changes

- feat: Add `wui-input-cancel` component


packages/ui/src/composites/wui-input-cancel/index.ts
```diff
@@ -0,0 +1,29 @@
+import { html, LitElement } from 'lit'
```
0xAsimetriq, line 1: Let's make this component more reusable i.e. something like general `wui-input-element` where any icon can be provided.

0xAsimetriq (changes requested):

0xAsimetriq (approved):

PR #1152, opened 2023-07-05
[v3] wui-overlay

0xAsimetriq:
# Changes

- Adds overlay component
- Adds new theme variables

### DIFF
--- a/packages/ui/src/layout/wui-overlay/styles.ts
+++ b/packages/ui/src/layout/wui-overlay/styles.ts
@@ -1,0 +1,23 @@
+import { css } from 'lit'
+
+export default css`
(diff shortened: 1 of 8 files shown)


packages/ui/src/layout/wui-overlay/styles.ts
```diff
@@ -0,0 +1,14 @@
+import { css } from 'lit'
+
+export default css`
+  :host {
+    top: 0;
```
svenvoskamp, line 5: I'd suggest to use `inset: 0;` instead of `top`, `right`, `bottom`, `left` separately. Saves a few lines of code.

PR #1153, opened 2023-07-05
V 3/ Add wui shimmer

svenvoskamp:
# Changes

- feat: Add `wui-shimmer` component

### DIFF
--- a/packages/ui/src/components/wui-shimmer/index.ts
+++ b/packages/ui/src/components/wui-shimmer/index.ts
@@ -1,4 +1,4 @@
-import { LitElement, html } from 'lit'
+import { LitElement } from 'lit'
 import { customElement, property } from 'lit/decorators.js'
 import styles from './styles'
 import type { BorderRadiusType } from '../../utils/TypesUtil'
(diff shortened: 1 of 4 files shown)


packages/ui/src/components/wui-shimmer/index.ts
```diff
@@ -0,0 +1,31 @@
...
+  @property() public borderRadius: BorderRadiusType = 'm'
+
+  public render() {
+    this.style.cssText = `
+      width: ${this.width};
+      height: ${this.height};
+      border-radius: ${`var(--wui-border-radius-${this.borderRadius})`};
+    `
+
+    return html``
```
0xAsimetriq, line 23: I guess there is no way to get around this in shimmers scenario? Like, can it return null or something?

0xAsimetriq (dismissed):

packages/ui/src/components/wui-shimmer/index.ts
svenvoskamp: `return null` works, so changed it to that

0xAsimetriq (approved):

PR #1160, opened 2023-07-06
V 3/ add wui-loading-spinner & wui-loading-thumbnail

svenvoskamp:
# Changes

- feat: add `wui-loading-spinner` & `wui-loading-thumbnail`


packages/ui/src/composites/wui-loading-spinner/index.ts
```diff
@@ -0,0 +1,28 @@
...
+import '../../components/wui-icon'
+import { resetStyles } from '../../utils/ThemeUtil'
+import styles from './styles'
+
+@customElement('wui-loading-spinner')
+export class WuiLoadingSpinner extends LitElement {
+  public static styles = [resetStyles, styles]
+
+  // -- State & Properties -------------------------------- //
+
```
0xAsimetriq, line 12:
```suggestion
```

packages/ui/src/composites/wui-loading-spinner/index.ts
```diff
@@ -0,0 +1,28 @@
...
+  @property({ type: Boolean }) public loading = true
+
+  // -- Render -------------------------------------------- //
+
```
0xAsimetriq, line 16:
```suggestion
```

packages/ui/src/composites/wui-loading-spinner/index.ts
```diff
@@ -0,0 +1,28 @@
...
+import { resetStyles } from '../../utils/ThemeUtil'
+import styles from './styles'
+
+@customElement('wui-loading-spinner')
+export class WuiLoadingSpinner extends LitElement {
+  public static styles = [resetStyles, styles]
+
+  // -- State & Properties -------------------------------- //
+
+  @property({ type: Boolean }) public loading = true
```
0xAsimetriq, line 13: To stick with common patterns, we can remove loading prop, this will simply remain as an active loading component at all times. Then in scaffold, we will determine when to render or not render the loaders. It is important to remove the from dom when they are not needed, not just set their opacity to 0. (we can animate from scaffold as well if we want)

packages/ui/src/composites/wui-loading-thumbnail/index.ts
```diff
@@ -0,0 +1,30 @@
...
+import { resetStyles } from '../../utils/ThemeUtil'
+import styles from './styles'
+
+@customElement('wui-loading-thumbnail')
+export class WuiLoadingThumbnail extends LitElement {
+  public static styles = [resetStyles, styles]
+
+  // -- State & Properties -------------------------------- //
+
+  @property({ type: Boolean }) public loading = true
```
0xAsimetriq, line 13: Same comment as in previous component re-loading prop

0xAsimetriq (changes requested):

PR #1161, opened 2023-07-06
V 3 / button refactor

svenvoskamp:
# Changes

- feat: Icon refactor & added changes for every component that uses `wui-icon`


packages/ui/src/components/wui-icon/index.ts
```diff
@@ -1,27 +1,67 @@
...
+        size = '18px'
+        break
+      default:
+        size = 'inherit'
     }
 
-    return html`<slot class="${classMap(classes)}"></slot>`
+    this.style.cssText = `
+      color: ${`var(--wui-color-${this.color})`};
+      width: ${size};
```
0xAsimetriq, line 60: Would it be easier to use something like `width: var(--wui-size=${this.size})` here instead of re-mapping names to values in switch statement? I think you started doing that latter down the line, so maybe it can be applied everywhere where these similar switch statements are used?

packages/ui/src/composites/wui-icon-box/index.ts
```diff
@@ -40,7 +42,7 @@ export class WuiIconBox extends LitElement {
 
     return html`
       <div class="wui-overlay-default ${classMap(classes)}" style=${bgColorStyle}>
```
0xAsimetriq, line 44: I think we might be able to get rid of this div now via using:

```
this.classList = classMap(classes)
this.style.cssText = bgColorStyle
```

and styling :host directly (with display block)

0xAsimetriq (dismissed):
This is awesome, left few suggestions!

packages/ui/src/composites/wui-icon-box/index.ts
```diff
@@ -31,18 +32,21 @@ export class WuiIconBox extends LitElement {
...
-    }
-
-    const bgColorStyle = `--wui-bg-value: var(--wui-color-${this.backgroundColor});`
-
-    return html`
-      <div class="wui-overlay-default ${classMap(classes)}" style=${bgColorStyle}>
-        <wui-icon color=${this.iconColor} size=${iconSize}><slot></slot></wui-icon>
-      </div>
-    `
+    this.style.cssText = `
```
0xAsimetriq, line 35: This looks a bit odd / inconsistent with everything else we have going on, plus I think we lost color-mix fallback of `--wui-overlay-020` that used to be here before. Previous approach with styles being in styles.ts looked more readable / cleaner.

What if instead of doing full style here, you just set css variables and still use styles in their own files? i.e. set up all variable options as local css variables (let's reserve --wui prefix for global) and then use them inside styles.ts

```ts
   const isLg = this.size === lg
   const bgMix = isLg ? '12%' : '16%'
   const borderRadius =  isLg ? '3xl' : 'xxs'
   
    this.style.cssText = `
      --local-bg-value: var(--wui-color-${this.backgroundColor});
      --local-bg-mix: ${bgMix};
      --local-border-radius: var(--wui-border-radius-${borderRadius});
      --local-size: var(--wui-icon-box-size-${this.size});
  `
```

0xAsimetriq (approved):

PR #1165, opened 2023-07-06
V 3 / Add wui tooltip

svenvoskamp:
# Changes

- feat: Add `wui-tooltip`


packages/ui/src/composites/wui-tooltip/index.ts
```diff
@@ -0,0 +1,35 @@
...
+import { elementStyles, resetStyles } from '../../utils/ThemeUtil'
+import type { PlacementType } from '../../utils/TypesUtil'
+import styles from './styles'
+
+@customElement('wui-tooltip')
+export class WuiTooltip extends LitElement {
+  public static styles = [resetStyles, elementStyles, styles]
+
+  // -- State & Properties -------------------------------- //
+
```
0xAsimetriq, line 14:
```suggestion
```

0xAsimetriq (changes requested):
I think placement options are inverted? i.e. when setting plcement `top` I expect tooltip to appear on the top, but right now it looks like it is set up like it is appearing at the bottom, same is true for left / right

0xAsimetriq (approved):

PR #1174, opened 2023-07-07
V 3 / wui chip

svenvoskamp:
# Changes

- feat: Add `wui-chip` component


apps/gallery/stories/composites/wui-chip.stories.ts
```diff
@@ -0,0 +1,44 @@
...
+  walletImageSrc,
+  walletImagesOptions
+} from '../../utils/PresetUtils'
+
+type Component = Meta<WuiChip>
+
+export default {
+  title: 'Composites/wui-chip',
+  args: {
+    src: walletImageSrc,
```
0xAsimetriq, line 18: Do we need both `src` and `imageSrc`?

packages/ui/src/composites/wui-chip/index.ts
```diff
@@ -0,0 +1,51 @@
...
+import { removeUrlPrefix } from '../../utils/HelperUtils'
+
+@customElement('wui-chip')
+export class WuiChip extends LitElement {
+  public static styles = [resetStyles, elementStyles, styles]
+
+  // -- State & Properties -------------------------------- //
+  @property() public variant: ChipType = 'fill'
+
+  @property() public imageSrc? = ''
```
0xAsimetriq, line 18: When property is optional, default to `undefined`

packages/ui/src/utils/HelperUtils.ts
```diff
@@ -9,3 +9,16 @@ export function getSpacingStyles(spacing: SpacingType | SpacingType[], index: nu
 
   return undefined
 }
+
+export function removeUrlPrefix(url: string) {
```
0xAsimetriq, line 13: I believe you are trying to get a domain name from a url here? In which case using `new URL()` and grabbing hostname on that would be more stable / correct.

0xAsimetriq (changes requested):
Added some comments, I also noticed that:

1. Vertical padding on this element (possibly button as well?) is inconsistent i.e. 9px top, 10px bottom

2. When `imgSrc` is not provided, padding on the left is too small (I believe it is initially smaller to make image look better)

0xAsimetriq:
Outline for the image is missing (included in designs)
Screenshot 2023-07-10 at 08 51 05

It looks like non uniform padding is set 8.5 / 10.5px
Screenshot 2023-07-10 at 08 52 34

0xAsimetriq (approved):

PR #1176, opened 2023-07-10
V 3 / Add wui-transaction-visual

svenvoskamp:
# Changes

- feat: Add `wui-transaction-visual` component


packages/ui/src/composites/wui-transaction-visual/index.ts
```diff
@@ -0,0 +1,54 @@
...
+import { customElement, property } from 'lit/decorators.js'
+import { elementStyles, resetStyles } from '../../utils/ThemeUtil'
+import '../wui-icon-box'
+import '../../components/wui-image'
+import styles from './styles'
+import type { TransactionIconType, TransactionType } from '../../utils/TypesUtil'
+
+@customElement('wui-transaction-visual')
+export class WuiTransactionVisual extends LitElement {
+  public static styles = [resetStyles, elementStyles, styles]
```
0xAsimetriq, line 11: Are `elementStyles` needed here, since no form elements are used?

packages/ui/src/composites/wui-transaction-visual/index.ts
```diff
@@ -0,0 +1,54 @@
...
+  // -- State & Properties -------------------------------- //
+  @property() public type: TransactionType = 'buy'
+
+  @property() public imageSrc = ''
+
+  // -- Render -------------------------------------------- //
+  public render() {
+    let color: 'blue-100' | 'error-100' | 'success-100' = 'blue-100'
+    let icon: TransactionIconType = 'arrowTop'
+    if (
```
0xAsimetriq, line 22: You can make small arrays here i.e `const outgoing = ['bought', 'buy', 'deposited']` etc. and then use them in if statements like `if (outgoing.includes(this.type))`, should help simplify thinsg

0xAsimetriq (changes requested):

svenvoskamp:
* Simplified statements in `wui-transaction-visual`
* Removed elementStyles

I will let the `imageSrc` be for what is right now as we will postpone this for a later version.

PR #1182, opened 2023-07-11
V 3/wui wallet image refactor

svenvoskamp:
# Changes

- feat:
- fix: `wui-wallet-image` refactor
- chore:


packages/ui/src/composites/wui-wallet-image/index.ts
```diff
@@ -30,7 +41,12 @@ export class WuiWalletImage extends LitElement {
       return html`<wui-image src=${this.imageSrc} alt=${this.name}></wui-image>`
     }
 
-    return html`<wui-icon size="inherit" color="inherit" name="walletPlaceholder"></wui-icon>`
+    return html`<wui-icon
+      parentSize=${this.size}
```
0xAsimetriq, line 45: Let's use `data-` attributes for properties that are not passed down to components and only used to target element is styling, i.e. `data-soze=...`

I believe there are other components where this is the case, so lets apply `data-` approach there as well.

For a second here I was confused to what this new `parentSize` prop is

0xAsimetriq (changes requested):

0xAsimetriq (approved):

PR #1186, opened 2023-07-11
V 3/wui account entry

svenvoskamp:
# Changes

- feat: Add `wui-account-entry`


0xAsimetriq (commented):
Let's make `wui-flex` part of the component, so that we only need to pass wu-text as children perhaps?

Also not sure about the name, we should think of components as something that can be used outside of modal i.e. outside of account view is a possibility as well.

We have similar `wui-list-select` component, what if we do something like `wui-list-button` / `wui-list-item` or something else?

0xAsimetriq (approved):

PR #1191, opened 2023-07-12
V 3/ Add wui avatar

svenvoskamp:
# Changes

- feat: Add `wui-avatar


packages/ui/src/composites/wui-avatar/index.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  // -- Private ------------------------------------------- //
+  public visualTemplate() {
+    if (this.imageSrc && this.alt) {
+      return html`<wui-image src=${this.imageSrc} alt=${this.alt}></wui-image>`
+    } else if (this.address) {
+      const cssColors = generateAvatarColors(this.address)
+      this.style.cssText = cssColors
+
+      return html`<div class="generated"></div>`
```
0xAsimetriq, line 32: Do we need div? Perhaps we can programatically set data attribute on host and style it accordingly? `this.dataset.variant = ...`

packages/ui/src/utils/HelperUtils.ts
```diff
@@ -15,3 +16,43 @@ export function getHostName(url: string) {
 
   return newUrl.hostname
 }
+
+export function generateAvatarColors(address: string): string {
```
0xAsimetriq, line 20: Do we need to specify return type of `: string` here? As it will be inherited by typescript, in general I tend to let ts do all the heavy lifting and only specify types when those are inherited as `any` only.

0xAsimetriq (commented):
Looks much better than what we have currently! Left few suggestions

0xAsimetriq (approved):

PR #1193, opened 2023-07-12
V 3/wui email input

svenvoskamp:
# Changes

- feat: Add `wui-email-input`


packages/ui/src/composites/wui-email-input/index.ts
```diff
@@ -0,0 +1,38 @@
...
+import { customElement, property } from 'lit/decorators.js'
+import { resetStyles } from '../../utils/ThemeUtil'
+import '../../components/wui-icon'
+import '../../components/wui-text'
+import '../../composites/wui-input'
+import styles from './styles'
+
+@customElement('wui-email-input')
+export class WuiEmailInput extends LitElement {
+  public static styles = [resetStyles, styles]
```
0xAsimetriq, line 11: Should this include element styles as well?

packages/ui/src/composites/wui-email-input/index.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  // -- State & Properties -------------------------------- //
+  @property() public errorMessage?: string
+
+  // -- Render -------------------------------------------- //
+  public render() {
+    return html` <wui-input placeholder="Email" icon="mail" size="md">
```
0xAsimetriq, line 18: Formatting looks off

0xAsimetriq:
This element needs to have some padding / bigger "hit box" to make it easier to click on, especially on mobile. Also, I don't see it in figma, but we probably want some sort of hover effect here? Either lighten background or maybe move arrow slightly to the right?

Screenshot 2023-07-12 at 16 42 13

0xAsimetriq (approved):

PR #1195, opened 2023-07-13
V 3 / Add wui-qr-code

svenvoskamp:
# Changes

- feat: Add `wui-qr-code`


packages/ui/src/composites/wui-qr-code/styles.ts
```diff
@@ -0,0 +1,51 @@
+import { css } from 'lit'
+
+export default css`
+  :host {
+    position: relative;
+    user-select: none;
+    display: block;
+    overflow: hidden;
+    aspect-ratio: 1 / 1;
+    animation: fadeIn ease 0.2s;
```
0xAsimetriq, line 10: In v3, we won't need fade in animation as qrcode will be available upon opening the modal

0xAsimetriq (commented):
1. Not sure if possible, but could these white bits be made transparent?
Screenshot 2023-07-13 at 12 45 17

2. This one applies to all elements that have / use "outlines". In dark mode we have them use light overlay, however in light mode they have these darker outlines. For "W" logo in particular, we should always use dark outline as it will always be placed against light background.

Screenshot 2023-07-13 at 12 45 42
Screenshot 2023-07-13 at 12 46 29

0xAsimetriq (approved):

PR #1198, opened 2023-07-14
chore: started adding dom tags + readme updates to ui elements

JonathanConn:
# Changes

- chore: started adding dom tags + readme updates to ui elements

Was going back and forth a lot building tests, adding tags
Decided to just to overshoot and add most of the remaining tags

Need to finish adding remaining views


packages/ui/readme.md
```diff
@@ -17,3 +17,111 @@ Your on-ramp to web3 multichain. Web3Modal is a versatile library that makes it
 ## Development
 
 Please follow [developer docs](./.github/docs/development.md) to set up web3modal locally.
+
```
0xAsimetriq, line 20: We need to put these docs under `.github/docs` somewhere as package readme's are user facing / displayed on npm.

JonathanConn: Good idea, I will move it

0xAsimetriq (commented):
Another bit (which you might've planned for latter), we probably want one index file with all of these as constants, then we apply constants throughout the code.

Having such file will mean we only need to tweak names in one place and tests can import them from there as well.

JonathanConn:
@0xAsimetriq You mentioned we moved to cloudflare pages, is it safe to ignore the failed vercel deployments?

0xAsimetriq (approved):
Looked over this again and got following question: Do we need testids on "components" i.e. each button within the modal will have these added now, and they accumulate quiet a lot to increase bundle size slightly.

Do we not want these only on concrete actionable elements instead? For example instead of having `data-testid=button-component` show up on all buttons, we probably want `data-testid=connect-btn` / `data-testid=view-all-btn` etc?

If you do need them on all components, we can keep as is 👍

PR #1200, opened 2023-07-14
V 3 / design feedback

svenvoskamp:
Resolved design feedback:

- font issues fixed
- Tweaked paddings for some elements according to design
- `:hover`, `:active` and `:focus-visible` states are now according to design for every component 
- updated mail and search icons 
- `wui-tag` according to design 
- `wui-input` according to design
- `wui-tooltip according to design
- tweaked `wui-transaction-visual`
- `wui-list-transaction` added


packages/ui/src/composites/wui-chip/styles.ts
```diff
@@ -36,7 +36,7 @@ export default css`
   a[data-variant='fill'],
   a[data-variant='shade'] {
     column-gap: var(--wui-spacing-xs);
-    padding: 8.5px 14px 9.5px 8px;
+    padding: 6px 14px 6px 8px;
```
0xAsimetriq, line 39: Should these be using css vars for spacing perhaps?

packages/ui/src/composites/wui-list-transaction/index.ts
```diff
@@ -0,0 +1,49 @@
...
+@customElement('wui-list-transaction')
+export class WuiListTransaction extends LitElement {
+  public static styles = [resetStyles, elementStyles, styles]
+
+  // -- State & Properties -------------------------------- //
+  @property() public type: TransactionType = 'bought'
+
+  @property({ type: Boolean }) public disabled = false
+
+  @property() public imageSrc = ''
```
0xAsimetriq, line 18: I think imgSrc should be optional and we should fallback to either generic transaction graphic or to fallback specific to transaction type.

packages/ui/src/composites/wui-list-transaction/index.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  @property() public date = ''
+
+  @property() public transactionDetail = ''
+
```
0xAsimetriq, line 23: Date should be of `Date` type and we should format it via https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat

0xAsimetriq (changes requested):

svenvoskamp:
* Added css variables where I could. 
* Added Date type

0xAsimetriq (approved):

PR #1226, opened 2023-07-24
Feat/e2e wallet

JonathanConn:
WIP

Refactored test structure
The basic functional tests are complete

Will be doing a separate PR for remaining tasks, UI Validators, Pipeline, Injected wallets, etc..


JonathanConn:
Even with splitting up the PR
There is a lot going on here

The key difference is the refactor using generic playwright structures to playwright fixtures
This created a significant improvement in the quality of code, and ability to write effective tests

laboratory/.github/workflows/playwright.yml
```diff
@@ -0,0 +1,27 @@
+name: Playwright Tests
+on:
+  push:
+    branches: [ main, master ]
+  pull_request:
+    branches: [ main, master ]
+jobs:
+  test:
+    timeout-minutes: 60
```
chris13524, line 9: I feel like we could get away with 30 or even 15 here, could we set this a little lower so we don't waste minutes in the case of a coding error?

laboratory/.github/workflows/playwright.yml
```diff
@@ -0,0 +1,27 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - uses: actions/setup-node@v3
+      with:
+        node-version: 18
+    - name: Install dependencies
+      run: npm ci
+    - name: Install Playwright Browsers
+      run: npx playwright install --with-deps
```
chris13524, line 19: Should we call the script instead? Not sure code style desires here. Doesn't matter to me.

```suggestion
      run: npm run playwright:install
```

(same for the test step)

laboratory/.gitignore
```diff
@@ -0,0 +1,9 @@
+node_modules/
+/test-results/
+/playwright-report/
+/playwright/.cache/
+/test-results/
+/playwright-report/
+/playwright/.cache/
+/tests/shared/data/
+!/tests/shared/data/extension.txt
```
chris13524, line 9: Feels odd to need `!`, is it possible to refactor to do without this? Generally I prefer being able to e.g. delete a build/cache/output folder and not worry about things inside needing to persist. But not sure exactly how this is used.

laboratory/playwright.config.ts
```diff
@@ -0,0 +1,88 @@
...
+     *     {
+     *     name: 'firefox',
+     *     use: { ...devices['Desktop Firefox'] }
+     *     },
+     *
+     *     {
+     *     name: 'webkit',
+     *     use: { ...devices['Desktop Safari'] }
+     *     }
+     */
```
chris13524, line 56: I think we can remove this comment since it's used above?

```suggestion
```

laboratory/tests/shared/pages/WalletPage.ts
```diff
@@ -0,0 +1,78 @@
+/* eslint-disable no-await-in-loop */
+import type { Locator, Page } from '@playwright/test'
+import type { SessionParams } from '../types'
+import { LOCAL_WALLET_URL } from '../constants'
+
+export class WalletPage {
+  private readonly baseURL = LOCAL_WALLET_URL
+
+  private readonly goToAccounts: Locator
```
chris13524, line 9: Minor readability, `goto` is easier to read than `goTo` imo and matches Playwright's `page.goto()` function.

```suggestion
  private readonly gotoAccounts: Locator
```

laboratory/tests/shared/validators/ModalValidator.ts
```diff
@@ -0,0 +1,43 @@
+import { expect } from '@playwright/test'
+import type { Page } from '@playwright/test'
+import { LOCAL_LAB_URL } from '../constants'
+
+export class ModalValidator {
+  private readonly baseURL = LOCAL_LAB_URL
+
+  constructor(public readonly page: Page) {}
+
+  async isConnected() {
```
chris13524, line 10: IMO "expect" represents the behavior of these functions better than "is" because they call "expect". I almost though this function returned a boolean.

```suggestion
  async expectConnected() {
```

chris13524 (dismissed):
Looks pretty solid! Good work!

Just left some thoughts on style.

laboratory/.github/workflows/playwright.yml
```diff
@@ -0,0 +1,27 @@
+name: Playwright Tests
+on:
+  push:
+    branches: [ main, master ]
```
0xAsimetriq, line 4: We don't have main and master, instead it's V2. Assuming same tests run during pull request, do we need them to re-run on push?

laboratory/.github/workflows/playwright.yml
```diff
@@ -0,0 +1,27 @@
...
+      with:
+        node-version: 18
+    - name: Install dependencies
+      run: npm ci
+    - name: Install Playwright Browsers
+      run: npx playwright install --with-deps
+    - name: Run Playwright tests
+      run: npx playwright test
+    - uses: actions/upload-artifact@v3
+      if: always()
```
0xAsimetriq, line 23: Seems redundant? if it's always ran, do we need the statement

laboratory/.gitignore
```diff
@@ -0,0 +1,9 @@
+node_modules/
+/test-results/
+/playwright-report/
+/playwright/.cache/
+/test-results/
+/playwright-report/
```
0xAsimetriq, line 6: Duplicated entries here

0xAsimetriq (commented):
Leaving few comments 👍

laboratory/.github/workflows/playwright.yml
```diff
@@ -0,0 +1,27 @@
+name: Playwright Tests
```
arein, line 1: Can you please add a `concurrency` definition similar to https://github.com/WalletConnect/rpc-proxy/blob/master/.github/workflows/ci.yaml#L18-L24

JonathanConn:
@chris13524 @0xAsimetriq @arein 
Thank you all for the feedback 🙏
The GitHub workflow was auto generated by playwright
I left it in as a reference for the next PR since workflows in subdirs won't run
Be back tomorrow 🤟

bkrem (approved):

laboratory/.github/workflows/playwright.yml
JonathanConn: Sorry I missed this
It is needed
If the tests were to fail, and that if statement was not there, then the reports would not be run

PR #1237, opened 2023-07-31
[v3] Mobile and other improvements

0xAsimetriq:
# Changes

- Keyboard management
- Mobile layout
- Dependency updates
- Balance formatting
- Scroll locking
- connection widget simplification

### DIFF
--- a/packages/scaffold-html/src/modal/w3m-modal/index.ts
+++ b/packages/scaffold-html/src/modal/w3m-modal/index.ts
@@ -5,6 +5,9 @@ import { customElement, state } from 'lit/decorators.js'
 import { animate } from 'motion'
 import styles from './styles'
 
+// -- Helpers --------------------------------------------- //
+const SCROLL_LOCK = 'scroll-lock'
+
 @customElement('w3m-modal')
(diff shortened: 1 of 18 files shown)


packages/scaffold-html/src/modal/w3m-modal/index.ts
```diff
@@ -51,14 +57,64 @@ export class W3mModal extends LitElement {
...
+        touch-action: none;
+        overflow: hidden;
+        overscroll-behavior: contain;
+        scrollbar-gutter: stable;
+      }
+    `
+    document.head.appendChild(styleTag)
+  }
+
+  private onScrollUnlock() {
```
svenvoskamp, line 88: I think you will go back to the top of the page if you scroll > then open modal > then close again

Could we set the body `scrollTop` when closing the modal? 

`document.body.style.top =  ${window.scrollY}px;`

0xAsimetriq: No need, this works as expected (same implementation as v2) you stay at the same position

PR #1238, opened 2023-08-01
Feat/e2e pipeline

JonathanConn:
WIP 

# Changes
- feat: Added e2e pipeline


.github/workflows/pr_e2e_tests.yml
```diff
@@ -0,0 +1,51 @@
+name: Pull request E2E tests
+
```
arein, line 2: why are you ignoring my previous PR feedback here?

Please either say you disagree with the feedback or agree with it but don't ignore it.

.github/workflows/pr_e2e_tests.yml
```diff
@@ -0,0 +1,51 @@
...
+on:
+  pull_request:
+    branches:
+      - V2
```
arein, line 6: feels like a regression to the previous iteration where you also ran upon `push`

arein (changes requested):

.github/workflows/pr_e2e_tests.yml
JonathanConn: I did not ignore your pervious feedback
This draft pr was never meant to be reviewed
I was using this PR to test solutions to an issue I linked in our slack
I agree with the changes you suggested and they will be there in the final version

.github/workflows/pr_e2e_tests.yml
JonathanConn: The previous iteration did not exist, I never wrote a previous iteration
That was auto generated by playwright
I later removed it to try and fix the confusion

PR #1242, opened 2023-08-01
V 3/account button

svenvoskamp:
# Changes

- feat: Add `wui-account-button`

### DIFF
--- a/packages/ui/src/composites/wui-account-button/index.ts
+++ b/packages/ui/src/composites/wui-account-button/index.ts
@@ -11,7 +11,7 @@ import styles from './styles'
 
 @customElement('wui-account-button')
 export class WuiAccountButton extends LitElement {
-  public static styles = [resetStyles, elementStyles, styles]
+  public static override styles = [resetStyles, elementStyles, styles]
 
   // -- State & Properties -------------------------------- //
   @property() public networkSrc?: string = undefined
@@ -23,7 +23,7 @@ export class WuiAccountButton extends LitElement {
   @property() public address = ''
 
   // -- Render -------------------------------------------- //
-  public render() {
+  public override render() {
     return html`
       <button>
         ${this.balanceTemplate()}


packages/ui/src/composites/wui-account-button/index.ts
```diff
@@ -0,0 +1,54 @@
...
+import { elementStyles, resetStyles } from '../../utils/ThemeUtil'
+import { UiHelperUtil } from '../../utils/UiHelperUtils'
+import styles from './styles'
+
+@customElement('wui-account-button')
+export class WuiAccountButton extends LitElement {
+  public static styles = [resetStyles, elementStyles, styles]
+
+  // -- State & Properties -------------------------------- //
+  @property() public imageSrc?: string = undefined
```
0xAsimetriq, line 16: We need `networkSrc` and `avatarSrc` props (in case user has ens profile, otherwise we show wui-avatar)

packages/ui/src/composites/wui-account-button/index.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  @property() public balance?: string
```
0xAsimetriq, line 18:
```suggestion
  @property() public balance?: string = undefined
```

packages/ui/src/composites/wui-account-button/index.ts
```diff
@@ -0,0 +1,54 @@
...
+            ${UiHelperUtil.getTruncateAddress(this.address, 4)}
+          </wui-text>
+        </wui-flex>
+      </button>
+    `
+  }
+
+  // -- Private ------------------------------------------- //
+  private balanceTemplate() {
+    if (this.balance && this.imageSrc) {
```
0xAsimetriq, line 39: Make network image optional and fall back to our network svg when it is not provided, we should only check for required data to be available to render component, which in this case is just balance.

packages/ui/src/utils/UiHelperUtils.ts
```diff
@@ -22,6 +22,10 @@ export const UiHelperUtil = {
     return newUrl.hostname
   },
 
+  getTruncateAddress(address: string, chars: number) {
```
0xAsimetriq, line 25: There is `truncateAddress` helper in core. For this component, but I agree that ui helpers might be more appropriate place. Could you move one from core into ui and apply it in account-view.

0xAsimetriq (changes requested):

0xAsimetriq (approved):

PR #1248, opened 2023-08-01
Feat/e2e ci

JonathanConn:
## Changes
- feat: added CI for e2e tests

### Points of discussion
- Run on every push or PR? Each has its advantages, I am in favor of every push
- How long should we store the test reports? They could build up over time if we decide to run on every `push`

### Ideas for the future
- Instead of building each time, we could utilize cloudflare pages to test 

Important: This [web-example PR](https://github.com/WalletConnect/web-examples/pull/260) needs to be merged before theses tests will pass, as they test against our wallet web-example

### DIFF
--- a/.github/workflows/playwright_tests.yml
+++ b/.github/workflows/playwright_tests.yml
@@ -2,10 +2,13 @@ name: Playwright Tests
 
 on: 
   push:  
-    branches: '**'
+    branches: [ "V2", "V3" ]
+    paths-ignore:
+      - "**.md"
 
   pull_request:
-    branches: [ "V2", "V3" ]    
+    paths-ignore:
+      - "**.md"
 
 concurrency:
   # Support push/pr as event types with different behaviors each:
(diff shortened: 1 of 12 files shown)


.github/workflows/playwright_tests.yml
```diff
@@ -0,0 +1,57 @@
...
+env:
+  NEXT_PUBLIC_PROJECT_ID: ${{ secrets.NEXT_PUBLIC_PROJECT_ID }}
+
+jobs:
+  e2e_tests:
+      runs-on: ubuntu-latest
+      timeout-minutes: 30
+      permissions:
+        contents: read
+        deployments: write
```
arein, line 27: curious: why do you need this?

arein (dismissed):

.github/workflows/playwright_tests.yml
JonathanConn: Great catch, 
that was leftover from researching with cloudflare
It was needed so the GitHub cloudflare action I was using so use GITHUB_TOKEN with the cloudflare api
Will be removed 🤘

JonathanConn:
Fixed some bugs in test that were passing locally, but failing in CI
Should be good to go now @arein

arein (dismissed):
LGTM!

.github/workflows/playwright_tests.yml
```diff
@@ -0,0 +1,57 @@
+name: Playwright Tests
+
+on: 
+  push:  
+    branches: '**'
```
arein, line 5: Screenshot 2023-08-02 at 10 01 04 PM

I think this causes the tests to runs twice upon PR. Can you narrow down to `V2` and `V3`?

JonathanConn: You're right, I actually had them swapped, fixed 👍

PR #1252, opened 2023-08-03
Feat/e2e v3

JonathanConn:
Added basic test suite to v3

** This is temporary until e2e package is created


JonathanConn:
Decided not to implement as it is a temporary solution and is not needed currently
Will revisit with a more permanent solution soon

PR #1254, opened 2023-08-03
fix: limiting scope of browser installs

JonathanConn:


arein (approved):
Awesome!

chris13524:
Why no longer testing on Webkit (Safari)?

JonathanConn:
> Why no longer testing on Webkit (Safari)?
Clipboard permission issue with webkit. All of these tests require a wallet connection. It will be added back in later when I implement out wallet sdk



> Installing all browsers takes 40s on my [original PR](https://github.com/WalletConnect/web3modal/actions/runs/5124348666/job/13868051727), why does it still take >3minutes here?

Not sure honestly, it seems sort of random, I have seen 30 seconds, and I have seen 3 minutes. Most likely because of the source

PR #1266, opened 2023-08-11
V 3/ button loading

svenvoskamp:
# Changes

- feat: add loading state to `wui-button` and `wui-list-item` & make modal scrollable when viewport height is lower than modal scrollheight


packages/ui/src/layout/wui-overlay/index.ts
```diff
@@ -7,7 +7,31 @@ import styles from './styles.js'
 export class WuiOverlay extends LitElement {
   public static override styles = [resetStyles, styles]
 
+  public constructor() {
+    super()
+    window.addEventListener('resize', () => this.handleResize())
```
0xAsimetriq, line 12: Instead of this heavy listener, let's just use media query for height with a pre-set default for max height.  We can take approximate height of "view all" screen which is and will be the tallest and use that.

Additionally some padding at top and bottom will be good, so modal doesn't stick completely to the edges of the screen.

It's ok if at some point modal will need to "jump" to top during flex settings transition in favor of avoiding js listeners.

0xAsimetriq (changes requested):
In addition to comment on resize listener, both button and list item should also become "dissabled" when loading state is set to true.

0xAsimetriq (approved):

PR #1282, opened 2023-08-17
Add theming to V3

svenvoskamp:
# Changes

* Add `ThemeController`
* Reworked `ThemeUtil.ts`
* Add `themeMode` and `themeVariables` to options 
* Add exports to `react.jsx` and `vue.ts`
* Update laboratory with theming options + dark/light switch


packages/core/src/utils/TypeUtils.ts
```diff
@@ -70,3 +70,7 @@ export interface ApiGetWalletsResponse {
   data: ApiWallet[]
   count: number
 }
+
+export type ThemeMode = 'dark' | 'light'
+
+export type ThemeVariables = Record<string, string>
```
0xAsimetriq, line 76: If we know concrete values here, can we have better type completion by defining specific interface like
```
interface {
   '--w3m-font-family': string,
   // ...
}
```

packages/scaffold/src/client.ts
```diff
@@ -45,6 +51,24 @@ export class Web3ModalScaffold {
...
+    return ThemeController.state.themeMode
+  }
+
+  public getThemeVariables() {
+    return ThemeController.state.themeVariables
+  }
+
+  public setThemeMode(themeMode: ThemeMode) {
+    ThemeController.setThemeMode(themeMode)
+    setColorTheme(themeMode)
```
0xAsimetriq, line 64: Should these become part of the Controller methods, so they are always in sync? We could then remove them from everywhere else and just use ThemeController

packages/wagmi/src/client.ts
```diff
@@ -45,6 +47,8 @@ export interface Web3ModalOptions {
   projectId: ProjectId
   chains?: Chain[]
   _sdkVersion?: SdkVersion
+  themeMode: ThemeMode
```
0xAsimetriq, line 50: Theme mode should also be optional and default to user system choice.

packages/wagmi/src/client.ts
```diff
@@ -66,6 +70,10 @@ export class Web3Modal extends Web3ModalScaffold {
       throw new Error('web3modal:constructor - projectId is undefined')
     }
 
+    if (!themeMode) {
```
0xAsimetriq, line 73: As per above comment, should be optional

packages/wagmi/src/react.tsx
```diff
@@ -33,5 +34,17 @@ export function useWeb3Modal() {
     await modal?.close()
   }
 
-  return { open, close }
+  function setThemeMode(themeMode: ThemeMode) {
```
0xAsimetriq, line 37: Let's create separate hook for theming i.e. `useWeb3ModalTheme`, this can export set functions for mode and variable changes.

Instead of
```ts
  const themeMode = modal?.getThemeMode()
  const themeVariables = modal?.getThemeVariables()
```

You should add these on client in addition to `getThemeMode` and `getThemeVariables` -> `subscribeThemeMode` and `subscribeThemeVariables`.

Then, create a state in `useWeb3ModalTheme` hook that sets up these subscribers and updates react / vue state. Don't forget to set unsubscribe as well to clean everything up when hooks unmount.

packages/wagmi/src/vue.ts
```diff
@@ -32,8 +33,24 @@ export function useWeb3Modal() {
     await modal?.close()
   }
 
+  function setThemeMode(themeMode: ThemeMode) {
```
0xAsimetriq, line 36: Same comment on new theme composable as in react hooks

apps/laboratory/package.json
```diff
@@ -14,6 +14,8 @@
     "@sentry/react": "7.64.0",
     "@web3modal/wagmi": "3.0.0-alpha.3",
     "framer-motion": "10.16.0",
-    "next": "13.4.19"
+    "next": "13.4.19",
+    "react-icons": "4.10.1",
+    "zustand": "4.4.1"
```
0xAsimetriq, line 19: Let's stick with valtio for consistency?

apps/laboratory/src/pages/index.tsx
```diff
@@ -18,6 +18,8 @@ import {
 } from 'wagmi/chains'
 import { ConnectButton } from '../components/ConnectButton'
 import { NetworksButton } from '../components/NetworksButton'
+import useThemeStore from '../utils/StoreUtil'
+import { setThemeVariables } from '@web3modal/ui'
```
0xAsimetriq, line 22: ui, core and scafold packages should never be exposed to end users, they should opperate on theme only from `@web3modal/wagmi` packages.

I believe this is here to set up initial theme? If so, this should be done inside `createModal` function or `new Web3Modal()` constructor. Subsequent updates are done with hook or method on web3modal, like you use setThemeMode here.

dangerfile.ts
```diff
@@ -60,7 +60,11 @@ async function checkUiPackage() {
     }
 
     if (diff?.added.includes('@state(')) {
-      fail(`${f} is using @state decorator, which is not allowd in ui package`)
+      fail(`${f} is using @state decorator, which is not allowed in ui package`)
+    }
+
+    if (diff?.added.includes('import @web3modal/core')) {
+      fail(`${f} is using importing @web3modal/core, which is not allowed in ui package`)
```
0xAsimetriq, line 67:
```suggestion
      fail(`${f} is importing @web3modal/core, which is not allowed in ui package`)
```

dangerfile.ts
```diff
@@ -141,6 +145,10 @@ async function checkCorePackage() {
   for (const f of created_core_controllers) {
     const diff = await diffForFile(f)
 
+    if (diff?.added.includes('import @web3modal/ui')) {
+      fail(`${f} is using importing @web3modal/ui, which is not allowed in core package`)
```
0xAsimetriq, line 149:
```suggestion
      fail(`${f} is importing @web3modal/ui, which is not allowed in core package`)
```

packages/core/index.ts
```diff
@@ -45,4 +48,19 @@ export { AssetUtil } from './src/utils/AssetUtil.js'
 export { ConstantsUtil } from './src/utils/ConstantsUtil.js'
 export { CoreHelperUtil } from './src/utils/CoreHelperUtil.js'
 export { StorageUtil } from './src/utils/StorageUtil.js'
+
+export type {
```
0xAsimetriq, line 52: Looks like there was a merge conflict here, if you look below, we are exporting `*` all types already, so these are not needed.

packages/core/src/controllers/ThemeController.ts
```diff
@@ -0,0 +1,34 @@
...
+  subscribeKey<K extends StateKey>(key: K, callback: (value: ThemeControllerState[K]) => void) {
+    return subKey(state, key, callback)
+  },
+
+  setThemeMode(themeMode: ThemeMode) {
+    state.themeMode = themeMode
+  },
+
+  setThemeVariables(themeVariables: ThemeVariables) {
+    state.themeVariables = themeVariables
```
0xAsimetriq, line 32: We need to allow partial overrides, i.e.

```suggestion
    state.themeVariables = {...state.themeVariables, themeVariables}
```

packages/core/tests/controllers/ThemeController.test.ts
```diff
@@ -0,0 +1,28 @@
+import { describe, expect, it } from 'vitest'
+import { ThemeController } from '../../index.js'
+
+// -- Tests --------------------------------------------------------------------
+describe('SnackController', () => {
```
0xAsimetriq, line 5:
```suggestion
describe('ThemeController', () => {
```

packages/scaffold/src/client.ts
```diff
@@ -43,11 +49,51 @@ export class Web3ModalScaffold {
...
+    ThemeController.setThemeMode(themeMode)
+  }
+
+  public setThemeVariables(themeVariables: ThemeVariables) {
+    ThemeController.setThemeVariables(themeVariables)
+  }
+
+  public subscribeThemeMode(setter: (value: ThemeMode) => void) {
+    this.unsubscribe.push(
+      ThemeController.subscribeKey('themeMode', val => {
```
0xAsimetriq, line 77: The unsubscribe array pattern is needed only for our web-components. Here, let's remove it + that unbount function. Instead just `return ThemeController.subscribeKey...` this returns unsubscribe method that we can use latter on i.e.

```
const unsubscribe = modal.subscribeTheme()
unsubscribe()
```

You can use this unsubscribe method in a hook's cleanup for example.

Also, let's combine theme subscriber into one function i.e. `subscribeTheme` that subscribes to whole ThemeController state.

packages/scaffold/src/modal/w3m-modal/index.ts
```diff
@@ -59,6 +58,21 @@ export class W3mModal extends LitElement {
     }
   }
 
+  private initializeTheming() {
+    initializeTheming(ThemeController.state.themeVariables)
+    if (ThemeController.state.themeMode) {
+      ThemeController.setThemeMode(ThemeController.state.themeMode)
+    } else if (typeof window !== 'undefined' && window.matchMedia) {
```
0xAsimetriq, line 65: Can we abstract all of this into ui helper that returns just light or dark value to be set.

packages/ui/src/components/wui-icon/index.ts
```diff
@@ -116,7 +116,7 @@ export class WuiIcon extends LitElement {
   // -- Render -------------------------------------------- //
   public override render() {
     this.style.cssText = `
-      color: ${`var(--wui-color-${this.color});`}
+      --local-color: ${`var(--wui-color-${this.color});`}
       width: ${`var(--wui-icon-size-${this.size});`}
```
0xAsimetriq, line 120: Should it also be `--local-width`?

packages/ui/src/components/wui-text/index.ts
```diff
@@ -23,7 +23,7 @@ export class WuiText extends LitElement {
       [`wui-color-${this.color}`]: true
     }
 
-    this.style.cssText = `--local-align: ${this.align}`
+    this.style.cssText = `--local-align: ${this.align}; --local-color: var(--wui-color-${this.color});`
```
0xAsimetriq, line 26: Format each on new line

packages/ui/src/utils/ThemeUtil.ts
```diff
@@ -1,231 +1,433 @@
-import { css } from 'lit'
+import { css, unsafeCSS } from 'lit'
+import type { ThemeVariables } from './TypesUtil.js'
 
 // -- Utilities ---------------------------------------------------------------
-export function initializeTheming() {
+export function initializeTheming(themeVariables: ThemeVariables) {
   const styleTag = document.createElement('style')
+  styleTag.classList.add('w3m')
```
0xAsimetriq, line 7: why do we need to set class here in addition to dataset / querying by dataset latter on as well? Also, if we do need this class it should be `wui-something` not w3m.

packages/ui/src/utils/ThemeUtil.ts
```diff
@@ -1,231 +1,433 @@
...
-    --wui-z-index: 89;
-
-    --wui-box-size-md: 100px;
-    --wui-box-size-lg: 120px;
-
-    --wui-path-network: path(
-      'M43.4605 10.7248L28.0485 1.61089C25.5438 0.129705 22.4562 0.129705 19.9515 1.61088L4.53951 10.7248C2.03626 12.2051 0.5 14.9365 0.5 17.886V36.1139C0.5 39.0635 2.03626 41.7949 4.53951 43.2752L19.9515 52.3891C22.4562 53.8703 25.5438 53.8703 28.0485 52.3891L43.4605 43.2752C45.9637 41.7949 47.5 39.0635 47.5 36.114V17.8861C47.5 14.9365 45.9637 12.2051 43.4605 10.7248Z'
-    );
+export function setThemeVariables(themeVariables: ThemeVariables) {
+  if (typeof document !== 'undefined') {
```
0xAsimetriq, line 18: Not sure if this is the best way to do this i.e. if we have modal open for some sort of "real time" editing, we probably just want to override data, not destroy and re-create it. Perhaps simple

```
styleTag.textContent = createRootStyles(themeVariables).cssText
```

here will do the trick on already existing styleTag?

0xAsimetriq (changes requested):

packages/wagmi/exports/react.tsx
```diff
@@ -20,6 +22,50 @@ export function createWeb3Modal(options: Omit<Web3ModalOptions, '_sdkVersion'>)
   return modal
 }
 
+export function useWeb3ModalTheme() {
+  if (!modal) {
+    throw new Error('Please call "createWeb3Modal" before using "useWeb3ModalTheme" hook')
+  }
+
+  function setThemeMode(themeMode: ThemeMode) {
```
0xAsimetriq, line 30: See how `OpenOptions` are defined to inherit from function params directly, same approach should be applied to theme methods

packages/wagmi/exports/react.tsx
```diff
@@ -20,6 +22,50 @@ export function createWeb3Modal(options: Omit<Web3ModalOptions, '_sdkVersion'>)
...
+  }
+
+  const [themeMode, setInternalThemeMode] = useState(getThemeMode())
+  const [themeVariables, setInternalThemeVariables] = useState(getThemeVariables())
+
+  useEffect(() => {
+    const unsubscribe = modal?.subscribeTheme(setInternalThemeMode, setInternalThemeVariables)
+
+    return () => {
+      if (unsubscribe) {
```
0xAsimetriq, line 53: Can be `unsubscribe?.()` I think

packages/wagmi/exports/react.tsx
```diff
@@ -20,6 +22,50 @@ export function createWeb3Modal(options: Omit<Web3ModalOptions, '_sdkVersion'>)
...
+        unsubscribe()
+      }
+    }
+  }, [])
+
+  return {
+    themeMode,
+    themeVariables,
+    getThemeMode,
```
0xAsimetriq, line 62: get methods here are redundant as same values are in exported state as well

packages/wagmi/exports/vue.ts
```diff
@@ -19,6 +20,55 @@ export function createWeb3Modal(options: Omit<Web3ModalOptions, '_sdkVersion'>)
...
+  const unsubscribe = modal?.subscribeTheme(
+    mode => {
+      themeMode.value = mode
+    },
+    mode => {
+      themeVariables.value = mode
+    }
+  )
+
+  onBeforeMount(() => {
```
0xAsimetriq, line 56: Should this be unmount instead? i.e. usnubscribe when unmounted?

packages/wagmi/exports/react.tsx
```diff
@@ -20,6 +22,50 @@ export function createWeb3Modal(options: Omit<Web3ModalOptions, '_sdkVersion'>)
...
+
+  function getThemeVariables() {
+    return modal?.getThemeVariables()
+  }
+
+  const [themeMode, setInternalThemeMode] = useState(getThemeMode())
+  const [themeVariables, setInternalThemeVariables] = useState(getThemeVariables())
+
+  useEffect(() => {
+    const unsubscribe = modal?.subscribeTheme(setInternalThemeMode, setInternalThemeVariables)
```
0xAsimetriq, line 50: Should be simplified to something like 

```
    const unsubscribe = modal?.subscribeTheme(newValue => {
     setInternalThemeMode(...)
     setInternalThemeVariables(...)
    })
```

0xAsimetriq (changes requested):

examples/vue-wagmi/src/App.vue
```diff
@@ -1,6 +1,12 @@
-<script setup>
+<script lang="ts" setup>
 import { mainnet, arbitrum } from '@wagmi/core/chains'
-import { createWeb3Modal, useWeb3Modal, defaultWagmiConfig } from '@web3modal/wagmi/vue'
+import {
+  createWeb3Modal,
+  useWeb3Modal,
+  defaultWagmiConfig,
+  useWeb3ModalTheme
```
0xAsimetriq, line 7: Unused import for `useWeb3ModalTheme`?

packages/scaffold/src/modal/w3m-modal/index.ts
```diff
@@ -60,8 +59,8 @@ export class W3mModal extends LitElement {
   }
 
   private initializeTheming() {
-    initializeTheming(ThemeController.state.themeVariables)
-    ThemeController.setThemeMode(getColorTheme(ThemeController.state.themeMode))
+    initializeTheming(ThemeController.state.themeVariables, ThemeController.state.themeMode)
```
0xAsimetriq, line 62: Could reduce repetition here via
```ts
const { themeVariables, themeMode } = ThemeController.state
```

packages/scaffold/src/modal/w3m-modal/index.ts
```diff
@@ -60,8 +59,8 @@ export class W3mModal extends LitElement {
...
+    ThemeController.setThemeMode(UiHelperUtil.getColorTheme(ThemeController.state.themeMode))
```
0xAsimetriq, line 63: For clarity, can separate into
```ts
const defaultThemeMode = UiHelperUtil.getColorTheme(ThemeController.state.themeMode)
ThemeController.setThemeMode(defaultThemeMode)
```

packages/ui/src/utils/UiHelperUtils.ts
```diff
@@ -72,5 +72,19 @@ export const UiHelperUtil = {
     }
 
     return regex.number.test(character)
+  },
+
+  getColorTheme(theme: ThemeType | undefined): ThemeType {
```
0xAsimetriq, line 77:
```ts
getColorTheme(theme?: ThemeType)
```

Also, we don't do manual return types i.e. no `: ThemeType` as everything should be inherited by ts and fail down the line. If return type here is resolving to `string`, you can use `as const`

packages/wagmi/exports/vue.ts
```diff
@@ -25,47 +27,33 @@ export function useWeb3ModalTheme() {
...
-  )
+  const unsubscribe = modal?.subscribeTheme(state => {
+    themeMode.value = state.themeMode
+    themeVariables.value = state.themeVariables
+  })
 
-  onBeforeMount(() => {
+  onUnmounted(() => {
     if (unsubscribe) {
       unsubscribe()
```
0xAsimetriq, line 48: Can be `unsubscribe?.()` without if

PR #1308, opened 2023-09-04
V 3/beta tweaks

svenvoskamp:
# Changes

* Added new all wallets visual
* Various UI tweaks (mostly disabled state)


0xAsimetriq:
1. Not sure why, but on this button icon color is same blue as button so it is not visible, to recreate: Connect with browser extension, switch network, decline switch

Screenshot 2023-09-05 at 09 39 18

2. Disabled button's icon and text color don't match

IMG_2635

0xAsimetriq (changes requested):
Left a comment with screenshots on few minor tweaks

0xAsimetriq (approved):

PR #1311, opened 2023-09-06
V 3/tabs

svenvoskamp:
# Changes

- feat: Implemented collapsed tab solution when tabs length > 3


packages/ui/src/composites/wui-tabs/index.ts
```diff
@@ -3,6 +3,7 @@ import { customElement, property, state } from 'lit/decorators.js'
 import { elementStyles, resetStyles } from '../../utils/ThemeUtil.js'
 import type { IconType } from '../../utils/TypesUtil.js'
 import styles from './styles.js'
+import { animate } from 'motion'
```
0xAsimetriq, line 6: `motion` is not a dependency of ui package (it is originally found in scaffold). Let's move it from there to ui to be more accurate with deps.

packages/ui/src/composites/wui-tabs/index.ts
```diff
@@ -13,36 +14,90 @@ export class WuiTabs extends LitElement {
...
   // -- Render -------------------------------------------- //
   public override render() {
-    const isDense = this.tabs.length > 3
+    this.isDense = this.tabs.length > 3
 
     this.style.cssText = `
       --local-tab: ${this.activeTab};
-      --local-tab-width: ${isDense ? '80px' : '100px'}
+      --local-tab-width: ${this.localTabWidth};
+      --local-dense-tab-width: max-content;
```
0xAsimetriq, line 32: `--local-dense-tab-width: max-content;` does this need to be here, seems like it is static so we can just include in css?

packages/ui/src/composites/wui-tabs/index.ts
```diff
@@ -13,36 +14,90 @@ export class WuiTabs extends LitElement {
...
 
+    this.dataset['type'] = this.isDense ? 'flex' : 'block'
+
     return this.tabs.map((tab, index) => {
       const isActive = index === this.activeTab
 
       return html`
-        <button @click=${() => this.onTabClick(index)} data-active=${isActive}>
+        <button
+          style="{{width:
```
0xAsimetriq, line 42: Looks odd, can we keep this in css?

packages/ui/src/composites/wui-tabs/index.ts
```diff
@@ -13,36 +14,90 @@ export class WuiTabs extends LitElement {
...
+          20px}}"
+          @click=${() => this.onTabClick(index)}
+          data-active=${isActive}
+        >
           <wui-icon size="sm" color="inherit" name=${tab.icon}></wui-icon>
-          <wui-text variant=${isDense ? 'tiny-600' : 'small-600'} color="inherit">
-            ${tab.label}
-          </wui-text>
+          <wui-text variant="small-600" color="inherit"> ${tab.label}</wui-text>
```
0xAsimetriq, line 48:
```suggestion
          <wui-text variant="small-600" color="inherit">${tab.label}</wui-text>
```

packages/ui/src/composites/wui-tabs/index.ts
```diff
@@ -13,36 +14,90 @@ export class WuiTabs extends LitElement {
...
         </button>
       `
     })
   }
 
+  override firstUpdated() {
```
0xAsimetriq, line 54: Could you explain what's going on here / why is this needed?

0xAsimetriq (changes requested):

packages/ui/src/composites/wui-tabs/index.ts
svenvoskamp: After a component has been mounted, we save button elements in an array. Then, we call a function to run the initial animation so that the first tab is active and visible. 

This is necessary because by default, all the text components from the buttons are absolutely positioned. We need to calculate the width that the active button should have.

packages/ui/src/composites/wui-tabs/index.ts
```diff
@@ -50,7 +50,7 @@ export class WuiTabs extends LitElement {
       this.buttons = [...this.shadowRoot.querySelectorAll('button')]
       setTimeout(() => {
         this.animateTabs(0, true)
-      }, 1)
+      }, 0.1)
```
0xAsimetriq, line 53: You can change this to 0, setTimeout triggers next cpu tick, so it should work fine for what you are trying to acheive

packages/ui/package.json
```diff
@@ -16,6 +16,7 @@
   },
   "dependencies": {
     "lit": "2.8.0",
+    "motion": "10.16.2",
```
0xAsimetriq, line 19: Moved to new place, but package lock was not uodated, run `npm run clean; npm install` to apply this change

0xAsimetriq (changes requested):

0xAsimetriq (approved):

PR #1315, opened 2023-09-06
V 3/tabs feedback

svenvoskamp:
# Changes

- chore: refine animation


0xAsimetriq (changes requested):
Needs to be updated with v3 and all vercel builds should pass

0xAsimetriq (approved):

PR #1316, opened 2023-09-07
V 3/beta tweaks v3

svenvoskamp:
# Changes

- fix: loader adjusts to master border-radius & scroll lock issue fix


0xAsimetriq (commented):
Looks good code wise, but can we expose border radius option in laboratory to play around with it and test different cases (check how they work)

PR #1320, opened 2023-09-08
V 3 / beta polish

svenvoskamp:
# Changes

- disabled ui states when mobile is in connecting view
- correct amount of loaders 
- correct truncation in account for ENS


packages/scaffold/src/partials/w3m-connecting-header/index.ts
```diff
@@ -1,24 +1,45 @@
-import type { Platform } from '@web3modal/core'
+import { ConnectionController, type Platform } from '@web3modal/core'
```
0xAsimetriq, line 1: Let's keep type imports separate i.e. `import type`

packages/scaffold/src/partials/w3m-connecting-header/index.ts
```diff
@@ -1,24 +1,45 @@
...
   @property({ type: Array }) public platforms: Platform[] = []
 
   @property() public onSelectPlatfrom?: (platform: Platform) => void = undefined
 
+  @state() private buffering = false
+
+  public constructor() {
+    super()
+    this.unsubscribe.push(
+      ConnectionController.subscribeKey('buffering', val => {
```
0xAsimetriq, line 22: Let's simplify to `ConnectionController.subscribeKey('buffering', val => (this.buffering = val))` and remove extra function

packages/scaffold/src/partials/w3m-connecting-wc-mobile/index.ts
```diff
@@ -45,7 +45,11 @@ export class W3mConnectingWcMobile extends W3mConnectingWidget {
   private onBuffering() {
     if (document.visibilityState === 'visible' && !this.error) {
       this.buffering = true
```
0xAsimetriq, line 47: Since buffering was moved out into controler state, let's refactor here to where this.buffering subscribes to state changes instead of manually setting it.

packages/scaffold/src/partials/w3m-header/index.ts
```diff
@@ -44,6 +46,9 @@ export class W3mHeader extends LitElement {
       RouterController.subscribeKey('view', val => {
         this.onViewChange(val)
         this.onHistoryChange()
+      }),
+      ConnectionController.subscribeKey('buffering', val => {
```
0xAsimetriq, line 50: Same comment as above, let's simplify to where extra function is not needed

packages/scaffold/src/partials/w3m-header/index.ts
```diff
@@ -76,6 +85,7 @@ export class W3mHeader extends LitElement {
       return html`<wui-icon-link
         id="dynamic"
         icon="chevronLeft"
+        ?disabled=${ConnectionController.state.buffering}
```
0xAsimetriq, line 88: Use `this.buffering` instead of state here for consistency

packages/ui/src/composites/wui-account-button/index.ts
```diff
@@ -26,17 +26,19 @@ export class WuiAccountButton extends LitElement {
 
   // -- Render -------------------------------------------- //
   public override render() {
+    const isEns = this.address.endsWith('.eth')
```
0xAsimetriq, line 29: This check and `isEns` variable are specific to ethereum, where as within ui / core and scafold we should remain chain agnostic. Instead, check for existance / length of `AccountController.state.profileName`

packages/ui/src/composites/wui-account-button/index.ts
```diff
@@ -26,17 +26,19 @@ export class WuiAccountButton extends LitElement {
...
+
     return html`
       <button ?disabled=${this.disabled}>
         ${this.balanceTemplate()}
-        <wui-flex gap="xxs" alignItems="center" class=${this.balance ? undefined : 'noBalance'}>
+        <wui-flex gap="xxs" alignItems="center" class=${!this.balance && 'noBalance'}>
```
0xAsimetriq, line 34: I think `this.balance ? undefined : 'noBalance'` was more readable to be honest, also let's do `local-no-balance` for class name here, to stick to newly established patterns we have.

packages/ui/src/utils/UiHelperUtils.ts
```diff
@@ -22,8 +22,12 @@ export const UiHelperUtil = {
     return newUrl.hostname
   },
 
-  getTruncateAddress(address: string, chars: number) {
-    return `${address.substring(0, chars)}...${address.substring(address.length - chars)}`
+  getTruncateString(string: string, chars: number, truncateStart = false) {
```
0xAsimetriq, line 25: Let's make this function more versatile by changing `truncateStart` from boolean to `location: 'start' | 'middle' | 'end'`. This will also make it more clear when this util is used.

0xAsimetriq (changes requested):

packages/ui/src/utils/UiHelperUtils.ts
```diff
@@ -22,8 +22,16 @@ export const UiHelperUtil = {
...
+  getTruncateString(string: string, chars: number, truncate: TruncateType) {
+    if (truncate === 'end') {
+      return `${string.substring(0, chars)}...`
+    } else if (truncate === 'start') {
+      return `...${string.substring(string.length - chars)}`
+    } else if (truncate === 'middle') {
+      return `${string.substring(0, chars)}...${string.substring(string.length - chars)}`
+    }
+
+    return null
```
0xAsimetriq, line 34: Instead of null let's fall back for middle truncation as default

0xAsimetriq (commented):
There is no need / substantial higher character limit should be used for ens and normal address truncation in account view.

1. Address should be `[4]...[4]` as before
2. Profile name should be as long as possible before it breaks things

PR #1324, opened 2023-09-10
fix: modal background adds blur effect

hhun:
# Breaking Changes

Chrome version lower than v 111 does not support color mix syntax
Chrome 版本低于v111不支持 color-mix 语法

# Changes

- feat:
- fix:
- chore:

# Associated Issues

closes #...

old
old

new
new


0xAsimetriq:
Ty for this, I saw your bug report, we will fix in a different way by falling back to solid white / dark colors when color mix is not available.

PR #1325, opened 2023-09-11
V 3/ fallback color mix

svenvoskamp:
# Changes

Current fallback was not working, because browser still thinks color-mixed color is defined. Changed to `@supports` fallback. Context: (https://www.smashingmagazine.com/2022/06/simplify-color-palette-css-color-mix/).

Tested on older browsers and now working as expected.


0xAsimetriq (approved):
🙏 Siiick

PR #1326, opened 2023-09-11
V 3/cross browser issues

svenvoskamp:
# Changes

* Fix color variables safari (moved `:root` inside `@supports`)
* Dropped motion package as it has some issues on safari. Implemented Web Animations API


packages/ui/src/composites/wui-tabs/index.ts
```diff
@@ -45,7 +44,9 @@ export class WuiTabs extends LitElement {
           data-active=${isActive}
         >
           <wui-icon size="sm" color="inherit" name=${tab.icon}></wui-icon>
-          <wui-text variant="small-600" color="inherit">${tab.label}</wui-text>
+          <wui-text style="{{opacity:" 0;}} variant="small-600" color="inherit"
+            >${tab.label}</wui-text
```
0xAsimetriq, line 48: Needs better formatting

apps/laboratory/src/pages/_document.tsx
```diff
@@ -5,7 +5,7 @@ export default function Document() {
   return (
     <Html lang="en">
       <Head />
-      <body>
+      <body style={{ minHeight: '100vh' }}>
```
0xAsimetriq, line 8: We need to apply this `min-heiught: 100vh` in `w3m-modal/index` file when calling `onScrollLock` function so it works correctly on all sites.

0xAsimetriq (approved):

PR #1327, opened 2023-09-11
V 3/alpha fixes

0xAsimetriq:
# Changes

- Remove box shadow from qrcode
- Update deps
- Remove some uneeded css
- Adjust modal height calculations
- Make react / vue optional peer dependencies
- Fix some images not loading in
- Increase page entries from 24 to 40 (so we have less loading states when scrolling all wallets view / less api calls)
- Switch to using `allSettled` instead of `all` when resolving promises that can break (i.e. images)
- Switch from using `ref` to `reactive` in vue compostables so objects can be de-structured 
- Other minor cleanup tasks


packages/ui/src/composites/wui-icon-link/styles.ts
```diff
@@ -7,12 +7,6 @@ export default css`
     padding: var(--wui-spacing-2xs);
   }
 
-  @media (max-width: 700px) {
```
svenvoskamp, line 10: I think we should add this again. I tried it on mobile and had a few moments where I just missed the hitbox.

svenvoskamp (approved):

PR #1363, opened 2023-09-28
V3 / ethers preview

svenvoskamp:
# Changes

* Added `ethers-5` package 
* Added support for `WalletConnectProvider`
* Added support for `InjectedProvider`
* Added support for `CoinbaseProvider`
* Added support for `EIP6963Provider`
* Created `utils` package to store shared `constants`, `presets` and `helpers`
* Created `scaffold-react` package to store shared react hooks 
* Created `scaffold-vue` package to store shared vue compositions


Countrychick88:
Hi

0xAsimetriq (commented):
Really really good start! I skipped laboratory related checks and currently glanced at ethers package that was introduced, here is a set of initial feedback.

1. All references to this package / folder name should be `ethers-5`
2. `@walletconnect/ethereum-provider` present in root package json, but should it not be part of ethers package (similarly to how it is a dependency of wamgi)?
3. Sdk version for this package should be `ethers-5`
4. A lot of duplicated code now, let's create `@web3modal/scaffold-react` and `@web3modal/scaffold-vue` to contain all hook logic, perhaps also `@web3modal/utils` for shared constants / presets / caip conversion utils etc. and other similar code. All of these should be set up in a chain agnostic way (i.e. separate into evm folders / names if needed, in case data is specific just to evm)
5. package json for ethers doesn't constrain it to just ethers 5, should this be the case or does ethers 6 also work? In this case ignore my comment about `ethers-5` and we can keep it as simply `ethers`
6. All custom localStorage entries, should be prefixed with `@w3m/`
7. There are references to wagmi in ethers constants
8. Default ethers config should not contain any required chains, and optional chains should be passed to it "from above" by dapp developer
9. There are refferences to wagmi in ethers tsconfig

0xAsimetriq:
And this is feedback from testing the app

1. I am unable to switch chain if it is not configured in my metamask wallet (injected) i.e. no proposal to add new chain
2. ens details dissapear if I switch away from ethereum
3. When I switch chain from within metamask wallet, change is not reflected in modal
4. If I switch account from metamask wallet, chain is not reflected in the modal

dangerfile.ts
```diff
@@ -1,7 +1,7 @@
 /* eslint-disable no-await-in-loop */
 import { danger, fail, message, warn } from 'danger'
 import corePackageJson from './packages/core/package.json' assert { type: 'json' }
-import { VERSION as WAGMI_PACKAGE_VERSION } from './packages/wagmi/src/utils/constants'
+import { VERSION as WAGMI_PACKAGE_VERSION } from './packages/utils/src/constants'
```
0xAsimetriq, line 4: Let's rename this to something universal i.e PACKAGE_VERSION?

packages/ethers-5/exports/react.tsx
```diff
@@ -0,0 +1,94 @@
...
+import { useEffect, useState } from 'react'
+import type { Web3ModalOptions } from '../src/client.js'
+import { Web3Modal } from '../src/client.js'
+import { VERSION } from '@web3modal/utils'
+import { ProviderController } from '../src/store/index.js'
+
+// -- Types -------------------------------------------------------------------
+export type { Web3ModalOptions } from '../src/client.js'
+
+declare global {
```
0xAsimetriq, line 18: I think we can move this global declaration into react's scaffold as well

packages/ethers-5/exports/react.tsx
```diff
@@ -0,0 +1,94 @@
...
+  }, [])
+
+  return {
+    provider,
+    providerType
+  }
+}
+
+export function useWeb3ModalAccount() {
+  const [address, setAddress] = useState(ProviderController.state.address)
```
0xAsimetriq, line 63: If ProviderController is based on valtio, why not use `useSnapshot()` here? (and anywhere else where we need to subscribe to value). Similarly to set values, we can use actions on controller.

packages/ethers-5/exports/vue.ts
```diff
@@ -0,0 +1,41 @@
...
+  W3mNetworkButton
+} from '@web3modal/scaffold'
+import type { Web3ModalOptions } from '../src/client.js'
+import { Web3Modal } from '../src/client.js'
+import { VERSION } from '@web3modal/utils'
+
+// -- Types -------------------------------------------------------------------
+export type { Web3ModalOptions } from '../src/client.js'
+
+export interface ComponentCustomProperties {
```
0xAsimetriq, line 14: I think this had to extend `declare module '@vue/runtime-core'` in order for auto complete to be available for `<w3m-` components inside vue files. Does removing it still work? I think it won't as this is also no longer a global?

0xAsimetriq, line 14: I think this had to extend `declare module '@vue/runtime-core'` in order for auto complete to be available for `<w3m-` components inside vue files. Does removing it still work? I think it won't as this is also no longer a global?

Similarly, this global declaration can be moved to vue scaffold

packages/ethers-5/exports/vue.ts
```diff
@@ -0,0 +1,41 @@
...
+  W3mButton: Pick<W3mButton, 'size' | 'label' | 'loadingLabel' | 'disabled' | 'balance'>
+  W3mNetworkButton: Pick<W3mNetworkButton, 'disabled'>
+}
+
+// -- Setup -------------------------------------------------------------------
+let modal: Web3Modal | undefined = undefined
+
+export function createWeb3Modal(options: Web3ModalOptions) {
+  if (!modal) {
+    modal = new Web3Modal({ ...options, _sdkVersion: `vue-wagmi-${VERSION}` })
```
0xAsimetriq, line 26: wagmi tag used

packages/ethers-5/package.json
```diff
@@ -0,0 +1,95 @@
...
+    "@coinbase/wallet-sdk": "3.7.2",
+    "@walletconnect/ethereum-provider": "2.10.1",
+    "@web3modal/polyfills": "3.1.0",
+    "@web3modal/scaffold": "3.1.0",
+    "@web3modal/scaffold-react": "3.1.0",
+    "@web3modal/scaffold-vue": "3.1.0",
+    "@web3modal/utils": "3.1.0"
+  },
+  "peerDependencies": {
+    "ethers": "5.7.2",
```
0xAsimetriq, line 55: This restricts whole package to just one version of ethers, make it `>=5` and less than or equal to 6 (don't remember the syntax from the top of my head though).

packages/ethers-5/src/client.ts
```diff
@@ -0,0 +1,788 @@
...
+import type EthereumProvider from '@walletconnect/ethereum-provider'
+import {
+  caipNetworkIdToNumber,
+  getEthersCaipDefaultChain,
+  getCaipTokens,
+  hexStringToNumber
+} from '@web3modal/utils'
+import type { Address, ProviderType } from './utils/types.js'
+import { ethers, utils } from 'ethers'
+import { ProviderController } from './store/index.js'
```
0xAsimetriq, line 43: Let's keep same patterns going and swap `store/index.js` to `controllers/ProviderController.js`

packages/ethers-5/src/client.ts
```diff
@@ -0,0 +1,788 @@
...
+
+// -- Types ---------------------------------------------------------------------
+export interface Web3ModalClientOptions extends Omit<LibraryOptions, 'defaultChain' | 'tokens'> {
```
0xAsimetriq, line 46: I think a lot of these types are now repeated in wagmi and ethers? We could move them to utilities as well?

Make sure to keep utilities organised i.e. split types/presets/constants etc into their own files

packages/ethers-5/src/client.ts
```diff
@@ -0,0 +1,788 @@
...
+        })
+      },
+
+      connectExternal: async ({ id, info, provider }) => {
+        if (id === INJECTED_CONNECTOR_ID) {
+          const injectedProvider = ethersConfig.injected
+          if (!injectedProvider) {
+            throw new Error('connectionControllerClient:connectInjected - connector is undefined')
+          }
+          await injectedProvider.send('eth_requestAccounts', []).then(() => {
```
0xAsimetriq, line 181: why is `await` + `.then()` needed here?

packages/ethers-5/src/client.ts
```diff
@@ -0,0 +1,788 @@
...
+            this.setInjectedProvider(ethersConfig)
+          })
+        } else if (id === EIP6963_CONNECTOR_ID && info && provider) {
+          const EIP6963Provider = provider as ethers.providers.Web3Provider
+          const EIP6963Info = info as Info
+
+          await EIP6963Provider.send('eth_requestAccounts', []).then(() => {
```
0xAsimetriq, line 188: Inconsistent casings in provider name, I think injectedProvider / coinbaseProvider should be `InjectedProvider` and `CoinbaseProvider`, this will make them look more similar to wagmi's connectors as well.

packages/ethers-5/src/client.ts
```diff
@@ -0,0 +1,788 @@
...
+            throw new Error('Chain is not supported')
+          }
+        }
+      }
+    } else if (providerType === INJECTED_CONNECTOR_ID) {
+      const injectedProvider = provider
+      if (injectedProvider) {
+        try {
+          await injectedProvider.send('wallet_switchEthereumChain', [
+            { chainId: `0x${chainId.toString(16)}` }
```
0xAsimetriq, line 628: `0x${chainId.toString(16)}` bit is repeating a lot, could make it a utility. Same for other places that repeat (if any).

packages/ethers-5/src/client.ts
```diff
@@ -0,0 +1,788 @@
...
+          ])
+          ProviderController.setChainId(chainId)
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        } catch (switchError: any) {
+          if (switchError.code === 4902 || switchError.code === 5000) {
```
0xAsimetriq, line 633: Make these codes constants, and descriptive

packages/ethers-5/src/client.ts
```diff
@@ -0,0 +1,788 @@
...
+    } else if (providerType === INJECTED_CONNECTOR_ID) {
+      const injectedProvider = provider
+      if (injectedProvider) {
+        try {
+          await injectedProvider.send('wallet_switchEthereumChain', [
+            { chainId: `0x${chainId.toString(16)}` }
+          ])
+          ProviderController.setChainId(chainId)
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        } catch (switchError: any) {
```
0xAsimetriq, line 632: Can you use `unknonw` instead of any here?

packages/ethers-5/src/client.ts
```diff
@@ -0,0 +1,788 @@
...
+          await EIP6963Provider.send('wallet_switchEthereumChain', [
+            { chainId: `0x${chainId.toString(16)}` }
+          ])
+          ProviderController.setChainId(chainId)
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        } catch (switchError: any) {
+          if (switchError.code === 4902 || switchError.code === 5000) {
+            await EIP6963Provider.send('wallet_addEthereumChain', [
+              {
+                chainId: `0x${chainId.toString(16)}`,
```
0xAsimetriq, line 667: All these payloads are repeated in each provider, can abstract them to utils as well.

packages/ethers-5/src/client.ts
```diff
@@ -0,0 +1,788 @@
...
+              type,
+              imageUrl: info.icon,
+              name: info.name,
+              provider: web3provider,
+              info
+            })
+
+            const name = info.name
+            const eip6963ProviderObj = {
+              name,
```
0xAsimetriq, line 778: just `name: info.name`, no need to separate it into a constant if it is not used multiple times

packages/ethers-5/src/utils/defaultEthersReactConfig.ts
```diff
@@ -0,0 +1,50 @@
+import '@web3modal/polyfills'
```
0xAsimetriq, line 1: We only needed to separate default config into core and react ones because there were slight type differences in `wagmi` and `wagmi/core` packages. Since you are using single ethers package, you just need one file with one `defaultConfig` definition.

packages/utils/index.ts
```diff
@@ -0,0 +1,9 @@
+export * from './src/constants.js'
```
0xAsimetriq, line 1: Let's scope everything like we do in ui and core, i.e. end result when using utils package should be something like

```ts
import { TypeUtil, PresetUtil, ConstUtil } from "@web3modal/utils"
```

packages/utils/package.json
```diff
@@ -0,0 +1,46 @@
...
+    "build": "npm run build:clean; tsc --build",
+    "watch": "npm run build:clean; tsc --watch",
+    "typecheck": "tsc --noEmit",
+    "lint": "eslint . --ext .js,.jsx,.ts,.tsx"
+  },
+  "dependencies": {
+    "@web3modal/polyfills": "3.1.0",
+    "@web3modal/scaffold": "3.1.0"
+  },
+  "peerDependencies": {
```
0xAsimetriq, line 22: This is not gonna work, utils should have no reliance on either wagmi or ethers, so if something in utils is using imports from these, those need to be moved to their respective packages / duplicated with diff types etc if needed.

packages/utils/src/ethers/helpers.ts
```diff
@@ -0,0 +1,22 @@
+import type { CaipNetwork } from '@web3modal/scaffold'
```
0xAsimetriq, line 1: Again, if it is only for ethers or only for wagmi, it should not be in utils, only shared bits should go in here.

packages/utils/src/presets.ts
```diff
@@ -0,0 +1,242 @@
...
+  7777777: 'ETH',
+  // Celo
+  42220: 'CELO',
+  // Base
+  8453: 'BASE',
+  // Aurora
+  1313161554: 'ETH'
+} as Record<string, string>
+
+export const NetworkRPCUrls = {
```
0xAsimetriq, line 175: These should be in ethers as wagmi provides them on chain objects.

packages/utils/src/wagmi/helpers.ts
```diff
@@ -0,0 +1,28 @@
+import type { CaipNetwork } from '@web3modal/scaffold'
```
0xAsimetriq, line 1: Should be in wagmi

packages/wagmi/exports/vue.ts
```diff
@@ -4,27 +4,18 @@ import type {
...
-type ThemeVariablesOptions = Parameters<Web3Modal['setThemeVariables']>[0]
-
-declare module '@vue/runtime-core' {
-  export interface ComponentCustomProperties {
-    W3mConnectButton: Pick<W3mConnectButton, 'size' | 'label' | 'loadingLabel'>
-    W3mAccountButton: Pick<W3mAccountButton, 'disabled' | 'balance'>
-    W3mButton: Pick<W3mButton, 'size' | 'label' | 'loadingLabel' | 'disabled' | 'balance'>
-    W3mNetworkButton: Pick<W3mNetworkButton, 'disabled'>
-  }
+export interface ComponentCustomProperties {
```
0xAsimetriq, line 14: Same comment about these and @vue/runtime-core, as well as moving them to vue scaffold

packages/wagmi/exports/react.tsx
```diff
@@ -6,20 +6,13 @@ import type {
...
 // -- Types -------------------------------------------------------------------
 export type { Web3ModalOptions } from '../src/client.js'
 
-type OpenOptions = Parameters<Web3Modal['open']>[0]
-
-type ThemeModeOptions = Parameters<Web3Modal['setThemeMode']>[0]
-
-type ThemeVariablesOptions = Parameters<Web3Modal['setThemeVariables']>[0]
-
 declare global {
```
0xAsimetriq, line 16: Same comment here as in ethers about moving to react scaffold

0xAsimetriq (changes requested):

examples/vue-wagmi/src/App.vue
```diff
@@ -42,13 +42,16 @@ const modal = useWeb3Modal()
 const state = useWeb3ModalState()
 const { setThemeMode, themeMode, themeVariables } = useWeb3ModalTheme()
 const events = useWeb3ModalEvents()
+
+console.log(themeMode)
```
0xAsimetriq, line 46: Any reason why this example was updated?

packages/ethers-5/exports/react.tsx
```diff
@@ -0,0 +1,64 @@
...
+import { useSnapshot } from 'valtio'
+
+// -- Types -------------------------------------------------------------------
+export type { Web3ModalOptions } from '../src/client.js'
+
+// -- Setup -------------------------------------------------------------------
+let modal: Web3ModalScaffold | undefined = undefined
+
+export function createWeb3Modal(options: Web3ModalOptions) {
+  if (!modal) {
```
0xAsimetriq, line 18: Why is modal init here different from what we do in wagmi, in particular using Web3ModalScaffold type explicitly vs Web3Modal?

packages/ethers-5/src/utils/helpers.ts
```diff
@@ -0,0 +1,76 @@
...
+import {
+  NetworkBlockExplorerUrls,
+  NetworkNames,
+  NetworkRPCUrls,
+  networkCurrenySymbols
+} from './presets.js'
+import type { ethers } from 'ethers'
+import type EthereumProvider from 'node_modules/@walletconnect/ethereum-provider/dist/types/EthereumProvider.js'
+
+export function getCaipDefaultChain(chain?: number) {
```
0xAsimetriq, line 12: Some of these utils look like they are not ethers specific (i.e. don't rely on using ethers or wagmi package), so they can be moved to shared utils? At least CAIP ones?

0xAsimetriq:
Few notes on initial full testing

1. In laboratory when you click on wagmi or ethers example and go to their respective pages, refreshing the page results in `404`. Same when you go straight to the page i.e. via `/library/wagmi`. Unable to check account state persistence due to this atm.
2. I have phantom wallet installed, when I go to ethers example, I am always instantly prompted to connect without performing any action (even with clean localstorage).

0xAsimetriq (changes requested):

packages/ethers-5/exports/react.tsx
svenvoskamp: It did this, because of new `getWeb3Modal(modal)` function. We need this function, so that we have `modal` available in shared scaffold packages. `modal` has shared `Web3ModalScaffold` type, so that is why I'm using `Web3ModalScaffold` type in specific `ethers` and `wagmi` packages. 

Do you think there is a better solution for this?

examples/vue-wagmi/src/App.vue
svenvoskamp: I was testing examples and saw there were a lot of buttons in this example. So I removed a few.

0xAsimetriq: I popped them in there on purpose to see how each variant behaves, could we get them back please.

packages/ethers-5/exports/react.tsx
0xAsimetriq: We should never expose Web3ModalScaffold (client or type) to end user, hence why each of the exports should only define modal as `Web3Modal`. For wagmi and ethers instance of Web3Modal is different, both extend scaffold modal, but have slightly different options and implementations, so here an ethers specific type should be used. If this means duplicating createWeb3Modal package, its fine, because modal instances that are created will be mostly similar for ethers and wagmi, yet different.

packages/ethers-5/src/utils/helpers.ts
svenvoskamp: We need to use specific presets that are `ethers` related, also `wagmi` takes in specific `wagmi` chain. So I think it is better to keep it like this.

packages/ethers-5/src/client.ts
```diff
@@ -0,0 +1,797 @@
...
+    }
+
+    return this.walletConnectProviderInitPromise
+  }
+
+  private async initWalletConnectProvider() {
+    const walletConnectProviderOptions: EthereumProviderOptions = {
+      projectId: this.projectId,
+      showQrModal: false,
+      chains: [0],
```
glitch-txs, line 312:
```suggestion

```

Let's remove required chains completely, some wallets crash with this config. (tested with Trust & MM)

packages/ethers-5/src/client.ts
```diff
@@ -0,0 +1,903 @@
...
+        selectedNetworkId: HelpersUtil.caipNetworkIdToNumber(state.selectedNetworkId)
+      })
+    )
+  }
+
+  public getAddress() {
+    return ProviderController.state.address
+  }
+
+  public getChaindId() {
```
glitch-txs, line 303:
```suggestion
  public getChainId() {
```

small typo

glitch-txs (dismissed):

examples/html-ethers-5/src/main.js
```diff
@@ -0,0 +1,36 @@
...
+if (!projectId) {
+  throw new Error('VITE_PROJECT_ID is not set')
+}
+
+// 2. Create wagmiConfig
+const chains = [1, 42161]
+
+// 3. Create modal
+const modal = createWeb3Modal({
+  ethersConfig: defaultConfig({
```
0xAsimetriq, line 14: Since these are examples, for clarity and consistency with wagmi, we should separate creation of `ethersConfig` into a separate step perhaps?

examples/html-ethers-5/src/main.js
```diff
@@ -0,0 +1,36 @@
...
+    metadata: {
+      name: 'Web3Modal',
+      description: 'Web3Modal Laboratory',
+      url: 'https://web3modal.com',
+      icons: ['https://avatars.githubusercontent.com/u/37784886']
+    },
+    coinbaseDefaultChainId: 1,
```
0xAsimetriq, line 21: Why do we have coinbase explicit options? These should be `defaultChain` and `rpcUrl` and propagate to all default providers. Ideally, we stay as close to current options api as we can.

examples/react-ethers-5/src/App.tsx
```diff
@@ -0,0 +1,65 @@
+import {
```
0xAsimetriq, line 1: Same comments as per html example

examples/vue-ethers-5/src/App.vue
```diff
@@ -0,0 +1,60 @@
+<script lang="ts" setup>
```
0xAsimetriq, line 1: Same comments as per html example

packages/core/src/controllers/NetworkController.ts
```diff
@@ -10,6 +10,8 @@ export interface NetworkControllerClient {
     approvedCaipNetworkIds: NetworkControllerState['approvedCaipNetworkIds']
     supportsAllNetworks: NetworkControllerState['supportsAllNetworks']
   }>
+  // eslint-disable-next-line @typescript-eslint/no-redundant-type-constituents, @typescript-eslint/no-explicit-any
+  getWalletConnectProvider?: () => Promise<any | undefined>
```
0xAsimetriq, line 14: This looks like a pattern going against what we have established via connectors in core. What is this, why is it getting provider in network controller? Can we use existing walletConnect connector pattern?

packages/wagmi/src/utils/provider.ts
```diff
@@ -68,7 +68,7 @@ export function walletConnectProvider<C extends Chain = Chain>({
       return null
     }
 
-    const baseHttpUrl = `${RPC_URL}/v1/?chainId=${NAMESPACE}:${chain.id}&projectId=${projectId}`
+    const baseHttpUrl = `${RPC_URL}/v1/?chainId=${ConstantsUtil.NAMESPACE}:${chain.id}&projectId=${projectId}`
```
0xAsimetriq, line 71: If this is now coming from "outside of evm package", name for the namespace should be more specific i.e `ConstantsUtil.EIP155`

packages/wagmi/src/utils/helpers.ts
```diff
@@ -1,46 +1,15 @@
...
   if (!chain) {
     return undefined
   }
 
   return {
-    id: `${NAMESPACE}:${chain.id}`,
+    id: `${ConstantsUtil.NAMESPACE}:${chain.id}`,
     name: chain.name,
-    imageId: NetworkImageIds[chain.id]
+    imageId: PresetsUtil.NetworkImageIds[chain.id]
```
0xAsimetriq, line 13: Presets should point to a more concrete data set now, i.e. `EIP155NetworkImageIds` etc.., try to get into the mindset where you are working on solana implementation in the future and will be defining new presets / constants etc in utils package.

packages/wagmi/src/client.ts
```diff
@@ -332,41 +320,47 @@ export class Web3Modal extends Web3ModalScaffold {
...
-          })
-          connector.isAuthorized({ info, provider })
-        }
-      })
-      window.dispatchEvent(new Event(EIP6963_REQUEST_EVENT))
+      const handler = this.eip6963EventHandler.bind(this, connector)
+      window.addEventListener(ConstantsUtil.EIP6963_ANNOUNCE_EVENT, handler)
+      window.dispatchEvent(new Event(ConstantsUtil.EIP6963_REQUEST_EVENT))
+      setTimeout(() => {
+        window.removeEventListener(ConstantsUtil.EIP6963_ANNOUNCE_EVENT, handler)
```
0xAsimetriq, line 362: Why are we removing listener here after 100ms? It should be active throughout whole app session. Per eip6963 spec there is no timeframe within which wallets should announce themselves, so we should listen all the time and filter out duplicate events if any happen.

0xAsimetriq (changes requested):
Left few more feedback points and suggestions

apps/laboratory/next.config.js
```diff
@@ -1,7 +1,8 @@
 /** @type {import('next').NextConfig} */
 const nextConfig = {
   reactStrictMode: true,
-  output: 'export'
+  output: 'export',
+  trailingSlash: true
```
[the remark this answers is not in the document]
svenvoskamp: This fixed the refresh `404` issue

glitch-txs (approved):

0xAsimetriq (approved):

shazow:
High level thought: There's very little ethers5-specific code here. Have you considered defining a minimal web3 provider interface that someone can implement to create wrappers for any provider?

For example, what about supporting ethers6?

glitch-txs:
Hi @shazow we are planning to add support for v6 as well, were currently figuring out the best approach

PR #1365, opened 2023-09-29
3.1.0

### LINKED ISSUE #871: Expose modal events: onAction, onError etc...
There currently is no way for users to hook up their own analytics or react to errors that happen within the modal due to it's encapsulated nature. We can solve this by exposing action and error (maybe more in the future) event listeners that can be tapped in, i.e.

```
web3modal.onAction(event => ...)
web3modal.onError(error => ...)
```

### LINKED ISSUE #1371: [bug] The name "w3m-account-button" has already been used with this registry

See https://github.com/WalletConnect/web3modal/issues/1371#issuecomment-1746525872

### Summary

```
Error: Failed to execute 'define' on 'CustomElementRegistry': the name "w3m-account-button" has already been used with this registry
```

It appears that the `@customElement('w3m-account-button')` code is being called twice via `createWeb3Modal`. I'm migrating from NextJS page router to app router and it appears that there may be some nesting going on where both are active at the same time which is causing `createWeb3Modal` to be run twice.

Can that custom element definition be gated by a check that it doesn't exist?

Thank you.

"@web3modal/wagmi": "^3.0.2"

### LINKED ISSUE #1374: [feature] Copy link entry update

With the prominency of the store button at the bottom of the modal and the Copy link used more rarely, we decided to reduce the Copy link component size to keep the modal clean and light

Please update
https://www.figma.com/file/gmX3t1YIFLvXdKEwbeuqEW/Web3Modal?type=design&node-id=6051-216821&mode=design&t=vVpPAJKwcYqFI5Qv-4
image

cc @ignaciosantise @kacperoak @quetool

### LINKED ISSUE #1377: [bug] Allow users to switch to supported chain in single chain setups

n/a

### Summary

Noticed in v3 that if my wallet is connected to unsupported chain, I couldn't been able to switch to supported chain from acccount view. I was able to switch in v2 but not in v3. this is happening only when I have single chain in supported chains list in wagmi config.

Screenshot 2023-10-06 163543

n/a

0xAsimetriq:
# Breaking Changes

N/A

# Changes
- feat: Add web3modal events system
```ts
// Html usage
modal.getEvent() // get last event
modal.subscribeEvents(event => console.log(event)) // subscribe to events

// React usage
import {  useWeb3ModalEvents } from '@web3modal/wagmi/react'
const events = useWeb3ModalEvents()

// Vue usage
import {  useWeb3ModalEvents } from '@web3modal/wagmi/vue'
const events = useWeb3ModalEvents()
```
- feat: Synchronise eip6963 wallets with WalletConnect cloud data through `rdns` value
- feat: Add `EIP6963Connector` to defaults 
- feat: Allow to disable recommended wallets by setting `featuredIds: []` or `customWallets: []` as empty
- feat: Add eip6963 rdns based checks and logic to deduplication utils
- fix: Don't show recommended wallets when `featuredIds` or `customWallets` are set
- fix: Allow to switch network when only 1 network is supported by dapp, but user is on un-supported one
- fix: Create custom `createElement` decorator for "lit" to add a check for already created web-components
- chore: Update styling for "Copy Link" button to make it less prominent
- chore: Rename various utility files for better consistency
- chore: Update dependencies
- chore: Improve ens and balance displays via #1372
- chore: Increase WalletConnect logo size in qrcode view via #1372

# Associated Issues

closes #871
closes #1374
closes #1377 
closes #1371


examples/html-wagmi/index.html
```diff
@@ -1,21 +1,17 @@
-<!-- Test: npx http-server -->
-
 <!doctype html>
 <html lang="en">
   <head>
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
-    <title>Html Wagmi Example</title>
+    <title>Vue wagmi example</title>
```
svenvoskamp, line 6: Is this correct?

svenvoskamp (approved):

svenvoskamp (dismissed):
LFG 🚀

PR #1372, opened 2023-10-04
Fix / ens balance

svenvoskamp:
# Changes

- fix: Adjusted ENS formatting length, address formatting & balance formatting


packages/core/src/utils/CoreHelperUtil.ts
```diff
@@ -129,15 +129,16 @@ export const CoreHelperUtil = {
...
-    if (balance === '0.0') {
+    if (balance === '0') {
       formattedBalance = '0'
-    } else if (typeof balance === 'string' && balance.length > 6) {
-      formattedBalance = balance.substring(0, 6)
     } else if (typeof balance === 'string') {
-      formattedBalance = balance
+      const number = Number(balance)
+      if (number) {
+        formattedBalance = number.toString().match(/^-?\d+(?:\.\d{0,3})?/u)?.[0]
```
ignaciosantise, line 137: what do you think?
`formattedBalance = Number.isInteger(number) ? number.toString() : number.toFixed(3)`

packages/core/src/utils/CoreHelperUtil.ts
```diff
@@ -129,15 +129,16 @@ export const CoreHelperUtil = {
   formatBalance(balance: string | undefined, symbol: string | undefined) {
```
ignaciosantise, line 129: are we also removing decimals from the account view balance? If not, we should receive the decimals as params

da748: I'm not very sure ,but soon we will all find out

svenvoskamp: We are not removing decimals, we only show 3 now

packages/ui/src/utils/UiHelperUtils.ts
```diff
@@ -22,19 +22,20 @@ export const UiHelperUtil = {
     return newUrl.hostname
   },
 
-  getTruncateString(string: string, chars: number, truncate: TruncateType) {
-    if (string.length <= chars) {
+  // eslint-disable-next-line max-params
```
0xAsimetriq, line 25: Remove `max-params` disable and change these to a single object argument i.e `getTruncateString({ string, charsStart ... })` We should do this in cases where there are 2+ func args to avoid mistakes of ordering them incorrectly, especially when they have same types like here `number`

packages/core/src/utils/CoreHelperUtil.ts
```diff
@@ -129,15 +129,16 @@ export const CoreHelperUtil = {
...
     let formattedBalance = undefined
 
-    if (balance === '0.0') {
+    if (balance === '0') {
       formattedBalance = '0'
```
0xAsimetriq, line 133: Should this be `0.000` instead (judging by default return further down below)

packages/core/src/utils/CoreHelperUtil.ts
0xAsimetriq, line 137: Add a comment explaining what's happening here

packages/ui/src/utils/UiHelperUtils.ts
```diff
@@ -22,19 +22,20 @@ export const UiHelperUtil = {
...
-      return `${string.substring(0, chars)}...`
+      return `${string.substring(0, charsStart)}...`
     } else if (truncate === 'start') {
-      return `...${string.substring(string.length - chars)}`
+      return `...${string.substring(string.length - charsEnd)}`
     }
 
-    return `${string.substring(0, Math.floor(chars / 2))}...${string.substring(
-      string.length - Math.floor(chars / 2)
+    return `${string.substring(0, Math.floor(charsStart))}...${string.substring(
```
0xAsimetriq, line 37: To avoid complexity split this into variables i.e. `strStart = string.substring(0, Math.floor(charsStart))` and then `${strStart}...${strEnd}`

0xAsimetriq (changes requested):

0xAsimetriq:
Closing in favor of merging these changes and addressing pr comments in https://github.com/WalletConnect/web3modal/pull/1365

PR #1414, opened 2023-10-23
chore(siwe-client): add SIWE controller & client

Cali93:
# Breaking Changes

N/A

# Changes

- feat: add SIWE controller & Web3ModalSIWEClient

# Associated Issues

closes #...


packages/core/src/controllers/SiweController.ts
```diff
@@ -0,0 +1,103 @@
+import { subscribeKey as subKey } from 'valtio/utils'
```
0xAsimetriq, line 1: We could probably take out some of the types and move them over to TypesUtil in core, keeping state type here only (similar to other controllers).

packages/siwe/package.json
```diff
@@ -0,0 +1,93 @@
...
+  },
+  "scripts": {
+    "build:clean": "rm -rf dist",
+    "build": "npm run build:clean; tsc --build",
+    "watch": "npm run build:clean; tsc --watch",
+    "typecheck": "tsc --noEmit",
+    "lint": "eslint . --ext .js,.jsx,.ts,.tsx"
+  },
+  "dependencies": {
+    "@web3modal/polyfills": "3.2.0-alpha.1",
```
0xAsimetriq, line 46: Check deps / devDeps here, see what you actually need or not (same for peer and optional ones). Some core / reused devDependencies are already defined in root package json.

packages/siwe/src/client.ts
```diff
@@ -0,0 +1,82 @@
...
+
+// -- Types ---------------------------------------------------------------------
+export interface Web3ModalSiweClientOptions extends ScaffoldOptions {
+  siweConfig: SiweControllerClient
+}
+
+export type Web3ModalState = PublicStateControllerState & SiweControllerClientState
+
+// -- Client --------------------------------------------------------------------
+export class Web3ModalSiwe extends Web3ModalScaffold {
```
0xAsimetriq, line 18: This one is interesting, siwe client should be it's own unique thing and not extend Web3ModalScaffold package. Instead siweClient will be passed to scaffold package as an option (of wagmi or ethers packages).

packages/core/src/controllers/SiweController.ts
```diff
@@ -0,0 +1,87 @@
...
```
0xAsimetriq, line 1: Not sure what's going on here, but this file name is `SiweController` while everywhere in the code it is `SIWEController`

packages/core/src/controllers/SiweController.ts
```diff
@@ -0,0 +1,87 @@
...
+import { proxy, subscribe as sub } from 'valtio/vanilla'
+import type {
+  CreateSIWEMessageArgs,
```
0xAsimetriq, line 4: Let's standardise naming for these types i.e. have `SIWE` at the beginning for all i.e. `SIWECreateMessageArgs` / `SIWEVerifyMessageArgs` etc.

packages/core/src/controllers/SiweController.ts
```diff
@@ -0,0 +1,87 @@
...
+  SIWESession,
+  VerifySIWEMessageArgs
+} from '../utils/TypeUtil.js'
+
+import { SIWEStatus } from '../utils/TypeUtil.js'
```
0xAsimetriq, line 9: Second TypeUtil import in the same file, you can combine with previous one. This should've been caught by eslint / prettier though, are they set up properly?

packages/core/src/controllers/SiweController.ts
```diff
@@ -0,0 +1,87 @@
...
+  signOut: () => Promise<boolean>
+
+  // -- Optional --------------------------------------------- //
+
+  // Defaults true
+  enabled?: boolean
+  // In milliseconds, defaults to 5 minutes
+  nonceRefetchInterval?: number
+  // In milliseconds, defaults to 5 minutes
+  sessionRefetchInterval?: number
```
0xAsimetriq, line 26: `sessionRefetchIntervalMs` to signify milliseconds in the name itself (same for other similar bits)

packages/core/src/controllers/SiweController.ts
```diff
@@ -0,0 +1,87 @@
...
+  ) {
+    return subKey(state, key, callback)
+  },
+
+  subscribe(callback: (newState: SIWEControllerClientState) => void) {
+    return sub(state, () => callback(state))
+  },
+
+  setSIWEClient(client: SIWEControllerClientState['_client']) {
+    state._client = client
```
0xAsimetriq, line 68: See `setClient` method in `ConnectionController` i.e. you need to use valtio's `ref()` helper when setting complex / nested values like this one that you don't need to be proxied / reactive.

packages/core/src/utils/TypeUtil.ts
```diff
@@ -200,3 +200,31 @@ export type Event =
...
+  chainId: number
+}
+
+export interface VerifySIWEMessageArgs {
+  message: string
+  signature: string
+}
+
+// eslint-disable-next-line no-shadow
+export enum SIWEStatus {
```
0xAsimetriq, line 223: We try to avoid enums, instead you can do

```
type SIWEStatus = 'uninitialized' | 'ready' | 'loading' ...
```
This gives you better string based autocomplete without the need to import enum every time

packages/siwe/src/client.ts
```diff
@@ -0,0 +1,61 @@
+import {
```
0xAsimetriq, line 1: import type {...} instead of multiple separate type imports

packages/wagmi/package.json
```diff
@@ -51,6 +51,7 @@
   },
   "peerDependencies": {
     "@wagmi/core": ">=1",
+    "@web3modal/siwe": "3.2.0-alpha.1",
```
0xAsimetriq, line 54: This should be added / marked as optional dependency down below.

packages/wagmi/src/client.ts
```diff
@@ -32,6 +34,7 @@ import { WALLET_CHOICE_KEY } from './utils/constants.js'
 export interface Web3ModalClientOptions extends Omit<LibraryOptions, 'defaultChain' | 'tokens'> {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   wagmiConfig: Config<any, any>
+  siweConfig?: SIWEControllerClient
```
0xAsimetriq, line 37: This should probably be of type `Web3ModalSIWEClient` to be more accurate?

packages/wagmi/src/client.ts
```diff
@@ -183,9 +187,16 @@ export class Web3Modal extends Web3ModalScaffold {
       disconnect
     }
 
+    // eslint-disable-next-line @typescript-eslint/no-redundant-type-constituents
+    let siweControllerClient: SIWEControllerClient | undefined = undefined
+    if (siweConfig) {
+      siweControllerClient = new Web3ModalSIWEClient({ siweConfig }).client
```
0xAsimetriq, line 193: SIWE client should not be created here. `@web3modal/siwe` package should export function like `createSiweConfig(...)` that user imports in their dapp and then passes down to wagmi / ethers packages like so (basically same way we pass down wagmi config):

```ts
const wagmiConfig = wagmiConfig(...)
const siweConfig = createSiweConfig(...)
const web3moal = createWeb3Modal({..., siweConfig, wagmiConfig})
```

packages/wagmi/src/client.ts
```diff
@@ -183,9 +187,16 @@ export class Web3Modal extends Web3ModalScaffold {
       disconnect
     }
 
+    // eslint-disable-next-line @typescript-eslint/no-redundant-type-constituents
+    let siweControllerClient: SIWEControllerClient | undefined = undefined
+    if (siweConfig) {
```
0xAsimetriq, line 192: `new Web3ModalSIWEClient({ siweConfig }).client` loos off, we should make it so that `new Web3ModalSIWEClient({ siweConfig })` is the client. Config here should be for options.

0xAsimetriq (changes requested):
Added few comments + ci should be green. Once we get there with these changes, we can probably merge so that this pr doesn't accumulate over time?

packages/core/src/controllers/SiweController.ts
Cali93: I updated the casing for consistency, but I think git doesn't take the change into account if you just update the casing of the file name. Will fix it 👍

packages/core/src/controllers/SiweController.ts
Cali93: `SIWEStatus` is an enum so can't be imported as a type.
I duplicated the import for consistency with current coding styles but I could have a single import like so:
Should I keep it as is or use the snippet below ?

```js
import {
  type SIWECreateMessageArgs,
  type SIWESession,
  type SIWEVerifyMessageArgs,
  SIWEStatus
} from '../utils/TypeUtil.js'
```

packages/wagmi/src/client.ts
Cali93: Not really but I updated the types to be clearer and more flexible.

packages/core/src/controllers/SIWEController.ts
```diff
@@ -0,0 +1,85 @@
+import { subscribeKey as subKey } from 'valtio/utils'
+import { proxy, ref, subscribe as sub } from 'valtio/vanilla'
+import type { SIWEClientMethods, SIWESession } from '../utils/TypeUtil.js'
+
+// -- Types --------------------------------------------- //
+// eslint-disable-next-line no-shadow
+export enum SIWEStatus {
```
0xAsimetriq, line 7: Don't use enums if possible, instead of this enum do following in `SIWEControllerClientState`

```
SIWEControllerClientState {
  status: 'uninitialized' | 'ready' | 'loading'
}
```

This way you will not need to export / import `SIWEStatus` anywhere and will still get autocomplete and type validation on status. If you absolutely need to use enums, I think they need to be set via `export const enum ...` to work properly when built for js.

packages/core/tests/controllers/SIWEController.test.ts
```diff
@@ -0,0 +1,63 @@
+import { describe, expect, it } from 'vitest'
+import { SIWEController, SIWEStatus, type SIWEControllerClient } from '../../index.js'
+
+// -- Mocks -------------------------------------------------------------
+const session = { address: '0x', chainId: 1 }
+const client: SIWEControllerClient = {
```
0xAsimetriq, line 6: You don't need explicit `: SIWEControllerClient` here. If you were to pass invalid type to `SIWEController.setSIWEClient` that's where type error should appear.

packages/siwe/exports/index.ts
```diff
@@ -0,0 +1,9 @@
+import type { SIWEConfig } from '@web3modal/core'
+import { Web3ModalSIWEClient } from '../src/client.js'
+
+export { Web3ModalSIWEClient }
```
0xAsimetriq, line 4: you can remove imports and just do `export {SIWEConfig} from "@web3modal/core"`. Same for others. But looking further down the file, I think only thing you want to export from this file is `createSIWEConfig`?

And if you still need to export `Web3ModalSIWEClient` and `SIWEConfig` you can do so directly from package index.

0xAsimetriq (changes requested):

packages/siwe/exports/index.ts
Cali93: The use case I'm thinking about here for exporting `type SIWEConfig` is for developers to be able to define a siweConfig outside of `createSIWEConfig` and still get type safety.
See example below:

```ts
// SIWEConfig can be auto-imported because it's exported by @web3modal/siwe
const mySiweConfig: SIWEConfig = {
  createMessage: () => console.log('create SIWE message')
  // ... get type safety and auto complete on other options
}

createSIWEConfig(mySiweConfig)
```

0xAsimetriq (approved):
LGTM 🔐

packages/scaffold/src/client.ts
```diff
@@ -196,6 +228,9 @@ export class Web3ModalScaffold {
 
     ConnectionController.setClient(options.connectionControllerClient)
 
+    if (options.siweControllerClient) {
```
0xAsimetriq, line 231: if SIWEControler can be undefined (optional) you can remove an if statement

packages/siwe/src/client.ts
```diff
@@ -0,0 +1,84 @@
...
+  SIWECreateMessageArgs,
+  SIWEVerifyMessageArgs,
+  SIWEControllerClient,
+  SIWEConfig,
+  SIWEClientMethods
+} from '@web3modal/core'
+
+import { ConstantsUtil } from './utils/ConstantsUtil.js'
+
+// -- Types --------------------------------------------- //
```
0xAsimetriq, line 11:
```suggestion
```

packages/wagmi/package.json
```diff
@@ -43,6 +43,7 @@
     "lint": "eslint . --ext .js,.jsx,.ts,.tsx"
   },
   "dependencies": {
+    "@web3modal/siwe": "3.2.0",
```
0xAsimetriq, line 46: `@web3modal/siwe` is a peer / optional dependency, so we shouldn't include it here. For dev purposes (i.e. types) we can add it to devDependencies

packages/scaffold/src/client.ts
Cali93: `SIWEController.setSIWEClient(client)` expects `client` to be defined.

0xAsimetriq (approved):

PR #1416, opened 2023-10-25
Siwe / UI 

svenvoskamp:
# How to test? 

The SIWE view will pop up after you've connected

# Changes

* Added `w3m-connecting-siwe-view`
* Added `w3m-connecting-siwe` partial
* Added `wui-thumbnail-visual` composite
* Added few colors
* Added few spacings
* Adjusted `wui-button` component 
* Adjusted `wui-text` component
* Added `setMetadata` function to `scaffold` client
* Added metadata to `OptionsController`
* Added `setImageUrl` and `getImageUrl` to `StorageUtil`


packages/scaffold/src/partials/w3m-connecting-siwe/index.ts
```diff
@@ -0,0 +1,52 @@
+import { customElement } from '@web3modal/ui'
+import { LitElement, html } from 'lit'
+import { property } from 'lit/decorators.js'
+import styles from './styles.js'
+
+@customElement('w3m-connecting-siwe')
+export class W3mConnectingSiwe extends LitElement {
+  public static override styles = [styles]
```
0xAsimetriq, line 8:
```suggestion
  public static override styles = styles
```

packages/scaffold/src/partials/w3m-connecting-siwe/index.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  // -- State & Properties -------------------------------- //
+  @property() public dappImageSrc?: string
+
+  @property() public walletImageSrc?: string
+
+  public override firstUpdated() {
+    const visuals = this.shadowRoot?.querySelectorAll('wui-visual-thumbnail')
+
+    if (visuals?.[0]) {
```
0xAsimetriq, line 18: You could make a private function that takes in a visual and applies these. To deduplicate

packages/scaffold/src/partials/w3m-header/index.ts
```diff
@@ -77,9 +78,11 @@ export class W3mHeader extends LitElement {
   }
 
   // -- Private ------------------------------------------- //
+
+  // Tempory added to test connecting with SIWE, replace with 'WhatIsAWallet' again when approved
   private onWalletHelp() {
     EventsController.sendEvent({ type: 'track', event: 'CLICK_WALLET_HELP' })
-    RouterController.push('WhatIsAWallet')
+    RouterController.push('ConnectingSiwe')
```
0xAsimetriq, line 85: Remember to revert before merging

packages/scaffold/src/views/w3m-connecting-siwe-view/index.ts
```diff
@@ -0,0 +1,68 @@
...
+import { LitElement, html } from 'lit'
+
+@customElement('w3m-connecting-siwe-view')
+export class W3mConnectingSiweView extends LitElement {
+  // -- Render -------------------------------------------- //
+  public override render() {
+    return html`
+      <wui-flex justifyContent="center" .padding=${['2xl', '0', 'xxl', '0'] as const}>
+        <w3m-connecting-siwe
+          dappImageSrc="https://explorer-api.walletconnect.com/v3/logo/lg/bff9cf1f-df19-42ce-f62a-87f04df13c00?projectId=2f05ae7f1116030fde2d36508f472bfb"
```
0xAsimetriq, line 12: We need to think ahead and see where these dapp and wallet images will come from. And ensure they have fallbacks when not defined.

For dapp, I think we can use icon from metadata option.
For wallet we need to make a flow that stores image in localStorage after succesful connection

Might be worth implementing these now, as they are not directly related to SIWE. You can reference logic where we set recent wallet and add wallet image storage at the same time (but will also need to add this to injected wallets and perhaps other custom connectors as we only store walletconnect connections as recent)

packages/scaffold/src/views/w3m-connecting-siwe-view/index.ts
```diff
@@ -0,0 +1,68 @@
...
+        <wui-button size="md" ?fullwidth=${true} variant="fill" @click=${this.onSign.bind(this)}>
+          Sign
+        </wui-button>
+      </wui-flex>
+    `
+  }
+
+  // -- Private ------------------------------------------- //
+  private onDappLink() {
+    CoreHelperUtil.openHref('https://app.uniswap.org', '_blank')
```
0xAsimetriq, line 53: Same as above, where would dapp link come from? We need to start storing it.

packages/ui/src/composites/wui-visual-thumbnail/index.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  @property({ type: Boolean }) public borderRadiusFull?: boolean
+
+  // -- Render -------------------------------------------- //
+  public override render() {
+    this.style.cssText = `
+    --local-border-radius: ${this.borderRadiusFull ? '1000px' : '20px'};
+`
+
+    return html` ${this.templateVisual()}`
```
0xAsimetriq, line 26:
```suggestion
    return html`${this.templateVisual()}`
```

0xAsimetriq (changes requested):

0xAsimetriq:
Also, while playing with the preview, when I went to siwe view, images took split sec to appear, we should try to preload these if possible beforehand

packages/scaffold/src/views/w3m-connecting-siwe-view/index.ts
```diff
@@ -0,0 +1,68 @@
...
+          <wui-icon size="sm" color="inherit" slot="iconRight" name="externalLink"></wui-icon>
+        </wui-button>
+      </wui-flex>
+      <wui-flex
+        .padding=${['0', '3xl', 'l', '3xl'] as const}
+        gap="s"
+        justifyContent="space-between"
+      >
+        <wui-text variant="small-400" align="center" color="fg-200"
+          >Sign this message to prove you owns this wallet and to continue</wui-text
```
arein, line 37:
```suggestion
          >Sign this message to prove you own this wallet and to continue</wui-text
```

packages/scaffold/src/partials/w3m-header/index.ts
```diff
@@ -22,6 +22,7 @@ function headings() {
     Account: undefined,
     ConnectingExternal: name ?? 'Connect Wallet',
     ConnectingWalletConnect: name ?? 'WalletConnect',
+    ConnectingSiwe: 'Sign In With Ethereum',
```
arein, line 25:
```suggestion
    ConnectingSiwe: 'Sign In',
```

arein (approved):

Cali93 (dismissed):

packages/scaffold/src/views/w3m-connecting-siwe-view/index.ts
da748: Automatic connect

packages/core/src/utils/StorageUtil.ts
```diff
@@ -4,6 +4,7 @@ import type { WcWallet } from './TypeUtil.js'
 // -- Helpers -----------------------------------------------------------------
 const WC_DEEPLINK = 'WALLETCONNECT_DEEPLINK_CHOICE'
 const W3M_RECENT = '@w3m/recent'
+const W3M_IMAGE_URL = '@3m/image_url'
```
0xAsimetriq, line 7: Let's be more expressive here, which image is it (connected wallet logo / dapp logo?).

packages/core/src/utils/StorageUtil.ts
```diff
@@ -62,5 +63,13 @@ export const StorageUtil = {
     }
 
     return []
+  },
+
+  setImageUrl(imageUrl: string) {
```
0xAsimetriq, line 68: As per above, add more expressive function names + try/catch blocks as with all other functions that operate on `localStorage`. Some weird edge cases on android devices exist where localStorage is not available or dapp's block it entirely for security reasons.

packages/ethers5/src/client.ts
```diff
@@ -236,6 +236,7 @@ export class Web3Modal extends Web3ModalScaffold {
     this.options = options
 
     this.metadata = ethersConfig.metadata
+    this.setMetadata(this.metadata)
```
0xAsimetriq, line 239: Metadata should be set as any other option and then passed to options controller inside scaffold.initControllers method. Why do we need to create and use separate method for it here?

packages/scaffold/src/client.ts
```diff
@@ -177,6 +178,10 @@ export class Web3ModalScaffold {
       AccountController.setAddressExplorerUrl(addressExplorerUrl)
     }
 
+  protected setMetadata(metadata: OptionsControllerState['metadata']) {
```
0xAsimetriq, line 181: As per above comment, this should not be needed.

packages/scaffold/src/partials/w3m-connecting-siwe/index.ts
```diff
@@ -0,0 +1,52 @@
...
+import { LitElement, html } from 'lit'
+import { property } from 'lit/decorators.js'
+import styles from './styles.js'
+
+@customElement('w3m-connecting-siwe')
+export class W3mConnectingSiwe extends LitElement {
+  public static override styles = styles
+
+  // -- State & Properties -------------------------------- //
+  @property() public dappImageSrc?: string
```
0xAsimetriq, line 11: As this is a partial, it can access these directly, so no need to pass them down and create extra "prop drilling" complexity. properties should mostly be on "simple components".

packages/scaffold/src/partials/w3m-connecting-wc-browser/index.ts
```diff
@@ -38,7 +38,8 @@ export class W3mConnectingWcBrowser extends W3mConnectingWidget {
       } else if (injectedConnector) {
         await ConnectionController.connectExternal(injectedConnector)
       }
-      ModalController.close()
+      // ModalController.close()
```
0xAsimetriq, line 41: I think this is for testing purposes only atm, we should always add that info within comment as well.

packages/scaffold/src/views/w3m-connecting-siwe-view/index.ts
```diff
@@ -0,0 +1,110 @@
...
+import { LitElement, html } from 'lit'
+import { state } from 'lit/decorators.js'
+
+@customElement('w3m-connecting-siwe-view')
+export class W3mConnectingSiweView extends LitElement {
+  // -- Members -------------------------------------------- //
+  private usubscribe: (() => void)[] = []
+
+  // -- State & Properties --------------------------------- //
+  @state() private dappIcons = OptionsController.state.metadata?.icons
```
0xAsimetriq, line 12: These can be non-state / readonly variables. They are not expected to change once dapp has initialized web3modal. Thus, no subscribe / unsubscribe is needed either.

packages/scaffold/src/views/w3m-connecting-siwe-view/index.ts
```diff
@@ -0,0 +1,110 @@
...
+          this.dappName = metadata.name
+        }
+      })
+    )
+  }
+
+  // -- Render -------------------------------------------- //
+
+  public override render() {
+    const imageUrl = StorageUtil.getImageUrl()
```
0xAsimetriq, line 39: This can can also be a readonly global, instead of accessing localStorage within each render call.

packages/scaffold/src/views/w3m-connecting-siwe-view/index.ts
```diff
@@ -0,0 +1,110 @@
...
+
+    return html`
+      <wui-flex justifyContent="center" .padding=${['2xl', '0', 'xxl', '0'] as const}>
+        <w3m-connecting-siwe
+          .dappImageSrc=${this.dappIcons ? this.dappIcons[0] : undefined}
```
0xAsimetriq, line 44:
```suggestion
          .dappImageSrc=${this.dappIcons?.[0]}
```

packages/wagmi/src/client.ts
```diff
@@ -194,6 +194,8 @@ export class Web3Modal extends Web3ModalScaffold {
 
     this.options = options
 
+    this.createMetadata(wagmiConfig)
```
0xAsimetriq, line 197: Why are we accessing and setting metadata from walletconnect connector, when it is available as an option? Plus same comment as in ethers package, this should be set from within scaffold's init method.

0xAsimetriq (changes requested):

packages/ethers5/src/client.ts
0xAsimetriq, line 239: You've now exposed metadata as option on createWeb3Modal, is this function still needed? I'm assuming we can now access metadata as and ordinary option and set it in scaffold.

packages/scaffold/src/client.ts
0xAsimetriq, line 181: Similar to comment above is this now needed? You have metadata available on options and can call `OptionsController.setMetadata(options.metadata)` inside `initControllers`

packages/scaffold/src/partials/w3m-connecting-siwe/index.ts
```diff
@@ -0,0 +1,53 @@
...
+import { property } from 'lit/decorators.js'
+import styles from './styles.js'
+import { OptionsController, StorageUtil } from '@web3modal/core'
+
+@customElement('w3m-connecting-siwe')
+export class W3mConnectingSiwe extends LitElement {
+  public static override styles = styles
+
+  // -- State & Properties -------------------------------- //
+  @property({ type: Array }) private dappImageUrl = OptionsController.state.metadata?.icons
```
0xAsimetriq, line 12: these don't need to be properties, just `private readonly dappImageUrl...`

packages/scaffold/src/views/w3m-connecting-siwe-view/index.ts
```diff
@@ -0,0 +1,81 @@
+import { CoreHelperUtil, OptionsController, RouterController } from '@web3modal/core'
+import { customElement } from '@web3modal/ui'
+import { LitElement, html } from 'lit'
+import { property } from 'lit/decorators.js'
+
+@customElement('w3m-connecting-siwe-view')
+export class W3mConnectingSiweView extends LitElement {
+  // -- State & Properties --------------------------------- //
+  @property() private dappUrl = OptionsController.state.metadata?.url
```
0xAsimetriq, line 9: Don't need to be properties, just readonly private vars

packages/ui/src/composites/wui-visual-thumbnail/index.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  @property() public alt?: string
+
+  @property({ type: Boolean }) public borderRadiusFull?: boolean
+
+  // -- Render -------------------------------------------- //
+  public override render() {
+    this.style.cssText = `
+    --local-border-radius: ${this.borderRadiusFull ? '1000px' : '20px'};
+`
```
0xAsimetriq, line 24: Formatting looks off

packages/wagmi/src/client.ts
```diff
@@ -226,6 +228,15 @@ export class Web3Modal extends Web3ModalScaffold {
   }
 
   // -- Private -----------------------------------------------------------------
+  private createMetadata(wagmiConfig: Web3ModalClientOptions['wagmiConfig']) {
```
0xAsimetriq, line 231: Remove in favor of using metadata from options now.

0xAsimetriq (changes requested):

packages/scaffold/src/partials/w3m-connecting-siwe/index.ts
```diff
@@ -0,0 +1,53 @@
...
+import { LitElement, html } from 'lit'
+
+import styles from './styles.js'
+import { OptionsController, StorageUtil } from '@web3modal/core'
+
+@customElement('w3m-connecting-siwe')
+export class W3mConnectingSiwe extends LitElement {
+  public static override styles = styles
+
+  // -- State & Properties -------------------------------- //
```
0xAsimetriq, line 11: `// -- State & Properties -------------------------------- //` should be replaced with `// -- Members -------------------------------- //` as everywhere else that uses local vars.

See if this applies in other places where you made properties local vars.

packages/scaffold/src/client.ts
0xAsimetriq, line 181: This function is still here, but it is not needed, correct? You set metadata in `initControllers`

0xAsimetriq (changes requested):

PR #1423, opened 2023-10-27
Ethers / chains

svenvoskamp:
# Changes

* Change chain format to the following:

```
 type Chain = {
  rpcUrl: string
  explorerUrl: string
  currency: string
  name: string
  chainId: number
}
```


glitch-txs (approved):

thyoondev:
When will this feature be released?

0xAsimetriq:
@thyoondev keep an eye on #1424

bel0v:
seems like this example needs to be updated? 
https://github.com/WalletConnect/web3modal/blob/V3/examples/react-ethers5/src/App.tsx

PR #1440, opened 2023-11-03
feat: transaction list

enesozturk:
# Overview

This pull request introduces the new transaction list feature within our web3modal interface. Aimed at increasing user engagement and providing transparent financial tracking, this update allows users to easily access and review their recent transactions directly from their web3modal.

[Designs](https://www.figma.com/file/gmX3t1YIFLvXdKEwbeuqEW/Web3Modal-v3?type=design&node-id=5326%3A211816&mode=design&t=jORsR9VY0ha6wZSB-1)

## Technical Details

- [x] New list item on the account view. 
- [x] Integration with blockchain APIs to fetch transaction data.
- [x] Transaction list view and list all the transactions depending on the response
- [x] Pagination to handle large transaction histories efficiently.

### Screenshots

https://github.com/WalletConnect/web3modal/assets/19428358/9ca3a597-83e1-4dd1-9fb1-e97b94dfa2e2


packages/ui/src/utils/TypeUtil.ts
```diff
@@ -169,21 +170,32 @@ export type ChipType = 'fill' | 'shade' | 'transparent'
...
-  | 'bought'
-  | 'buy'
-  | 'cryptoSent'
-  | 'deposited'
-  | 'minted'
-  | 'nftSent'
-  | 'received'
-  | 'swapped'
-  | 'withdrawed'
+  | 'approve' // in
```
enesozturk, line 173: remove comments

packages/core/src/controllers/TransactionsController.ts
```diff
@@ -0,0 +1,49 @@
...
+      .then(response => {
+        state.loading = false
+        state.transactions = response.data
+        state.empty = response.data.length === 0
+        state.next = response.next
+      })
+      .catch(error => {
+        state.loading = false
+        state.empty = true
+        console.log(error)
```
enesozturk, line 46: console

packages/core/src/utils/TypeUtil.ts
```diff
@@ -149,12 +149,18 @@ export interface TransactionMetadata {
 }
 
 export interface TransactionTransfer {
-  fungible_info?: any
+  fungible_info?: {
+    name: string
```
geekbrother, line 153: `name` and `symbol` are optional as well

packages/core/src/utils/TypeUtil.ts
```diff
@@ -169,12 +175,12 @@ export interface TransactionContent {
 
 export interface TransactionPreview {
   url: string
-  content_type?: any
+  content_type?: null
```
geekbrother, line 178: I think it can be a video object as well or gif, so possible we should check for the MIME type (eg. `image/png`)

packages/core/tests/controllers/TransactionsController.test.ts
```diff
@@ -0,0 +1,15 @@
+import { describe, expect, it } from 'vitest'
```
enesozturk, line 1: This test is failing due to a package resolve issue in CI and cannot reproduce it in local. 

Might there be some caching issue since we're using Turbo? I had times that it's causing this when we use remote cache feature

@0xAsimetriq 

- https://github.com/WalletConnect/web3modal/actions/runs/6812592630/job/18525268647?pr=1440
- https://turbo.build/repo/docs/core-concepts/remote-caching

packages/scaffold/src/partials/w3m-all-wallets-search/index.ts
```diff
@@ -45,8 +45,8 @@ export class W3mAllWalletsSearch extends LitElement {
         <wui-flex justifyContent="center" alignItems="center" gap="s" flexDirection="column">
           <wui-icon-box
             size="lg"
-            iconcolor="fg-200"
-            backgroundcolor="fg-300"
+            iconColor="fg-200"
```
enesozturk, line 48: Lets keep the props names and usage consistent

packages/ui/src/composites/wui-transaction-list-item-loader/styles.ts
```diff
@@ -0,0 +1,14 @@
+import { css } from 'lit'
+
+export default css`
+  :host > wui-flex:first-child {
```
enesozturk, line 4: I still might need to search for the best approaches to writing CSS with web components. Somehow I don't really feel quite quite good in some cases. 

Please if you see some bad practice in the CSS, let me know.

packages/ui/src/composites/wui-transaction-list-item/index.ts
```diff
@@ -0,0 +1,83 @@
+import { html, LitElement } from 'lit'
```
enesozturk, line 1: Imports looking a bit chaotic imo. 

@0xAsimetriq What about using Trivago's import sorter? I'm using it for almost every project and love it.

- https://github.com/trivago/prettier-plugin-sort-imports

packages/ui/src/composites/wui-transaction-list-item/index.ts
```diff
@@ -0,0 +1,83 @@
...
+            >${TransactionTypePastTense[this.type]}</wui-text
+          >
+          <wui-flex class="description-container">
+            ${firstDesc
+              ? html`<wui-text variant="small-500" color="fg-200">
+                  <span>${firstDesc}</span>
+                </wui-text>`
+              : null}
+            ${secondDesc
+              ? html`
```
enesozturk, line 58: No need nested `html` and might be anti pattern

packages/ui/src/composites/wui-transaction-visual/index.ts
```diff
@@ -2,71 +2,120 @@ import { html, LitElement } from 'lit'
...
+  @property({ type: Boolean }) public isNFT?: boolean
 
-  @property() public imageSrc?: string
+  @property() public imageURL?: string
+
+  @property() public secondImageURL?: string
 
   // -- Render -------------------------------------------- //
   public override render() {
+    this.style.cssText = `--local-border-radius: ${this.isNFT ? '14px' : '50%'};`
```
enesozturk, line 30: Use css values here

packages/ui/src/composites/wui-transaction-list-item/index.ts
svenvoskamp: I think you can use `nothing` for this: https://lit.dev/docs/v1/lit-html/writing-templates/#rendering-nothing

apps/gallery/stories/composites/wui-transaction-list-item.stories.ts
```diff
@@ -0,0 +1,59 @@
...
+} from '../../utils/PresetUtils'
+
+type Component = Meta<WuiTransactionListItem>
+
+export default {
+  title: 'Composites/wui-transaction-list-item',
+  args: {
+    type: 'approve',
+    imageURL:
+      'https://uploads-ssl.webflow.com/61fe6bb74158b468a1112105/61fe8900148b6e6ee7c2c4e4_Profile%20Pic_Smoker-p-500.jpeg',
```
svenvoskamp, line 19: Can you add this imageURL to `PresetUtils`, so this file stays more clean?

apps/gallery/stories/composites/wui-transaction-visual.stories.ts
```diff
@@ -2,23 +2,42 @@ import type { Meta } from '@storybook/web-components'
...
 type Component = Meta<WuiTransactionVisual>
 
 export default {
   title: 'Composites/wui-transaction-visual',
   args: {
-    type: 'bought',
-    imageSrc:
-      'https://uploads-ssl.webflow.com/61fe6bb74158b468a1112105/61fe8900148b6e6ee7c2c4e4_Profile%20Pic_Smoker-p-500.jpeg'
+    type: 'approve',
+    imageURL:
```
svenvoskamp, line 18: Same as comment above ^

packages/core/tests/controllers/TransactionsController.test.ts
svenvoskamp: Can you try updating `common` package to `3.2.1`?

packages/ui/src/composites/wui-icon-box/index.ts
```diff
@@ -33,40 +35,29 @@ export class WuiIconBox extends LitElement {
 
   // -- Render -------------------------------------------- //
   public override render() {
-    let iconSize: SizeType = 'xxs'
```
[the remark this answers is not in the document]
svenvoskamp: This component got more extensive over time due to design changes etc. We might need to refactor this component in the future.

packages/common/index.ts
```diff
@@ -0,0 +1,4 @@
+// -- Utils -------------------------------------------------------------------
```
0xAsimetriq, line 1: Note for follow up PR's, not this one, we should also probably start moving utils from core and ui here

packages/common/package.json
```diff
@@ -0,0 +1,41 @@
+{
+  "name": "@web3modal/common",
+  "version": "3.2.0",
```
0xAsimetriq, line 3: I think there is a version miss-alignment here with other packages (3.2.1 is latest), this could be causing issues with tests as Sven suggested.

packages/ui/src/composites/wui-transaction-list-item/index.ts
svenvoskamp: +1 imo

packages/common/package.json
```diff
@@ -0,0 +1,41 @@
...
+    "watch": "npm run build:clean; tsc --watch",
+    "typecheck": "tsc --noEmit",
+    "test": "vitest run --dir tests",
+    "lint": "eslint . --ext .js,.jsx,.ts,.tsx"
+  },
+  "dependencies": {
+    "dayjs": "1.11.10"
+  },
+  "keywords": [
+    "web3",
```
0xAsimetriq, line 23: Adjust keywords to match the package a bit more i.e. lit and webcomponents are not needed here probably

packages/common/tests/DateUtil.test.ts
```diff
@@ -0,0 +1,16 @@
+import { describe, expect, it } from 'vitest'
+import { DateUtil } from '../src/utils/DateUtil.js'
+
+const date = new Date()
+const year = new Date().getFullYear()
+const oneHourAgo = new Date(date.setHours(date.getHours() - 1)).toISOString()
+
+// -- Tests --------------------------------------------------------------------
+describe('DateUtil', () => {
```
0xAsimetriq, line 9: Not 100% sure about this, but by the time you declare `date` above, it might be out of sync with tests already. If it is the case you can use some of vitests tools to "mock" `Date` and set it to be a frozen value while they run.

packages/core/src/controllers/TransactionsController.ts
```diff
@@ -0,0 +1,93 @@
...
+
+// -- Types --------------------------------------------- //
+type TransactionByYearMap = Record<number, Transaction[]>
+
+export interface TransactionsControllerState {
+  transactions: Transaction[]
+  transactionsByYear: TransactionByYearMap
+  loading: boolean
+  empty: boolean
+  next: string | null
```
0xAsimetriq, line 15:
```suggestion
  next?: string
```

We try to stay away from nulls and prefer `undefined`

packages/core/src/controllers/TransactionsController.ts
```diff
@@ -0,0 +1,93 @@
...
+      throw new Error("Transactions can't be fetched without a projectId and an accountAddress")
+    }
+
+    state.loading = true
+    await BlockchainApiController.fetchTransactions({
+      account: accountAddress,
+      projectId,
+      cursor: state.next
+    })
+      .then(response => {
```
0xAsimetriq, line 49: Split out into 2 awaits and wrap in `try / catch / finaly` block to follow other async actions style.

packages/core/src/utils/TypeUtil.ts
```diff
@@ -121,6 +123,17 @@ export interface BlockchainApiIdentityResponse {
   name: string
 }
 
+export interface BlockchainApiTransactionsRequest {
+  account: string
+  projectId: string
+  cursor: string | null
```
0xAsimetriq, line 129:
```suggestion
  cursor?: string
```

packages/ui/src/composites/wui-transaction-visual/styles.ts
svenvoskamp, line 1: Do we need to put `:host` in front of all the selectors? I thought it was directly accessible

svenvoskamp, line 1: Same comment as above ^

packages/scaffold/src/views/w3m-transactions-view/index.ts
```diff
@@ -0,0 +1,211 @@
...
+    this.paginationObserver = new IntersectionObserver(([element]) => {
+      if (element?.isIntersecting && !this.loading) {
+        this.fetchTransactions()
+        EventsController.sendEvent({ type: 'track', event: 'LOAD_MORE_TRANSACTIONS' })
+      }
+    }, {})
+  }
+
+  // Private Methods ------------------------------------- //
+  private async fetchTransactions() {
```
0xAsimetriq, line 178: Looks like un-necessary method wrapping a 1 liner?

packages/ui/src/composites/wui-transaction-list-item-loader/index.ts
```diff
@@ -0,0 +1,31 @@
+import { html, LitElement } from 'lit'
+import '../../components/wui-text/index.js'
+import { elementStyles, resetStyles } from '../../utils/ThemeUtil.js'
+import { customElement } from '../../utils/WebComponentsUtil.js'
+import '../wui-transaction-visual/index.js'
+import styles from './styles.js'
+
+@customElement('wui-transaction-list-item-loader')
+export class WuiTransactionListItemLoader extends LitElement {
+  public static override styles = [resetStyles, elementStyles, styles]
```
0xAsimetriq, line 10: I don't think you need all of these styles here i.e. element ones are when we use things like buttons / forms etc. but just in case double verify 😅 been a while since I touched this area.

packages/ui/src/composites/wui-transaction-list-item/index.ts
0xAsimetriq: So, we actually tried this through prettier / as eslint rules and it created more issues than it solved, people didn't have certain plugins installed / reorganizing these caused a bunch of merge conflicts etc. So for now lets keep import and prop ordering relaxed. The only thing that we try to stick with is separate type imports i.e. using `import type {} ...` this will be caught by tsc.

packages/ui/src/composites/wui-transaction-list-item/index.ts
```diff
@@ -0,0 +1,81 @@
+import { html, LitElement } from 'lit'
+import { property } from 'lit/decorators.js'
+import '../../components/wui-text/index.js'
+import { elementStyles, resetStyles } from '../../utils/ThemeUtil.js'
+import { customElement } from '../../utils/WebComponentsUtil.js'
+import '../wui-transaction-visual/index.js'
+import styles from './styles.js'
+import { type TransactionType, TransactionTypePastTense } from '../../utils/TypeUtil.js'
+import { type TransactionStatus, type TransactionDirection } from '@web3modal/common'
```
0xAsimetriq, line 9:
```suggestion
import type { TransactionStatus, TransactionDirection } from '@web3modal/common'
```

packages/ui/src/composites/wui-transaction-list-item/index.ts
```diff
@@ -0,0 +1,81 @@
...
+
+@customElement('wui-transaction-list-item')
+export class WuiTransactionListItem extends LitElement {
+  public static override styles = [resetStyles, elementStyles, styles]
```
0xAsimetriq, line 13: Same as above (and I guess for any other cases where style array is present) verify that all of these are necessary

packages/ui/src/composites/wui-transaction-list-item/index.ts
0xAsimetriq: You can split out into template methods to cleanup a bit.

packages/ui/src/composites/wui-transaction-visual/index.ts
```diff
@@ -2,71 +2,120 @@ import { html, LitElement } from 'lit'
 import { property } from 'lit/decorators.js'
 import '../../components/wui-image/index.js'
 import { resetStyles } from '../../utils/ThemeUtil.js'
-import type { TransactionIconType, TransactionType } from '../../utils/TypeUtil.js'
+import { type TransactionDirection, type TransactionStatus } from '@web3modal/common'
```
0xAsimetriq, line 5:
```suggestion
import type { TransactionDirection, TransactionStatus } from '@web3modal/common'
```

packages/ui/src/composites/wui-transaction-visual/index.ts
```diff
@@ -2,71 +2,120 @@ import { html, LitElement } from 'lit'
...
+import { type TransactionIconType, type TransactionType } from '../../utils/TypeUtil.js'
```
0xAsimetriq, line 6:
```suggestion
import type { TransactionIconType, TransactionType } from '../../utils/TypeUtil.js'
```

packages/ui/src/utils/TransactionUtil.ts
```diff
@@ -0,0 +1,126 @@
...
+const FLOAT_FIXED_VALUE = 3
+
+export const TransactionUtil = {
+  getTransactionGroupTitle(year: number) {
+    const currentYear = DateUtil.getYear()
+    const isCurrentYear = year === currentYear
+    const groupTitle = isCurrentYear ? 'This Year' : year
+
+    return groupTitle
+  },
```
0xAsimetriq, line 15: Add line breaks for better readability after each method

0xAsimetriq (changes requested):
Vey solid first contribution 🚀 Left some ideas and suggestions

0xAsimetriq:
I ran into an issue with this flow

1. Connected with my metamask extension and loaded up the transactions
2. DIsconnected
3. Connected using zerion through qrcode
4. Going to transactions now shows me my old list from metamask / is not loading new data

apps/gallery/stories/composites/wui-transaction-list-item.stories.ts
enesozturk: Definitely

packages/common/index.ts
enesozturk: Definitely, let me note this. and see what we can move from there.

packages/common/package.json
enesozturk: Added `utils`, `types`, `web3`, `web3modal`

enesozturk:
> I ran into an issue with this flow
> 
> 1. Connected with my metamask extension and loaded up the transactions
> 2. DIsconnected
> 3. Connected using zerion through qrcode
> 4. Going to transactions now shows me my old list from metamask / is not loading new data

This should be fixed now

packages/core/src/controllers/TransactionsController.ts
enesozturk: @0xAsimetriq Can you check this again please?

packages/common/tests/DateUtil.test.ts
enesozturk: Mocked with `vi.setSystemTime`. Thanks 💯

- https://vitest.dev/api/vi#vi-setsystemtime

enesozturk:
@svenvoskamp @0xAsimetriq Seems like we still have the test error in CI. I cannot reproduce this locally. I'm still looking for it.

```js
Error: Failed to resolve entry for package "@web3modal/common". The package may have incorrect main/module/exports specified in its package.json.
```

I'm planning to use `act` to debug the issue but since I don't have env variables, it might not be as easy as I think. Did you use it before?

- https://github.com/nektos/act

packages/core/src/controllers/TransactionsController.ts
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    state.loading = true
+
+    try {
+      await BlockchainApiController.fetchTransactions({
+        account: accountAddress,
+        projectId,
+        cursor: state.next
+      }).then(response => {
```
0xAsimetriq, line 53: Should this be not `.then()` but rather `const response = await BlockchainApiController...` ?

0xAsimetriq (changes requested):
Few small suggestions.

I just realized I missed another major bit during last review:

1. Connect with metamask extension (or any wallet that allows you to switch account)
2. Load up transactions view
3. Switch (from wallet) to a different address

Same transactions remain in the list, when they should be re-fetched for a new account. You'll find few examples of how to do this in account view where we subscribe to address changes, once this happens you need to reset your tx state and fetch new ones.

packages/common/src/utils/TypeUtil.ts
```diff
@@ -0,0 +1,63 @@
...
+}
+
+export interface TransactionContent {
+  preview?: TransactionPreview
+  detail?: TransactionDetail
+}
+
+export interface TransactionPreview {
+  url: string
+  content_type?: null
```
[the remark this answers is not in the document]
enesozturk: Yes, this value comes from the backend (from Zerion) If there is no next page. The `next` will be `null`

packages/core/src/controllers/TransactionsController.ts
enesozturk: That works as well. The `then` will be working only if the fetch doesn't throw any errors. So the we could reach the same functionality with `const response = await BlockchainApiController ` approach as well. It's just different style of handling async functions.

Seems like we don't use `then` approach so I'll update it now

packages/common/src/utils/TypeUtil.ts
```diff
@@ -0,0 +1,68 @@
+export type TransactionStatus = 'confirmed' | 'failed' | 'pending'
+
+export type TransactionDirection = 'in' | 'out' | 'self'
+
+export type TransactionImage = {
+  type: 'FUNGIBLE' | 'NFT' | undefined
```
0xAsimetriq, line 6:
```suggestion
  type?: 'FUNGIBLE' | 'NFT'
```

packages/core/src/controllers/TransactionsController.ts
```diff
@@ -0,0 +1,107 @@
...
+// -- Controller ---------------------------------------- //
+export const TransactionsController = {
+  state,
+
+  subscribe(callback: (newState: TransactionsControllerState) => void) {
+    return sub(state, () => callback(state))
+  },
+
+  async fetchTransactions() {
+    const projectId = OptionsController.state.projectId
```
0xAsimetriq, line 39:
```suggestion
    const { projectId } = OptionsController.state
```

packages/scaffold/src/views/w3m-transactions-view/index.ts
```diff
@@ -0,0 +1,219 @@
...
+
+// -- Helpers --------------------------------------------- //
+const PAGINATOR_ID = 'last-transaction'
+const LOADING_ITEM_COUNT = 7
+
+@customElement('w3m-transactions-view')
+export class W3mTransactionsView extends LitElement {
+  public static override styles = styles
+
+  // -- Members -------------------------------- //
```
0xAsimetriq, line 17: nit: align comment length with others. Perhaps make a new danger js rule to check this like we do with other comments.

packages/scaffold/src/views/w3m-transactions-view/index.ts
```diff
@@ -0,0 +1,219 @@
...
+
+  @state() private empty = TransactionsController.state.empty
+
+  @state() private next = TransactionsController.state.next
+
+  // -- Lifecycle ----------------------------------------- //
+  public constructor() {
+    super()
+    this.unsubscribe.push(
+      ...[
```
0xAsimetriq, line 37: As there is one controller to unsubscribe from I think you can just push it in directly without `...[`

packages/scaffold/src/views/w3m-transactions-view/index.ts
```diff
@@ -0,0 +1,219 @@
...
+            to grow your wallet!</wui-text
+          >
+        </wui-flex>
+      </wui-flex>
+    `
+  }
+
+  private templateLoading() {
+    return Array(LOADING_ITEM_COUNT)
+      .fill(null)
```
0xAsimetriq, line 160: Maybe you can fill array with placeholder components directly without map?

packages/scaffold/src/views/w3m-transactions-view/index.ts
```diff
@@ -0,0 +1,219 @@
...
+
+  private async fetchTransactions() {
+    await TransactionsController.fetchTransactions()
+  }
+
+  private getTransactionListItemProps(transaction: Transaction) {
+    const date = DateUtil.getRelativeDateFromNow(transaction?.metadata?.minedAt)
+    const descriptions = TransactionUtil.getTransactionDescriptions(transaction)
+
+    const transfer = transaction?.transfers?.[0]
```
0xAsimetriq, line 191:
```ts
const [transfer, secondTransfer] = transaction?.transfers ?? [null, null]
```

packages/core/src/controllers/TransactionsController.ts
```diff
@@ -34,9 +34,9 @@ export const TransactionsController = {
     return sub(state, () => callback(state))
   },
 
-  async fetchTransactions() {
+  async fetchTransactions(address?: string) {
```
enesozturk, line 37: @0xAsimetriq 

What would be the best approach connecting to Account state inside the `TransactionsController`? Or do you think we should do it?

Instaed passing the new address on the time we changed the wallet, I'd love to watch the account state and have the updated state here.

packages/scaffold/src/views/w3m-transactions-view/index.ts
enesozturk: I think now we have two)

packages/core/src/controllers/TransactionsController.ts
0xAsimetriq: You shouldn't connect them, instead subscribe to address change in transaction list view. When address changes re-triger tx fetching logic from within that page by calling appropriate method(s) of `TransactionsController`

packages/common/src/utils/TypeUtil.ts
enesozturk: I think making a prop optional and making it undefined'able is not the same. `?` will reach the same functionality in tye TS worlds. But the meaning will be different

packages/scaffold/src/views/w3m-transactions-view/index.ts
enesozturk: We don't need second transfer here now, so skipping this

0xAsimetriq (dismissed):

packages/ui/src/assets/svg/swapHorizontalBold.ts
svenvoskamp, line 1: Can you replace `svg` for `html` and then save file. This will format the `svg` correctly. After it has been formatted, you can remove `html` and add `svg` again.

svenvoskamp (approved):

packages/scaffold/src/views/w3m-transactions-view/index.ts
enesozturk: Removed

packages/ui/src/composites/wui-icon-box/index.ts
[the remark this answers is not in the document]
enesozturk: I agree, I've noted this as well

0xAsimetriq (approved):

svenvoskamp (approved):

PR #1444, opened 2023-11-03
Feat/ethers disconnect

svenvoskamp:
* Add  function `disconnect` to exports

Closes https://github.com/WalletConnect/web3modal/issues/1439


packages/ethers5/exports/react.tsx
```diff
@@ -46,11 +48,26 @@ export function useWeb3ModalAccount() {
...
   const chainId = state.chainId
+  const provider = state.provider
+  const providerType = state.providerType
+
+  async function disconnect() {
+    localStorage.removeItem(WALLET_ID)
+    ProviderController.reset()
+
+    if (providerType === 'injected' || providerType === 'eip6963') {
+      provider?.emit('disconnect')
```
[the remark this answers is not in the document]
svenvoskamp: This triggers `disconnectHandler` inside the client, which clears `localStorage`

packages/ethers5/src/client.ts
```diff
@@ -313,6 +313,20 @@ export class Web3Modal extends Web3ModalScaffold {
     return ProviderController.subscribe(callback)
   }
 
+  public async disconnect() {
+    const provider = ProviderController.state.provider
```
0xAsimetriq, line 317:
```suggestion
    const {provider, providerType} = ProviderController.state
```

packages/ethers5/exports/react.tsx
```diff
@@ -46,11 +48,26 @@ export function useWeb3ModalAccount() {
   const address = state.address
   const isConnected = state.isConnected
   const chainId = state.chainId
+  const provider = state.provider
+  const providerType = state.providerType
+
+  async function disconnect() {
```
0xAsimetriq, line 54: Let's keep consistent and do `useDisconnect` that returns disconnect function. Also, it seems like same logic as in `client` is replicated there. Can we not just use `ConnectionController.disconnect` at this stage (same as all other hooks use controllers for things like theme etc?)

0xAsimetriq (changes requested):

0xAsimetriq (approved):
Much cleaner 👌 ty for changes

glitch-txs (approved):

GorgeousPuree:
Please update documentation with this new hook.
https://docs.walletconnect.com/web3modal/react/about

PR #1483, opened 2023-11-17
refactor: replace outline stylings with box shadow

enesozturk:
In old versions of Safari (Mac and iOS) and Chrome (in old iOS versions) don't have the ability to shape the outline depending on the component. Instead, another approach to giving items outline is using the box-shadow property. 

# Changes

- fix: replace all outline stylings with box-shadow

# Associated Issues

- [Slack conversation](https://walletconnect.slack.com/archives/C03RVH94K5K/p1700216019771239)
- [Slack conversation](https://walletconnect.slack.com/archives/C03RVH94K5K/p1700164765905539)

### DIFF
--- a/packages/ui/src/components/wui-shimmer/styles.ts
+++ b/packages/ui/src/components/wui-shimmer/styles.ts
@@ -3,8 +3,7 @@ import { css } from 'lit'
 export default css`
   :host {
     display: block;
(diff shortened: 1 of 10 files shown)


svenvoskamp (approved):
Thanks for jumping on this one! ✅

PR #1484, opened 2023-11-18
add ethers v6 support

glitch-txs:
# Breaking Changes

ethers v6 package

ethers provider [doesn't allow events from wallet providers](https://github.com/ethers-io/ethers.js/issues/4334), so I'm using the raw providers and then exposing them to the users, they can wrap themselves the raw provider with ethers.

# Changes

- feat: ethers v6 support
image

# Some considerations

There are two things I think we should change in both libraries:

- Disconnect event is used to disconnect the user from the session, this overlaps with the actual use of disconnect following EIP-1193 where it should only be called when the provider is disconnected from the blockchain

- `name` property from EIP-6963 is being used to filter out duplicated providers, we should use uuid instead as some wallets could impersonate each other


packages/ethers/exports/react.tsx
```diff
@@ -0,0 +1,74 @@
...
+      _sdkVersion: `react-ethers-${ConstantsUtil.VERSION}`
+    })
+  }
+  getWeb3Modal(modal)
+
+  return modal
+}
+
+// -- Hooks -------------------------------------------------------------------
+export function useWeb3ModalProvider() {
```
svenvoskamp, line 30: Shouldn't we add `signer` to export here, same on how we do it in `ethers-5`? 

This way we keep consistency between ethers packages.

packages/ethers/exports/vue.ts
```diff
@@ -0,0 +1,93 @@
...
+      _sdkVersion: `vue-ethers-${ConstantsUtil.VERSION}`
+    })
+    getWeb3Modal(modal)
+  }
+
+  return modal
+}
+
+// -- Composites --------------------------------------------------------------
+export function useWeb3ModalProvider() {
```
svenvoskamp, line 26: Same as comment above ^

packages/ethers/src/utils/constants.ts
svenvoskamp, line 1: We use same `constants` and `helpers` in `ethers` and `ethers-5`. Maybe we can rename `@web3modal/utils` package to `@web3modal/scaffold-utils` and add an ethers specific folder to create shared `constants`, `helpers`, etc

packages/ethers/src/utils/helpers.ts
svenvoskamp, line 1: Same comments as above ^

svenvoskamp (changes requested):
Left a few comments and suggestions about duplicate code. Other than that looks good 👍

Can you also fix CI status to green?

apps/laboratory/package.json
```diff
@@ -13,11 +13,14 @@
...
-    "@web3modal/ethers5": "3.3.2",
+    "@web3modal/ethers": "3.3.2",
     "@web3modal/wagmi": "3.3.2",
     "framer-motion": "10.16.4",
     "next": "13.5.6",
     "react-icons": "4.11.0",
     "valtio": "1.11.2"
+  },
+  "devDependencies": {
+    "ethers": "6.8.1"
```
ignaciosantise, line 24: i think all dev dependencies are added in the project root

packages/ethers/src/utils/constants.ts
glitch-txs: I'd suggest creating a global library like @web3modal/connectors or something like that, we can add the whole logic there, then for packages like @web3modal/ethers we can just have the exports there, and pass down to the Web3Modal class the functions needed for fetching data like Balance/ENS. We could eventually easily add web3js support as well.

thoguhts?

glitch-txs: Helpers use particular types from ethers v5 library so I wasn't able to move that one

PR #1491, opened 2023-11-21
feat(swap): 1inch swaps

Cali93:
# Breaking Changes

N/A

# Changes

- feat: Add swaps

# Associated Issues

closes #...


packages/wagmi/src/client.ts
```diff
@@ -168,7 +172,42 @@ export class Web3Modal extends Web3ModalScaffold {
         }
       },
 
-      signMessage: async message => signMessage(this.wagmiConfig, { message })
+      signMessage: async message => signMessage(this.wagmiConfig, { message }),
+
+      sendTransaction: async (data: SendTransactionArgs) => {
+        // eslint-disable-next-line no-useless-catch
+        try {
```
peculiarity, line 179: What do you think about this one?

```javascript
try {
   const txPayload = {
      account: data.address,
      to: data.to,
      value: data.value,
      gas: data.gas,
      gasPrice: data.gasPrice,
      data: data.data,
      type: 'legacy'
   }

    await prepareTransactionRequest(this.wagmiConfig, txPayload)
    await tx = await wagmiSendTransaction(this.wagmiConfig, txPayload)
    await waitForTransactionReceipt(this.wagmiConfig, { hash: tx, timeout: 25000 })

    return tx
  } catch (error) {
      throw error
  }
}
```

packages/core/src/controllers/ConvertApiController.ts
```diff
@@ -0,0 +1,350 @@
...
+      throw new Error('No address found to swap the tokens from.')
+    }
+
+    return {
+      fromAddress: address as `0x${string}`,
+      slippage: DEFAULT_SLIPPAGE_TOLERANCE
+    }
+  },
+
+  // /gas-price/v1.5/${chainId}
```
enesozturk, line 150: Remove comments

packages/core/src/controllers/ConvertController.ts
```diff
@@ -0,0 +1,664 @@
+import { subscribeKey as subKey } from 'valtio/utils'
```
enesozturk, line 1: Too big file, maybe we could separate some functions to different controllers/utils?

packages/core/src/controllers/ConvertController.ts
```diff
@@ -0,0 +1,664 @@
...
+
+    if (fromAddress && toTokenAddress && toTokenDecimals && !state.loading) {
+      const transaction = await this.getTransaction()
+      this.setTransactionDetails(transaction)
+    }
+  },
+
+  calculateGasPriceInEther(gas: bigint, gasPrice: bigint) {
+    const totalGasCostInWei = gasPrice * gas
+    const totalGasCostInEther = Number(totalGasCostInWei) / 1e18
```
enesozturk, line 375: Define the values like `1e18` into different places, also, we should be sure that every calculations have been considered as big number and calculations are valid.

packages/core/src/controllers/ConvertController.ts
```diff
@@ -0,0 +1,664 @@
...
+    const transaction = await this.getTransaction()
+    this.setTransactionDetails(transaction)
+    state.loading = false
+  },
+
+  async getTransaction() {
+    const { fromAddress, sourceTokenAddress, sourceTokenAmount, sourceTokenDecimals } =
+      this.getParams()
+
+    if (
```
enesozturk, line 445: We should handle this state in UI. User shouldn't be able to run this function

packages/core/src/controllers/ConvertController.ts
```diff
@@ -0,0 +1,664 @@
...
+
+    // return toTokenConvertedAmount with 18 decimals
+    return toTokenConvertedAmount.multipliedBy(multiplyer).toString()
+  },
+
+  async createTokenAllowance() {
+    const { fromAddress, sourceTokenAddress } = this.getParams()
+
+    if (!sourceTokenAddress) {
+      throw new Error('>>> createTokenAllowance - No source token address found.')
```
enesozturk, line 496: Remove throw, instead this should be handled in UI, users shouldn't be able to call this function.

packages/core/src/controllers/ConvertController.ts
```diff
@@ -0,0 +1,664 @@
...
+  async createConvert() {
+    const {
+      fromAddress,
+      sourceTokenAddress,
+      sourceTokenDecimals,
+      sourceTokenAmount,
+      toTokenAddress
+    } = this.getParams()
+
+    if (!sourceTokenAmount || !sourceTokenAddress || !toTokenAddress || !sourceTokenDecimals) {
```
enesozturk, line 558: Same, if somehow user should go into this state, they might not get a feedback

packages/core/src/controllers/ConvertController.ts
```diff
@@ -0,0 +1,664 @@
...
+
+      setTimeout(() => {
+        this.resetValues()
+        this.getMyTokensWithBalance()
+      }, 1000)
+
+      return transactionHash
+    } catch (err) {
+      const error = err as TransactionError
+      console.log('>>> sendTransactionForConvert - error', error)
```
enesozturk, line 626: Remove the log

packages/core/src/utils/ConstantsUtil.ts
```diff
@@ -61,7 +61,99 @@ export const ConstantsUtil = {
     Base: 'base'
   },
 
-  WC_COINBASE_ONRAMP_APP_ID: 'bf18c88d-495a-463b-b249-0b9d3656cf5e'
+  WC_COINBASE_ONRAMP_APP_ID: 'bf18c88d-495a-463b-b249-0b9d3656cf5e',
+
+  SUGGESTED_TOKENS: [
```
enesozturk, line 66: These are predefined and temporary values, we should select which tokens to show first

packages/core/src/utils/ConstantsUtil.ts
```diff
@@ -61,7 +61,99 @@ export const ConstantsUtil = {
...
+    'USDT',
+    'ARB',
+    'BAL',
+    'BICO',
+    'CRV',
+    'ENS',
+    'MATIC',
+    'OP',
+    // Some Polygon tokens
+    'DAI',
```
enesozturk, line 125: For testing, I've whitelisted some tokens to list in the token select page, we shouldn't trim the token array. Since it's too big list, the performance of the token select dropping. Waiting the Lit team to merge this: https://github.com/lit/lit/pull/4564. Then we could render large lists, then we don't need to filter the tokens

packages/core/src/utils/CoreHelperUtil.ts
```diff
@@ -54,6 +54,16 @@ export const CoreHelperUtil = {
     return caipAddress.split(':')[2]
   },
 
+  getEvmChainId(caipNetworkId?: `${string}:${string}`) {
```
enesozturk, line 57: This function might already be implemented

packages/core/src/utils/TypeUtil.ts
```diff
@@ -339,6 +339,18 @@ export type Event =
         network: string
       }
     }
+  | {
+      type: 'track'
+      event: 'CLICK_CONVERT'
+    }
+  | {
+      type: 'track'
+      event: 'CLICK_SELECT_TOKEN_TO_SWAP'
```
enesozturk, line 348: Remove "Swap" wordings and use events for certain actions

enesozturk: Ask Sam to which events to trigger for convert

packages/ethers/src/client.ts
```diff
@@ -274,6 +279,67 @@ export class Web3Modal extends Web3ModalScaffold {
...
+
+      formatUnits: (value: bigint, decimals: number) => formatUnits(value, decimals),
+
+      async getEstimatedGas(data) {
+        const chainId = EthersStoreUtil.state.chainId
+        const provider = EthersStoreUtil.state.provider
+        const address = EthersStoreUtil.state.address
+
+        if (!provider) {
+          throw new Error('connectionControllerClient:sendTransaction - provider is undefined')
```
enesozturk, line 294: Check how we handle these kind of exceptions in the clients. Might needed to be updated

packages/scaffold/src/views/w3m-approve-transaction-view/index.ts
```diff
@@ -2,7 +2,17 @@ import { customElement } from '@web3modal/ui'
...
-import { ModalController, ConnectorController, ThemeController } from '@web3modal/core'
+import {
+  ModalController,
+  ConnectorController,
+  ThemeController,
+  RouterController
+} from '@web3modal/core'
+
+// -- Variables ------------------------------------------- //
+const PAGE_HEIGHT = 400
```
enesozturk, line 13: Previously we were updating the iframe styles depending on the changes to it's parent sizes. While this method seem okay, at some situations, it brings lack of ability or performance issues on animations. Instead, using the certain values of the parent would give us better experience. 

So right now, we are calculating the left and top values with these values.

packages/scaffold/src/views/w3m-approve-transaction-view/index.ts
```diff
@@ -20,34 +30,34 @@ export class W3mApproveTransactionView extends LitElement {
...
-      ModalController.subscribeKey('open', val => {
-        if (!val) {
-          this.onHideIframe()
-        }
-      })
+      ...[
+        ModalController.subscribeKey('open', isOpen => {
+          if (!isOpen) {
+            this.onHideIframe()
+            RouterController.popTransactionSuccessAction()
```
enesozturk, line 39: Double check if there is a state that approve transaction page is triggered but the transaction success stack is not updated.

packages/scaffold/src/views/w3m-convert-select-network-view/index.ts
```diff
@@ -0,0 +1,88 @@
+import { customElement } from '@web3modal/ui'
```
enesozturk, line 1: This a component from the hackweek and we currently don't need/use this. We can remove

packages/core/src/controllers/ConvertController.ts
enesozturk: I mean if they somehow press button but some of these values are empty, the function returns null which mean do nothing, and if the user don't get any feedback in UI, this wouldn't be good.

packages/core/src/controllers/ConvertController.ts
```diff
@@ -0,0 +1,664 @@
...
+
+  calculateGasPriceInUSD(gas: bigint, gasPrice: bigint) {
+    try {
+      const totalGasCostInEther = this.calculateGasPriceInEther(gas, gasPrice)
+      const networkPriceNumber = NumberUtil.bigNumber(state.networkPrice)
+      const totalCostInUSD = networkPriceNumber.multipliedBy(totalGasCostInEther)
+
+      return totalCostInUSD.toNumber()
+    } catch (error) {
+      return 0
```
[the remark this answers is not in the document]
enesozturk: We don't the try catch here I think

packages/core/src/controllers/ConvertApiController.ts
```diff
@@ -0,0 +1,350 @@
...
+        gas: OneInchAPIEndpoints.gasPrice(chainId),
+        gasPrice: OneInchAPIEndpoints.gasPrice(chainId),
+        swap: OneInchAPIEndpoints.swap(chainId),
+        tokens: OneInchAPIEndpoints.tokens(chainId),
+        tokensCustom: OneInchAPIEndpoints.tokensCustom(chainId),
+        tokenPrices: OneInchAPIEndpoints.tokensPrices(chainId),
+        search: OneInchAPIEndpoints.search(chainId),
+        balance: OneInchAPIEndpoints.balance(chainId, address),
+        quote: OneInchAPIEndpoints.quote(chainId)
+      }
```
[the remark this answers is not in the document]
enesozturk: Agreed. Updated

packages/common/src/utils/NumberUtil.ts
```diff
@@ -0,0 +1,49 @@
...
+    if (a === undefined || b === undefined) {
+      // eslint-disable-next-line new-cap
+      return BigNumber(0)
+    }
+
+    const aBigNumber = new BigNumber(a)
+    const bBigNumber = new BigNumber(b)
+
+    return aBigNumber.multipliedBy(bBigNumber)
+  }
```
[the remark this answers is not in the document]
enesozturk: While calculating numbers with too small values, we might need to show the user like `0.000..0634`, etc. This is not quite possible to handle with JS's number utilities. bignumber.js allows us to make calculations with both strings and numbers and return the valid equivalents.

It might have a downside of having another dependency and bundle size, but at the same time, IMO trying to handle numbers manually might tent to have wrong outputs.

One thing here I want to double check that, in none of the UIs, we don't show really long tails of the number. This means that we 'might' not need this actually. I'll double check this

PR #1499, opened 2023-11-22
[fix] Fix all wallets displayed count

saintmaxi:
# Breaking Changes

N/A

# Changes

- fix displayed count for All Wallets when less than 10 options

# Associated Issues

closes https://github.com/WalletConnect/web3modal/issues/1474


glitch-txs:
LGTM @saintmaxi , could you run prettier?

saintmaxi:
> LGTM @saintmaxi , could you run prettier?

ran it, looks like nothing changed

glitch-txs:
open the terminal in the scaffold folder and try running `npx prettier --write .`

saintmaxi:
ah thanks, done

glitch-txs (approved):

packages/scaffold/src/views/w3m-connect-view/index.ts
```diff
@@ -200,15 +200,17 @@ export class W3mConnectView extends LitElement {
       return null
     }
 
-    const roundedCount = Math.floor(ApiController.state.count / 10) * 10
+    const rawCount = ApiController.state.count
+    const roundedCount = rawCount < 10 ? rawCount : Math.floor(rawCount / 10) * 10
```
enesozturk, line 204: @saintmaxi @glitch-txs This logic will only put a `+` prefix for the numbers between 10 - 20. But if it's higher than 20, it will just render the number. Should this be discussed design-wise again? 

Not sure what is expected logic here is.

glitch-txs: Mmm let's say

const rawCount = 31

then

```ts
const roundedCount = Math.floor(31/10) * 10
// 30
```
This means  roundedCount < rawCount = true

Therefore it would get the + sign as 30+

Am I missing anything? 🤔

enesozturk (approved):

enesozturk:
@saintmaxi 

Sorry, in another PR I needed to get these changes and make a few updates: https://github.com/WalletConnect/web3modal/blob/V3/packages/scaffold/src/views/w3m-connect-view/index.ts#L203-L207

I should have merged this PR first but I totally forgot. Let me add your name to the release notes once we publish a new version. Thank you for your contribution, we appreciate your support any time.

cc @glitch-txs

saintmaxi:
> cc

Sounds good to me! Thank you for letting me take a stab at it.

PR #1517, opened 2023-11-30
fix(siwe): WalletConnect Provider address casing and remove NEXTAUTH_URL

Cali93:
# Breaking Changes

N/A

# Changes

- feat:
- fix: WalletConnect Provider address casing
- chore:

# Associated Issues

closes #...


apps/laboratory/.env.example
```diff
@@ -1,2 +1,3 @@
 # Obtain a project ID from https://cloud.walletconnect.com
 NEXT_PUBLIC_PROJECT_ID=""
+NEXTAUTH_SECRET=mysecret123
```
0xAsimetriq, line 3: New line

packages/ethers5/src/client.ts
```diff
@@ -313,7 +313,7 @@ export class Web3Modal extends Web3ModalScaffold {
   }
 
   public getAddress() {
-    return ethers.utils.getAddress(EthersStoreUtil.state.address as string)
+    return utils.getAddress(EthersStoreUtil.state.address as string)
```
0xAsimetriq, line 316: if `as string` is needed due to undefined being a possibility, check for that case explicitly. Otherwise ok 👍

packages/ethers5/src/client.ts
```diff
@@ -467,7 +467,9 @@ export class Web3Modal extends Web3ModalScaffold {
       EthersStoreUtil.setProviderType('walletConnect')
       EthersStoreUtil.setProvider(WalletConnectProvider as unknown as Provider)
       EthersStoreUtil.setIsConnected(true)
-      EthersStoreUtil.setAddress(WalletConnectProvider.accounts[0] as Address)
+      const address = WalletConnectProvider.accounts?.[0]
```
0xAsimetriq, line 470: Looks like we are doing identical things in ethers and ethers5 packages, can all of this logic be abstracted under ethers scaffold package and not repeated? Also cc @svenvoskamp here for further feedback on ethers.

Cali93: I added some abstraction on each client but we can't share that logic because each of our ethers package need a different utils function from the OG ethers package.

PR #1522, opened 2023-11-30
Feat/email login ethers

svenvoskamp:
# Breaking Changes

N/A

# Changes

- feat:
- fix:
- chore:

# Associated Issues

closes #...


packages/ethers/src/client.ts
```diff
@@ -34,6 +34,8 @@ import {
 } from '@web3modal/scaffold-utils/ethers'
 import type { EthereumProviderOptions } from '@walletconnect/ethereum-provider'
 import type { Eip1193Provider } from 'ethers'
+import { W3mFrameProvider } from '@web3modal/smart-account'
+import type { CombinedProvider } from 'packages/scaffold-utils/dist/types/src/ethers'
```
0xAsimetriq, line 38: `CombinedProvider` should be available on `@web3modal/scafold-utils` instead of dirrect import

packages/ethers5/package.json
```diff
@@ -60,6 +60,7 @@
     "ethers": ">=5.0.0 <6.0.0",
     "react": ">=17",
     "react-dom": ">=17",
+    "viem": ">=1",
```
0xAsimetriq, line 63: Why is viem needed in ethers package?

PR #1560, opened 2023-12-09
fix: remove EIP6963 connector from localStore on account change

glitch-txs:
# Breaking Changes

Override onAccountChange function & added `this.storage?.removeItem(connectedRdnsKey)`

# Changes

- feat:
- fix: EIP6963 wallets were triggered on reload when were manually disconnected, this would trigger the accountsChanged event and clear the storage without calling the overriden `disconnect` function
- chore:

# Associated Issues

closes https://github.com/WalletConnect/web3modal/issues/1488

### DIFF
--- a/packages/wagmi/src/connectors/EIP6963Connector.ts
+++ b/packages/wagmi/src/connectors/EIP6963Connector.ts
@@ -55,10 +55,11 @@ export class EIP6963Connector extends InjectedConnector {
     if (accounts.length === 0) {
       this.storage?.removeItem(connectedRdnsKey)
       this.emit('disconnect')
-    } else
+    } else if (accounts[0]) {
       this.emit('change', {
-        account: getAddress(accounts[0] as string)
+        account: getAddress(accounts[0])
       })
+    }
   }
 
   public override async disconnect() {
--- a/packages/wagmi/src/connectors/EIP6963Connector.ts
+++ b/packages/wagmi/src/connectors/EIP6963Connector.ts
@@ -1,4 +1,4 @@
-import { type Chain, type WindowProvider } from '@wagmi/core'
+import type { Chain, WindowProvider } from '@wagmi/core'
 import { InjectedConnector } from '@wagmi/core/connectors/injected'
 import { getAddress } from 'viem'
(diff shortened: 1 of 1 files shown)


packages/wagmi/src/connectors/EIP6963Connector.ts
```diff
@@ -50,6 +51,17 @@ export class EIP6963Connector extends InjectedConnector {
     return data
   }
 
+  protected override onAccountsChanged = (accounts: string[]) => {
```
enesozturk, line 54: Are we sure that we definitely have an array of strings as a parameter? I couldn't find a source where this event was triggered. If it's in case that it could be `undefined` or something else, the `accounts.length` or `accounts[0]` property access' might fail.

Could you share with me some links where I can find more context about the EIP-6963? I have the link below. More would be appreciated.

- https://eips.ethereum.org/EIPS/eip-6963

packages/wagmi/src/connectors/EIP6963Connector.ts
```diff
@@ -1,5 +1,6 @@
-import type { Chain, WindowProvider } from '@wagmi/core'
+import { type Chain, type WindowProvider } from '@wagmi/core'
```
enesozturk, line 1: Seems like `import type { ... }` is our convention across the repo. Any specific reason to use this way?

packages/wagmi/src/connectors/EIP6963Connector.ts
glitch-txs: Hi, this comes from EIP-1193 standard
https://eips.ethereum.org/EIPS/eip-1193

What I basically did here was copy/paste the function from Wagmi's injectedConnector and add the line to remove the custom storage item.

packages/wagmi/src/connectors/EIP6963Connector.ts
glitch-txs: Prettier or something idk, I'll fix it, makes no sense in this format lol

0xAsimetriq (approved):

PR #1561, opened 2023-12-12
fix: network image access in account button and account page

enesozturk:
# Changes

We're accessing the network image SRC in different ways in different components. We can use custom chain images if we use `AssetUtil.getNetworkImage`. So instead of manually getting the network image, using `AssetUtil` will fix the issue linked below.

# Changes

- fix: custom network image rendering

# Test

Add a map to the `createWeb3Modal` function as parameter like the following:

```js
const modal = createWeb3Modal({
  ...
  chainImages: {
    1: 'image_url...'
  }
})
```

# Associated Issues

Closes https://github.com/WalletConnect/web3modal/issues/1406
Closes https://github.com/WalletConnect/web3modal/issues/1405

### DIFF
--- a/packages/ethers5/src/client.ts
+++ b/packages/ethers5/src/client.ts
@@ -8,10 +8,10 @@ import type {
   LibraryOptions,
   NetworkControllerClient,
   PublicStateControllerState,
+  SIWEControllerClient,
   Token
 } from '@web3modal/scaffold'
 import { Web3ModalScaffold } from '@web3modal/scaffold'
-import type { Web3ModalSIWEClient } from '@web3modal/siwe'
 import { ConstantsUtil, PresetsUtil, HelpersUtil } from '@web3modal/scaffold-utils'
 
 import EthereumProvider from '@walletconnect/ethereum-provider'
@@ -35,7 +35,7 @@ import type { EthereumProviderOptions } from '@walletconnect/ethereum-provider'
 // -- Types ---------------------------------------------------------------------
 export interface Web3ModalClientOptions extends Omit<LibraryOptions, 'defaultChain' | 'tokens'> {
   ethersConfig: ProviderType
-  siweConfig?: Web3ModalSIWEClient
+  siweConfig?: SIWEControllerClient
   chains?: Chain[]
   defaultChain?: Chain
   chainImages?: Record<number, string>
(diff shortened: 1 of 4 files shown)


packages/ethers5/src/client.ts
```diff
@@ -35,7 +35,7 @@ import type { EthereumProviderOptions } from '@walletconnect/ethereum-provider'
 // -- Types ---------------------------------------------------------------------
 export interface Web3ModalClientOptions extends Omit<LibraryOptions, 'defaultChain' | 'tokens'> {
   ethersConfig: ProviderType
-  siweConfig?: Web3ModalSIWEClient
+  siweConfig?: SIWEControllerClient
```
Cali93, line 38: Curious why you had to make that change ? `siweConfig` should be of type `Web3ModalSIWEClient`

enesozturk: @Cali93 Seems like both types do not match:

```bash
@web3modal/ethers5:build: src/client.ts(251,7): error TS2322: Type 'Web3ModalSIWEClient | undefined' is not assignable to type 'SIWEControllerClient | undefined'.
@web3modal/ethers5:build:   Property 'signIn' is missing in type 'Web3ModalSIWEClient' but required in type 'SIWEControllerClient'.
```

```ts
export declare class Web3ModalSIWEClient {
    options: SIWEControllerClient['options'];
    methods: SIWEClientMethods;
    constructor(siweConfig: SIWEConfig);
    getNonce(): Promise<string>;
    createMessage(args: SIWECreateMessageArgs): string;
    verifyMessage(args: SIWEVerifyMessageArgs): Promise<true>;
    getSession(): Promise<import("@web3modal/core").SIWESession>;
    signOut(): Promise<boolean>;
}

export interface SIWEControllerClient extends SIWEClientMethods {
    signIn: () => Promise<SIWESession>;
    options: {
        enabled: boolean;
        nonceRefetchIntervalMs: number;
        sessionRefetchIntervalMs: number;
        signOutOnDisconnect: boolean;
        signOutOnAccountChange: boolean;
        signOutOnNetworkChange: boolean;
    };
}
```

I guess you have more context, can you check this when have a time?

enesozturk: Reverted changes back since shouldn't be this PRs responsibility. Also the CI is not failing with builds so might be related to my environment

svenvoskamp (approved):
LGTM, but please figure out the type issue with Cali

PR #1563, opened 2023-12-12
fix: syncing profile info on network switch

enesozturk:
# Changes

When we switch network, we don't sync the profile information. This causes displayin ENS name or avatar in all networks, which is not expected. 

# Steps to reproduce

(Considering you have ENS name set to your account):
- Connect wallet to Ethereum Mainnet
- See ENS name
- Switch to Avalanche

You still see the ENS name. 

# Changes

- fix: sync profile on the network switch and consider not only mainnet

# Associated Issues

Closes https://github.com/WalletConnect/web3modal/issues/1407
Closes https://github.com/WalletConnect/web3modal/issues/1544
Closes https://github.com/WalletConnect/web3modal/issues/1519#issue-2018270402

### DIFF
--- a/packages/core/src/controllers/AccountController.ts
+++ b/packages/core/src/controllers/AccountController.ts
@@ -10,8 +10,8 @@ export interface AccountControllerState {
   address?: string
   balance?: string
   balanceSymbol?: string
-  profileName?: string
-  profileImage?: string
+  profileName?: string | null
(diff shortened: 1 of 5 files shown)


svenvoskamp (dismissed):
LGTM

svenvoskamp (approved):

PR #1568, opened 2023-12-13
fix: all wallets shimmer bug

enesozturk:
# Changes

When we set `includeWalletIds` and `featuredWalledIds` as the same array, in case people want to show only a few wallets in the wallet select page, the all wallets button and the all wallets page have a few issues:

- All wallets button shows a wrong number of wallet options on the right side. This is fixed in https://github.com/WalletConnect/web3modal/pull/1499/files but we should consider featured apps for the whole count too.
- All wallet lists should be considered if there are only featured apps. When we calculate the `count` value in `ApiController`, the `featured` wallets are not counted, but we fetch the featured ones in all wallet lists 

# Steps to reproduce

- In the `createWeb3Modal` function, pass the `includeWalletIds` and pass `featuredWalletIds` as the same array of Wallet id's.

# Test

- [x] Wagmi examples (default, siwe, email)
- [x] Ethers examples (default, siwe, email)

# Changes

- fix: all wallets count calculation
- fix: all wallet list shimmer rendering logic when only have featured items

# Associated Issues

Closes https://github.com/WalletConnect/web3modal/issues/1566

### DIFF
--- a/packages/scaffold/src/views/w3m-connect-view/index.ts
+++ b/packages/scaffold/src/views/w3m-connect-view/index.ts
@@ -200,7 +200,11 @@ export class W3mConnectView extends LitElement {
       return null
     }
 
-    const roundedCount = Math.floor(ApiController.state.count / 10) * 10
+    const count = ApiController.state.count
+    const featuredCount = ApiController.state.featured?.length
+    const rawCount = count + featuredCount
+    const roundedCount = rawCount < 10 ? rawCount : Math.floor(rawCount / 10) * 10
+    const tagLabel = roundedCount < rawCount ? `${roundedCount}+` : `${roundedCount}`
 
     return html`
       <wui-list-wallet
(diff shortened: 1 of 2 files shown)


packages/scaffold/src/views/w3m-connect-view/index.ts
```diff
@@ -200,15 +200,19 @@ export class W3mConnectView extends LitElement {
       return null
     }
 
-    const roundedCount = Math.floor(ApiController.state.count / 10) * 10
+    const count = ApiController.state.count
+    const featuredCount = ApiController.state.featured?.length
+    const rawCount = count + featuredCount
```
tomiir, line 205: could this result in a problem if `featuredCount` is undefined?.
If `ApiController.state.featured` is undefined, this breaks I think 🤔

enesozturk: Not sure how I went with optional property access operator since it's not expected to be undefined

svenvoskamp (approved):
LGTM

tomiir (approved):

PR #1570, opened 2023-12-14
Chore/error hook ethers

svenvoskamp:
# Changes

Added new `useWeb3ModalError` hook/composition for `ethers` and `ethers5` package to get internal errors

# Associated Issues

closes https://github.com/WalletConnect/web3modal/issues/1490


packages/ethers/src/client.ts
```diff
@@ -777,8 +821,12 @@ export class Web3Modal extends Web3ModalScaffold {
             this.setAddressExplorerUrl(undefined)
           }
           if (this.hasSyncedConnectedAccount) {
-            await this.syncProfile(address)
-            await this.syncBalance(address)
+            try {
```
enesozturk, line 824: @svenvoskamp , I'm considering the error handling strategies within our methods. Currently, it seems like `syncBalance` includes internal error handling, but `syncProfile` does not. Given this, I'm questioning whether it's necessary to use additional try/catch blocks when calling these methods.

My suggestion is to concentrate our error handling efforts within the smallest possible functional units. This approach promotes better encapsulation by ensuring that error handling is managed within each method itself. As a result, the calling functions wouldn't need to be concerned with catching errors from syncBalance or syncProfile, since these methods would already be handling their own errors effectively.

enesozturk:
@svenvoskamp Do you think it would be beneficial to implement similar error handlings for `wagmi` as well?

svenvoskamp:
I don't think we need them, because I think wagmi handles this already internally

packages/ethers/src/client.ts
svenvoskamp: I agree, I removed a few of them. Could you check again?

enesozturk (approved):
🚀

packages/ethers/src/client.ts
enesozturk: Awesome, I think better now, thanks!

PR #1575, opened 2023-12-15
Show copy link button when QR code is loaded

svenvoskamp:
# Changes

- fix: Wait till qr code is loaded to show copy link button 

# Associated Issues

closes https://github.com/WalletConnect/web3modal/issues/1443


enesozturk (approved):
LGTM ⚡️

Causes a layout shift but I think it's still good since we have layout animation;

https://github.com/WalletConnect/web3modal/assets/19428358/60b607fb-dd4d-4608-8ab1-d19f0f22dd0b

svenvoskamp:
Yes was aware of that 

> LGTM ⚡️
> 
> Causes a layout shift but I think it's still good since we have layout animation;
> 
>  Screen.Recording.2023-12-15.at.16.00.43.mov

Yes was aware of that, but I think it is not disruptive or anything

PR #1576, opened 2023-12-15
Fix incorrect network order when connected with wc 

svenvoskamp:
# Changes

- fix: Fix incorrect network order when connected with wc 

# Associated Issues

closes https://github.com/WalletConnect/web3modal/issues/1388


enesozturk (approved):
LGTM

packages/scaffold/src/views/w3m-networks-view/index.ts
```diff
@@ -63,9 +63,26 @@ export class W3mNetworksView extends LitElement {
       NetworkController.state
     const approvedIds = approvedCaipNetworkIds
     const requested = requestedCaipNetworks
```
enesozturk, line 65: It would be beneficial to maintain more descriptive variable names;

```suggestion
    const requestedNetworks = requestedCaipNetworks
```

arein (approved):
Does this need to have a unit/functional test?

svenvoskamp:
> Does this need to have a unit/functional test?

It makes sense to do some functional testing here as the order depends on the specific wallet you're connected with.

PR #1577, opened 2023-12-15
3.6.0-alpha.0

0xAsimetriq:
# Breaking Changes

N/A

# Changes

- feat: Email resend functionality
- feat: Email "send action" throttling of 30s
- feat: Add and pass initial chain selection when connecting via email
- chore: Add custom logic / workaround to improve initial loading time when email login was not used beforehand
- chore: Handle ZodError format
- chore: Handle email iframe sdk loading error
- fix: Size issues when overlaying iframe "approve tx" view over modal
- chore: Various small ui updates to align with figma for email views
- feat: Add ui changes when only email connector is enabled
- feat: Add "Upgrade Wallet" view
- chore: Further build improvements, separating build:clean command from build
- chore: Update dependencies

# Associated Issues

closes #...


package.json
```diff
@@ -21,7 +21,7 @@
     "prettier:format": "prettier --write .",
     "clean": "rm -rf `find . -type d -name node_modules -o -name dist -o -name .next -o -name out`; rm package-lock.json",
     "new-version": "lerna version --no-git-tag-version --exact",
-    "pre-publish": "turbo lint typecheck prettier test; npm run new-version; npm install; turbo run build;",
+    "pre-publish": "turbo lint typecheck prettier test; npm run new-version; npm install",
```
tomiir, line 24: did you re-add the build step somewhere else?

packages/scaffold/src/views/w3m-account-view/index.ts
```diff
@@ -88,7 +87,7 @@ export class W3mAccountView extends LitElement {
         <wui-avatar
           alt=${this.address}
           address=${this.address}
-          imageSrc=${ifDefined(this.profileImage)}
+          imageSrc=${ifDefined(this.profileImage === null ? undefined : this.profileImage)}
```
tomiir, line 90: Maybe `this.profileImage ?? undefined` works and is a bit cleaner

tomiir (approved):
LGTM! Some minor comments / questions

packages/core/src/utils/CoreHelperUtil.ts
```diff
@@ -181,5 +181,18 @@ export const CoreHelperUtil = {
 
       return v.toString(16)
     })
+  },
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  parseError(error: any): string {
```
enesozturk, line 187: I guess we're using `any` since we don't know what the error will be here. Would `unknown` work instead of `any`, so we wouldn't have to override ESLint rules?

packages/wagmi/src/client.ts
```diff
@@ -414,6 +414,8 @@ export class Web3Modal extends Web3ModalScaffold {
     if (typeof window !== 'undefined' && connector) {
       super.setLoading(true)
       const provider = await connector.getProvider()
+      const isLoginEmailUsed = provider.getLoginEmailUsed()
+      super.setLoading(isLoginEmailUsed)
```
enesozturk, line 418: Probably it's because I don't see all the flow, what will happen once `isLoginEmailUsed` is true? Where do we set `loading` to false? Can you double-check this, please?

packages/wallet/src/W3mFrameProvider.ts
```diff
@@ -89,8 +90,20 @@ export class W3mFrameProvider {
...
+  }
+
   public async connectEmail(payload: W3mFrameTypes.Requests['AppConnectEmailRequest']) {
     await this.w3mFrame.frameLoadPromise
+    const lastEmailLoginTime = W3mFrameStorage.get(W3mFrameConstants.LAST_EMAIL_LOGIN_TIME)
+    if (lastEmailLoginTime) {
+      const difference = W3mFrameHelpers.getTimeDifferenceMs(Number(lastEmailLoginTime))
+      if (difference < 30_000) {
+        const cooldownSec = Math.ceil((30_000 - difference) / 1000)
+        throw new Error(`Please try again after ${cooldownSec} seconds`)
```
enesozturk, line 104: Wouldn't it be better to show an error to the user instead of throwing an Error here? It'll break the app as I know, right?

enesozturk (approved):
Looks awesome, left a few comments 🚀

svenvoskamp (commented):
LGTM 🚀

PR #1579, opened 2023-12-15
Fix issue on ethers when switching chains

svenvoskamp:
# Changes

- fix: Fix issue on ethers when switching chains

# Associated Issues

closes https://github.com/WalletConnect/web3modal/issues/1526


glitch-txs:
Fixes issue 1, but still experiencing 2. (using Network button to connect doesn't switch to that chain as default)

arein (approved):
does this need a test for regressions?

0xAsimetriq (approved):

PR #1581, opened 2023-12-17
feat(lab): scaffolds a basic UI test

arein:
# Breaking Changes

N/A

# Changes

- feat(lab): scaffolds a basic UI test

Next: https://github.com/WalletConnect/web3modal/issues/1582

# Associated Issues

closes #...

### DIFF
--- a/.github/workflows/ui_tests.yml
+++ b/.github/workflows/ui_tests.yml
@@ -1,0 +1,53 @@
+name: UI Tests
+
+on:
+  push:
+    branches: ['V3']
+    paths-ignore:
+      - '**.md'
+
+  pull_request:
+    branches: ['V3']
+    paths-ignore:
+      - '**.md'
+
+concurrency:
+  # Support push/pr as event types with different behaviors each:
+  # 1. push: queue up builds
+  # 2. pr: only allow one run per PR
+  group: ${{ github.workflow }}-${{ github.event.type }}${{ github.event.pull_request.number }}
+  # If there is already a workflow running for the same pull request, cancel it
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
+jobs:
+  ui_tests:
+    name: 'Playwright Tests'
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - name: checkout
+        uses: actions/checkout@v3
+      - name: setup-node
+        uses: actions/setup-node@v3
+        with:
+          node-version: 18.x
+          cache: 'npm'
+          cache-dependency-path: 'package-lock.json'
+      - name: install
+        run: npm ci
+      - name: build
+        run: npm run build
+      - name: Install Playwright Browsers
+        working-directory: ./apps/laboratory/
+        run: npm run playwright:install
+      - name: Run Playwright tests
+        env:
+          NEXT_PUBLIC_PROJECT_ID: ${{ secrets.NEXT_PUBLIC_PROJECT_ID }}
+        working-directory: ./apps/laboratory/
+        run: npm run playwright:test
+      - uses: actions/upload-artifact@v3
+        if: always()
+        with:
+          name: playwright-report
+          path: ./apps/laboratory/playwright-report/
+          retention-days: 7
--- a/apps/laboratory/package.json
+++ b/apps/laboratory/package.json
@@ -5,7 +5,12 @@
   "scripts": {
     "dev:laboratory": "next dev",
     "build:laboratory": "next build",
-    "lint": "eslint . --ext .js,.jsx,.ts,.tsx"
+    "build:all": "turbo run build",
+    "lint": "eslint . --ext .js,.jsx,.ts,.tsx",
+    "playwright:start": "npm run build:all; npm run dev:laboratory",
+    "playwright:install": "npx playwright install --with-deps",
+    "playwright:test": "npx playwright test",
+    "playwright:debug": "npx playwright test --debug"
   },
   "dependencies": {
     "@chakra-ui/react": "2.8.2",
@@ -24,6 +29,8 @@
     "valtio": "1.11.2"
   },
   "devDependencies": {
+    "@playwright/test": "1.40.1",
+    "dotenv": "16.3.1",
     "ethers": "6.9.0"
   }
 }
--- a/apps/laboratory/tests/shared/constants/index.ts
+++ b/apps/laboratory/tests/shared/constants/index.ts
@@ -1,0 +1,2 @@
+// Allow localhost
+export const LOCAL_LABS_URL = 'http://localhost:3000/library/wagmi/'
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ lerna-debug.log
 .env.local
 .env
 .turbo
+playwright-report/
--- a/apps/laboratory/playwright.config.ts
+++ b/apps/laboratory/playwright.config.ts
@@ -1,0 +1,73 @@
+import { defineConfig, devices } from '@playwright/test'
+import { LOCAL_LABS_URL } from './tests/shared/constants'
+
+import { config } from 'dotenv'
+config({ path: './.env.local' })
+
+export default defineConfig({
+  testDir: './tests',
+
+  fullyParallel: true,
+  retries: process.env['CI'] ? 2 : 0,
+  workers: process.env['CI'] ? 1 : undefined,
+  reporter: [['list'], ['html']],
+
+  expect: {
+    timeout: (process.env['CI'] ? 60 : 5) * 1000
+  },
+  timeout: 60 * 1000,
+
+  use: {
+    /* Base URL to use in actions like `await page.goto('/')`. */
+    baseURL: LOCAL_LABS_URL,
+
+    /* Collect trace when retrying the failed test. See https://playwright.dev/docs/trace-viewer */
+    trace: 'on-first-retry',
+
+    video: process.env['CI'] ? 'off' : 'on-first-retry'
+  },
+
+  /* Configure projects for major browsers */
+  projects: [
+    {
+      name: 'chromium',
+      use: { ...devices['Desktop Chrome'] }
+    },
+
+    {
+      name: 'firefox',
+      use: { ...devices['Desktop Firefox'] }
+    }
+
+    /* Test against mobile viewports. */
+    /*
+     * {
+     *   name: 'Mobile Chrome',
+     *   use: { ...devices['Pixel 5'] },
+     * },
+     * {
+     *   name: 'Mobile Safari',
+     *   use: { ...devices['iPhone 12'] },
+     * },
+     */
+
+    /* Test against branded browsers. */
+    /*
+     * {
+     *   name: 'Microsoft Edge',
+     *   use: { ...devices['Desktop Edge'], channel: 'msedge' },
+     * },
+     * {
+     *   name: 'Google Chrome',
+     *   use: { ..devices['Desktop Chrome'], channel: 'chrome' },
+     * },
+     */
+  ],
+
+  /* Run your local dev server before starting the tests */
+  webServer: {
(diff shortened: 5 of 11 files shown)


apps/laboratory/package.json
```diff
@@ -5,7 +5,11 @@
   "scripts": {
     "dev:laboratory": "next dev",
     "build:laboratory": "next build",
-    "lint": "eslint . --ext .js,.jsx,.ts,.tsx"
+    "lint": "eslint . --ext .js,.jsx,.ts,.tsx",
+    "playwright:start": "next dev",
```
0xAsimetriq, line 9: If tests are for lab, this should be `npm run laboratory` to start in in dev mode. Optionally, you could combine with a fresh build as well, so `turbo run build; npm run laboratory`

apps/laboratory/tests/shared/constants/index.ts
```diff
@@ -0,0 +1 @@
+export const LOCAL_LABS_URL = 'https://lab.web3modal.com/library/wagmi/'
```
0xAsimetriq, line 1: Should this be localhot:3000 or should variable be re-named?

0xAsimetriq (commented):
Left few minor suggestions. I think at the moment this is testing against lab.web3modal.com and not branch code?

chris13524 (changes requested):
Use [`webServer.command`](https://playwright.dev/docs/test-configuration) to spin up a dev version of the tests when running. Similar to how it was done for [V2](https://github.com/WalletConnect/web3modal/blob/2a0e946f9c848c25d6c9cf3c53ab4a04ed372345/laboratory/playwright.config.ts#L69)

arein:
> Use [`webServer.command`](https://playwright.dev/docs/test-configuration) to spin up a dev version of the tests when running. Similar to how it was done for [V2](https://github.com/WalletConnect/web3modal/blob/2a0e946f9c848c25d6c9cf3c53ab4a04ed372345/laboratory/playwright.config.ts#L69)

thank you was wondering how this works and missed this when I copied from V2

.github/workflows/ui_tests.yml
```diff
@@ -0,0 +1,57 @@
...
+      - name: build
+        run: npm run build
+      - name: Install Playwright Browsers
+        working-directory: ./apps/laboratory/
+        run: npm run playwright:install
+      - name: Run Playwright tests
+        env:
+          NEXT_PUBLIC_PROJECT_ID: ${{ secrets.NEXT_PUBLIC_PROJECT_ID }}
+        working-directory: ./apps/laboratory/
+        run: npm run playwright:test:${{ matrix.project }}
```
chris13524, line 51: We should get rid of this matrix and use Playwright's built-in parallelization which will test against all browsers at once. This will save build minutes building the same thing twice.

```suggestion
        run: npm run playwright:test
```

chris13524 (changes requested):

apps/laboratory/package.json
chris13524: What's the difference between `npm run build` and `turbo run build`?

apps/laboratory/package.json
```diff
@@ -5,7 +5,11 @@
   "scripts": {
     "dev:laboratory": "next dev",
     "build:laboratory": "next build",
-    "lint": "eslint . --ext .js,.jsx,.ts,.tsx"
+    "lint": "eslint . --ext .js,.jsx,.ts,.tsx",
+    "playwright:start": "turbo run build; npm run dev:laboratory",
```
chris13524, line 9: Since I don't have `turbo` installed, I can't run `npm run playwright:test` locally. We should make a `npm run turbo-build -> turbo run build` script and call that here.

chris13524 (changes requested):
1 more thing 🙏

apps/laboratory/package.json
arein: I think I can remove the `turbo run build` from the command. Would that address your concern?

0xAsimetriq: You can run `npm run build` instead

chris13524: @arein the script is broken now; you can't run `npm run build` in the lab directory; that script does not exist. `turbo run build` worked because it recurses into the parent directory.

0xAsimetriq: You are right actually. For everything to work properly / build packages etc. Set up and commands should be ran from the monorepo root. Similarly something like `npm run laboratory` should be ran from there. Doing otherwise i.e. running from individual packages could have unexpected side effects probably

arein: ok fixed

chris13524 (approved):

0xAsimetriq (changes requested):
This sets everything up nicely, but does not include new approach we discussed with Jon, Sven and others for v3 testing.

Due to fast moving pace and changes, we'd need to spend substantial amount of time on re-updating tests if they are based on granular checks i.e. like you check for `All Wallets`.

Road forward with v3 that will keep us moving fast while providing good feedback on breaking something is through playwrights visual diff tests

More context here https://playwright.dev/docs/test-snapshots

TL;DR these are extremely cheap to set up dev time wise for existing and upcoming features while providing really good feedback on breaking changes i.e. if something is off, pixels will be off as well (missing notification. modal not open etc.). Updating these will be easy as well, i.e. if something changed and it was expected, we just update reference screenshot.

@arein Could you please switch from using "All Wallets" detection in the test, to one using screenshot diffing i.e. snap of open modal.

.github/workflows/ui_tests.yml
```diff
@@ -0,0 +1,53 @@
...
+      - name: Install Playwright Browsers
+        working-directory: ./apps/laboratory/
+        run: npm run playwright:install
+      - name: Run Playwright tests
+        env:
+          NEXT_PUBLIC_PROJECT_ID: ${{ secrets.NEXT_PUBLIC_PROJECT_ID }}
+        working-directory: ./apps/laboratory/
+        run: npm run playwright:test
+      - uses: actions/upload-artifact@v3
+        if: always()
```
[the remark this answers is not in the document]
chris13524: `if: always()` I believe allows it to still upload failure results e.g. screenshots when we set those up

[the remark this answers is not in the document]
arein: yes

arein:
> You are right actually. For everything to work properly / build packages etc. Set up and commands should be ran from the monorepo root. Similarly something like npm run laboratory should be ran from there. Doing otherwise i.e. running from individual packages could have unexpected side effects probably

What's actually requested?
As a developer I must be able to run e2e tests from monorepo root?

> This sets everything up nicely, but does not include new approach we discussed with Jon, Sven and others for v3 testing.
Due to fast moving pace and changes, we'd need to spend substantial amount of time on re-updating tests if they are based on granular checks i.e. like you check for All Wallets.
Road forward with v3 that will keep us moving fast while providing good feedback on breaking something is through playwrights visual diff tests
More context here https://playwright.dev/docs/test-snapshots
TL;DR these are extremely cheap to set up dev time wise for existing and upcoming features while providing really good feedback on breaking changes i.e. if something is off, pixels will be off as well (missing notification. modal not open etc.). Updating these will be easy as well, i.e. if something changed and it was expected, we just update reference screenshot.
@arein Could you please switch from using "All Wallets" detection in the test, to one using screenshot diffing i.e. snap of open modal.

I'm scaffolding everything here @0xAsimetriq specifically to avoid comments like these. I am not actually testing anything here. The `All Wallets` is just a placeholder tests. I don't see how this setup cannot be used later for actual diffs

apps/laboratory/package.json
```diff
@@ -5,7 +5,12 @@
   "scripts": {
     "dev:laboratory": "next dev",
     "build:laboratory": "next build",
-    "lint": "eslint . --ext .js,.jsx,.ts,.tsx"
+    "build:all": "turbo run build",
```
0xAsimetriq, line 8: Turborepo and monorepo wide commands should not be ran from here. There is global `npm run build` command that does exactly this in the root, can that be reused? It also sets up clean builds properly for all packages.

You will likely need to move playwright commands and test dependencies to monorepo root as well and just target tests inside laboratory folder.

0xAsimetriq (changes requested):

chris13524:
> Due to fast moving pace and changes, we'd need to spend substantial amount of time on re-updating tests if they are based on granular checks i.e. like you check for All Wallets.
> 
> Road forward with v3 that will keep us moving fast while providing good feedback on breaking something is through playwrights visual diff tests

Taking full advantage of Playwright will involve a lot more than screenshots; every edge case should have a test case and needing to update something trivial like "All Wallets" will be the least of your concerns when changing stuff around.

0xAsimetriq:
> > You are right actually. For everything to work properly / build packages etc. Set up and commands should be ran from the monorepo root. Similarly something like npm run laboratory should be ran from there. Doing otherwise i.e. running from individual packages could have unexpected side effects probably
> 
> What's actually requested? As a developer I must be able to run e2e tests from monorepo root?
> 
> > This sets everything up nicely, but does not include new approach we discussed with Jon, Sven and others for v3 testing.
> > Due to fast moving pace and changes, we'd need to spend substantial amount of time on re-updating tests if they are based on granular checks i.e. like you check for All Wallets.
> > Road forward with v3 that will keep us moving fast while providing good feedback on breaking something is through playwrights visual diff tests
> > More context here https://playwright.dev/docs/test-snapshots
> > TL;DR these are extremely cheap to set up dev time wise for existing and upcoming features while providing really good feedback on breaking changes i.e. if something is off, pixels will be off as well (missing notification. modal not open etc.). Updating these will be easy as well, i.e. if something changed and it was expected, we just update reference screenshot.
> > @arein Could you please switch from using "All Wallets" detection in the test, to one using screenshot diffing i.e. snap of open modal.
> 
> I'm scaffolding everything here @0xAsimetriq specifically to avoid comments like these. I am not actually testing anything here. The `All Wallets` is just a placeholder tests. I don't see how this setup cannot be used later for actual diffs

So for Q1, right now there seems to be a `build:all` setup in laboratory using global turborepo command. This should not be the case case, build depends on all packages, so it should be only ran and set up in root of the monorepo. I left separate suggestion about moving playwright tests to root as well.

0xAsimetriq:
> > Due to fast moving pace and changes, we'd need to spend substantial amount of time on re-updating tests if they are based on granular checks i.e. like you check for All Wallets.
> > Road forward with v3 that will keep us moving fast while providing good feedback on breaking something is through playwrights visual diff tests
> 
> Taking full advantage of Playwright will involve a lot more than screenshots; every edge case should have a test case and needing to update something trivial like "All Wallets" will be the least of your concerns when changing stuff around.

Yes, but remember this is ui product so majority of issues will bubble up to ui, which is why screenshot diffing is great here. I.e if you set an option to disable metamask, all you have to do is screenshot that and it should not be there, if it is, then we made a regression. This applies to almost every config / action in web3modal, if something is wrong it will result in a different visual to what is expected (even text in error notifications for example)

chris13524:
> Yes, but remember this is ui product so majority of issues will bubble up to ui, which is why screenshot diffing is great here. I.e if you set an option to disable metamask, all you have to do is screenshot that and it should not be there, if it is, then we made a regression. This applies to almost every config / action in web3modal, if something is wrong it will result in a different visual to what is expected (even text in error notifications for example)

Yes, but I feel like this is the easy way out and is easier to mess up. There's 1 command that will reset all of the screenshots. So say you have a new feature that adds a new wallet to the list of wallets, and you reset all of the screenshots. You get your new wallet, but you also removed MetaMask without realizing it. Simply the screenshot checks will not detect this problem.

Now since the screenshots are committed and reviewed, hopefully someone will notice that the one called "MetaMask exists" does indeed still contain a MetaMask.

I guess my point is that it's easier to mess up an image than an explicit code removal/modifiction of the MetaMask test case.

I haven't used these snapshots before, so maybe I'll see the light :D

PR #1588, opened 2023-12-19
feat: wagmi v2

tomiir:
# Breaking Changes
- Users will be required to use viem2 and tanstack query in order for this to work.
- Around half of Wagmi's interfaces changed.
- `WagmiConfig` is no longer supported, `WagmiProvider` should be used instead.
- Wagmi's hooks are no longer valid when using the new client.

For complete notes see: 
- Core: https://rc.wagmi.sh/core/guides/migrate-from-v1-to-v2
- React: https://rc.wagmi.sh/react/guides/migrate-from-v1-to-v2

# Changes

- feat: wagmi-v2 integration
- fix:
- chore:

# Notes
- Created new V4 branch off of commit 707c5053. 
- v2 includes multi account support oob. This means we no longer need our custom connector and no longer flag these connectors as ANNOUNCED. Maybe there's something we want to do here in order to keep in line with other versions of the lib?
- Right now this is using default viem `http` transport. We were previously providing the WC Blockchain API transport as well but now we need to expose a client with a specific transport. Do we want to use the WC one as main or keep the public default client as main? 

# Associated Issues
https://github.com/WalletConnect/web3modal/issues/1587


apps/laboratory/src/pages/library/wagmi-v2.tsx
```diff
@@ -0,0 +1,98 @@
...
+  enableAnalytics: true,
+  metadata,
+  termsConditionsUrl: 'https://walletconnect.com/terms',
+  privacyPolicyUrl: 'https://walletconnect.com/privacy'
+})
+
+ThemeStore.setModal(modal)
+
+export default function Wagmi() {
+  const [ready, setReady] = useState(false)
```
glitch-txs, line 75: this can be removed, we could add SSR config instead: https://beta.wagmi.sh/react/guides/ssr

examples/react-wagmi/src/App.tsx
```diff
@@ -21,7 +25,11 @@ const wagmiConfig = defaultWagmiConfig({
   chains,
   projectId,
   metadata: {
-    name: 'Web3Modal React Example'
+    name: 'Web3Modal React Example',
+    description: 'Web3Modal React Example',
+    url: '',
+    icons: [],
+    verifyUrl: ''
```
glitch-txs, line 32: Do we actually use this flag?

glitch-txs (approved):

packages/wagmi/src/utils/defaultWagmiCoreConfig.ts
```diff
@@ -33,46 +31,44 @@ export function defaultWagmiConfig({
...
+    connectors.push(emailConnector({ chains, options: { projectId } }))
   }
 
   return createConfig({
-    autoConnect: true,
+    chains: chains as [Chain, ...Chain[]],
+    client: ({ chain }) =>
+      createClient({
+        chain,
+        // TOD0: How to use WC transport? Do we need it for analytics?
```
glitch-txs, line 68: For this, what about wrapping the `http` function with a new function called `web3modalTransport`

the config could work like this:

```ts
export const config = createConfig({
  chains: [mainnet, sepolia],
  connectors: [injected()],
  transports: {
    [mainnet.id]: web3modalTransport(mainnet.id), 
    [sepolia.id]: web3modalTransport(sepolia.id), 
  },
})
```

internally

```ts
  client({ chain }) {
    return createClient({ chain, transport: web3modalTransport(chain.id) })
  },
```

glitch-txs: also, there are usually issues with our RPC because it doesn't support some chains.
Since we'll be getting the chain id here we could check and through a custom message like 
"Web3Modal Transport does not support chain X, please use a different Transport using the custom configuraiton"

glitch-txs:
made another QA round and found a couple of issues:

1. For some reason I cannot connect with Trust Wallet extension
2. I fI connect to MM and change chain to a non-supported change it desconnects, (tried switching to Linear)

glitch-txs:
There might be errors that come from Wagmi, they are constantly updating v2 and fixing them*

glitch-txs:
closes https://github.com/WalletConnect/web3modal/issues/1718

svenvoskamp:
Did some QA, seems to work fine apart from a small UX issue: 

When connecting with wc, all requested chains show up in the network view. We only want to show supported networks.

svenvoskamp (approved):

glitch-txs:
image

packages/wagmi/tsconfig.json
```diff
@@ -3,12 +3,8 @@
     "composite": true,
     "outDir": "./dist/esm",
     "declarationDir": "./dist/types",
-    "baseUrl": "../../",
+    "baseUrl": "../../"
     // https://github.com/microsoft/TypeScript/issues/48212
```
0xAsimetriq, line 7: Let's remove this comment as well if it's no longer needed

packages/wagmi/src/client.ts
```diff
@@ -324,37 +301,45 @@ export class Web3Modal extends Web3ModalScaffold {
...
     })
     this.setBalance(balance.formatted, balance.symbol)
   }
 
   private syncConnectors(wagmiConfig: Web3ModalClientOptions['wagmiConfig']) {
     const w3mConnectors: Connector[] = []
-    wagmiConfig.connectors.forEach(({ id, name }) => {
-      if (![ConstantsUtil.EIP6963_CONNECTOR_ID, ConstantsUtil.EMAIL_CONNECTOR_ID].includes(id)) {
+
+    const coinbaseConnector = wagmiConfig.connectors.find(c => c.id === 'com.coinbase.wallet')
```
0xAsimetriq, line 330: all coinbase strings checked here are better moved to be constants.

0xAsimetriq:
Few notes after manual testing

1. Email login + wagmi, sending test transaction to vitalik doesn't seem to work (throws error within walelt view), while it works fine on lab.web3modal.com

2. On metamask mobile chain switching doesn't seem to work, but I'm not 100% sure it's not a metamask issue. Id did work for me in lab.web3modal.com though

0xAsimetriq (commented):
Was wagmi1 package set up properly within root package.json, lerna and turborepo files? Can't seem to find references to these

packages/scaffold-utils/src/PresetsUtil.ts
```diff
@@ -64,6 +65,7 @@ export const PresetsUtil = {
     [ConstantsUtil.INJECTED_CONNECTOR_ID]: 'Browser Wallet',
     [ConstantsUtil.WALLET_CONNECT_CONNECTOR_ID]: 'WalletConnect',
     [ConstantsUtil.COINBASE_CONNECTOR_ID]: 'Coinbase',
+    [ConstantsUtil.COINBASE_SDK_CONNECTOR_ID]: 'Coinbase',
```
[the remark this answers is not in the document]
tomiir: They are in fact different, just map to the same value. 

`COINBASE_CONNECTOR_ID` => Injected coinbase wallet extension
`COINBASE_SDK_CONNECTOR_ID` => coinbase SDK connector injected by wagmi

packages/scaffold/src/views/w3m-connect-view/index.ts
```diff
@@ -170,7 +168,7 @@ export class W3mConnectView extends LitElement {
       return html`
         <wui-list-wallet
           imageSrc=${ifDefined(AssetUtil.getConnectorImage(connector))}
-          .installed=${Boolean(announced)}
+          .installed=${true}
```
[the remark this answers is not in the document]
tomiir: hm good point, I'm going to run some tests in mobile but I'm not sure how 6963 works on mobile environments 🤔

packages/wagmi/src/utils/defaultWagmiReactConfig.ts
```diff
@@ -1,62 +1,68 @@
...
-    autoConnect: true,
+    chains: chains as [Chain, ...Chain[]],
+    client: ({ chain }) =>
+      createClient({
+        chain,
+        transport: http()
+      }),
     connectors,
-    publicClient
+    multiInjectedProviderDiscovery: enableEIP6963 !== false
```
glitch-txs, line 66: in order to support more features (like SSR) in our defaultWagmiConfig can we add a spread operator?

```suggestion
    multiInjectedProviderDiscovery: enableEIP6963 !== false,
    ...wagmiOptions
```

PR #1591, opened 2023-12-20
chore(next-auth-url): add missing env var NEXTAUTH_URL

Cali93:
# Breaking Changes

`NEXTAUTH_URL` is required to run SIWE locally.

# Changes

- feat:
- fix:
- chore(next-auth-url): add missing env var NEXTAUTH_URL


0xAsimetriq (commented):
Looks like package-lock is updated, but there are no package.json changes. Is this expected?

0xAsimetriq (commented):
You'll also want to update dangerjs rule to exclude checking for localhost in these files or, simply remove localhost as default value in favor of `""` <- this might be better to not polute dangerfile

0xAsimetriq (dismissed):

Cali93:
I just pulled V3 and `npm run clean; npm install` and I still have a diff for package-lock.json 👀

PR #1646, opened 2023-12-21
feat(ui-tests): cover ethers

arein:
# Breaking Changes

N/A

# Changes

- feat(ui-tests): cover ethers

# Associated Issues

closes #...

### DIFF
--- a/apps/laboratory/tests/shared/constants/index.ts
+++ b/apps/laboratory/tests/shared/constants/index.ts
@@ -1,9 +1,7 @@
 import type { SessionParams } from '../types'
 
 // Allow localhost
-export const BASE_URL = 'http://localhost:3000/library/'
-// Allow localhost
-export const LOCAL_LABS_URL = `${BASE_URL}wagmi/`
+export const BASE_URL = 'http://localhost:3000/'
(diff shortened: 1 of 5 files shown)


chris13524 (dismissed):
Very cool approach!

apps/laboratory/tests/shared/constants/index.ts
```diff
@@ -1,7 +1,9 @@
 import type { SessionParams } from '../types'
 
 // Allow localhost
-export const LOCAL_LABS_URL = 'http://localhost:3000/library/wagmi/'
+export const BASE_URL = 'http://localhost:3000/library/'
+// Allow localhost
+export const LOCAL_LABS_URL = `${BASE_URL}wagmi/`
```
tomiir, line 6: why set the base url to wagmi?

arein: `BASE_URL` now is not specific. `LOCAL_LABS_URL` is the entrypoint for the test and needs to be a website that doesn't HTTP404 so I chose this one. Could also use `http://localhost:3000/` I presume

arein: Fixed

tomiir (approved):

PR #1647, opened 2023-12-21
chore(ui-tests): cover sign

arein:
# Changes

- chore(ui-tests): cover sign

@chris13524 @tomiir apologies I decided to expand this PR a bit to actually introduce more coverage.

### DIFF
--- a/apps/laboratory/tests/basic-tests.spec.ts
+++ b/apps/laboratory/tests/basic-tests.spec.ts
@@ -2,7 +2,7 @@ import { testM, expect } from './shared/fixtures/w3m-fixture'
 
 testM.describe('Modal only tests', () => {
   testM('Should be able to open modal', async ({ modalPage }) => {
-    await modalPage.page.getByText('Connect Wallet').click()
-    await expect(modalPage.page.getByText('All Wallets')).toBeVisible()
+    await modalPage.page.getByTestId('connect-button').click()
+    await expect(modalPage.page.getByTestId('all-wallets')).toBeVisible()
   })
 })
--- a/apps/laboratory/tests/shared/pages/WalletPage.ts
+++ b/apps/laboratory/tests/shared/pages/WalletPage.ts
@@ -36,12 +36,15 @@ export class WalletPage {
    */
   async handleSessionProposal(opts: SessionParams) {
     const variant = opts.accept ? `approve` : `reject`
+    await this.page.getByTestId(`session-${variant}-button`).isEnabled()
     await this.page.getByTestId(`session-${variant}-button`).focus()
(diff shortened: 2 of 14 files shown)


tomiir (dismissed):
🔥 
Just thinking ahead but would be nice to have some kind of rules for test-id names in the future.

apps/laboratory/tests/basic-tests.spec.ts
```diff
@@ -2,7 +2,7 @@ import { testM, expect } from './shared/fixtures/w3m-fixture'
 
 testM.describe('Modal only tests', () => {
   testM('Should be able to open modal', async ({ modalPage }) => {
-    await modalPage.page.getByText('Connect Wallet').click()
-    await expect(modalPage.page.getByText('All Wallets')).toBeVisible()
+    await modalPage.page.getByTestId('connect-button').click()
```
chris13524, line 5: Can we use a global constant for this so we have static type checking? Then the test ID itself can be pretty arbitrary, and it's easier to rename everything as things evolve.

chris13524 (dismissed):

apps/laboratory/tests/basic-tests.spec.ts
arein: Made myself a follow up ticket for it https://github.com/WalletConnect/web3modal/issues/1582

apps/laboratory/tests/shared/pages/WalletPage.ts
```diff
@@ -36,12 +36,15 @@ export class WalletPage {
...
     await this.page.getByTestId(`session-${variant}-button`).focus()
     await this.page.keyboard.press('Space')
   }
 
   async handleRequest({ accept }: { accept: boolean }) {
     const variant = accept ? `approve` : `reject`
-    await this.page.getByTestId(`request-button-${variant}`).click()
+    await this.page.getByTestId(`session-${variant}-button`).isEnabled()
+    await this.page.getByTestId(`session-${variant}-button`).focus()
+    await this.page.keyboard.press('Space')
```
chris13524, line 48: Why `.press('Space')` instead of `.click()`?

chris13524 (approved):
LGTM

apps/laboratory/tests/shared/pages/WalletPage.ts
arein: Took me a while to figure out. Will add a comment in the next iteration. But for whatever reason on Chromium it has to be this flavor and `click` doesn't work for this specific button. In Firefox it works. I can click myself.
Low prio to change this in react-wallet which is used here but managed in a different repo

PR #1669, opened 2023-12-26
fix: account button hide balance on small device views

tomiir:
# Breaking Changes

N/A

# Changes

- feat:
- fix: wui-account-button responsive styles. Removed balance on small screens.
- chore:

# Associated Issues
https://github.com/WalletConnect/web3modal/issues/1668


svenvoskamp (dismissed):
Approved, but wondering why we have different height for mobile

svenvoskamp (approved):

PR #1676, opened 2023-12-28
fix: email +  UI responsive issues

tomiir:
# Breaking Changes

N/A

# Changes

- feats: 
  - add `charsStart` and  `charsEnd` properties to `wui-account-button`.
  - add OTP timeout timer
  - only enable input once previous number has been enetered
  - add local email validation
- fix: responsive styling for `wui-account-button` and `wui-connect-button`
- chore: remove connect button from standalone email integration.

## Screenshots

image
image

https://github.com/WalletConnect/web3modal/assets/12738697/bdb4a329-abe2-4d33-9428-7696b5f2a329

# Related Issues
https://github.com/WalletConnect/web3modal/issues/1624
[Add new component for "Resend Email" with countdown#1627](https://github.com/WalletConnect/web3modal/issues/1627)
[Web3Modal Copy Changes#1628](https://github.com/WalletConnect/web3modal/issues/1628)


packages/scaffold/src/partials/w3m-email-login-widget/index.ts
```diff
@@ -71,47 +75,41 @@ export class W3mEmailLoginWidget extends LitElement {
...
     this.email = event.detail
+    this.error = ''
   }
 
   private async onSubmitEmail(event: Event) {
     try {
       if (this.loading) {
         return
       }
+      const { success: isEmailValid } = z.string().email().safeParse(this.email)
```
0xAsimetriq, line 100: Is this validation necessary here? When you submit invalid email it should be already validated upstream through schema of `wallet` package and throw an error that's shown to user.

packages/scaffold/src/partials/w3m-email-login-widget/styles.ts
```diff
@@ -24,7 +24,7 @@ export default css`
   wui-icon-link,
   wui-loading-spinner {
     position: absolute;
-    top: 50%;
+    top: 26px;
```
0xAsimetriq, line 27: If I remember correctly hardcoding this value caused layout issues in safari or firefox, could you validate that everything looks as expected there? If possible we should use dynamic values like these to account for potential ui component size changes in the future.

packages/scaffold/src/views/w3m-email-verify-otp-view/index.ts
```diff
@@ -26,12 +26,26 @@ export class W3mEmailVerifyOtpView extends LitElement {
   // -- State & Properties -------------------------------- //
   @state() private loading = false
 
+  @state() private timeoutTimeLeft = 30
```
0xAsimetriq, line 29: Logic for email timeout already exists here in `wallet` package under helpers here
https://github.com/WalletConnect/web3modal/blob/V3/packages/wallet/src/W3mFrameHelpers.ts#L31

It's also more robust and based on `localStorage`. You should make a helper within core perhaps, that uses that value (if it exists) to bootstrap timers here.

packages/scaffold/src/partials/w3m-email-login-widget/index.ts
tomiir: this was to prevent the api throwing an error + showing the error toast since this is a local validation and we'd need to show the input's error message. I guess I can remove this and somehow parse actual error but seemed a bit hackier than this

packages/scaffold/src/partials/w3m-email-login-widget/styles.ts
tomiir: hm fair enough, will double check. 

This was changed because when the input has an error text below, it's height grows, and the 50% makes it so the icon slides out of the input. 

Maybe there's a way to move the relative point on the input so it properly considers the icon positioning when errors occur. 
Maybe there's a container where we can fix its height + make it relative and would fix it elegantly

packages/scaffold/src/partials/w3m-email-login-widget/index.ts
tomiir: effectively checked the validation type, not the cleanest but it works without relying on zod. 

might be worth reviewing what other kinds of errors could happen that should show the input error instead of the snackbar error. from Diane's notes, it's anything related to input validation but can't think of anything else other than mail formatting right now.

packages/wallet/src/W3mFrameHelpers.ts
```diff
@@ -37,5 +37,17 @@ export const W3mFrameHelpers = {
         throw new Error(`Please try again after ${cooldownSec} seconds`)
       }
     }
+  },
+
+  getTimeToNextEmailLogin() {
+    const lastEmailLoginTime = W3mFrameStorage.get(W3mFrameConstants.LAST_EMAIL_LOGIN_TIME)
+    if (lastEmailLoginTime) {
+      const difference = Date.now() - Number(lastEmailLoginTime)
+      if (difference < 30_000) {
```
svenvoskamp, line 46: I think we should store `30_000` in a constant

PR #1677, opened 2023-12-28
feat(ui-test): cover reject signing

arein:
We were covering successful signing. This now also covers the wallet rejecting the sign request.

# Changes

- feat(ui-test): cover reject signing

# Associated Issues

closes #...


tomiir (approved):
Thanks for the comments 🙏

apps/laboratory/tests/shared/validators/ModalValidator.ts
```diff
@@ -13,6 +14,12 @@ export class ModalValidator {
   }
 
   async expectAcceptedSign() {
-    await expect(this.page.getByText('Success')).toBeVisible()
+    // We use Chakra Toast and it's not quite straightforward to set the `data-testid` attribute on the toast element.
```
chris13524, line 17: If you can select by css class or another element pattern that would work too

chris13524 (approved):

PR #1679, opened 2023-12-29
feat(ui-test): cover siwe

arein:
Adds a simple happy-case Siwe test.

# Changes

- feat(ui-test): cover siwe

# Associated Issues

closes #...

### DIFF
--- a/.github/workflows/ui_tests.yml
+++ b/.github/workflows/ui_tests.yml
@@ -43,6 +43,7 @@ jobs:
       - name: Run Playwright tests
         env:
           NEXT_PUBLIC_PROJECT_ID: ${{ secrets.NEXT_PUBLIC_PROJECT_ID }}
+          NEXTAUTH_SECRET: ${{ secrets.TESTS_NEXTAUTH_SECRET }}
         working-directory: ./apps/laboratory/
         run: npm run playwright:test
       - uses: actions/upload-artifact@v3
--- a/apps/laboratory/tests/shared/fixtures/w3m-wallet-fixture.ts
+++ b/apps/laboratory/tests/shared/fixtures/w3m-wallet-fixture.ts
@@ -1,4 +1,4 @@
-import { testM as base } from './w3m-fixture'
+import { testM as base, testMSiwe as siwe } from './w3m-fixture'
 import { WalletPage } from '../pages/WalletPage'
 import { WalletValidator } from '../validators/WalletValidator'
 
@@ -10,6 +10,23 @@ interface ModalWalletFixture {
 
 // MW -> test Modal + Wallet
 export const testMW = base.extend<ModalWalletFixture>({
+  walletPage: async ({ context, browserName }, use) => {
+    // WalletPage needs clipboard permissions with chromium to paste URI
+    if (browserName === 'chromium') {
+      await context.grantPermissions(['clipboard-read', 'clipboard-write'])
+    }
+
+    // Use a new page, to open alongside the modal
+    const walletPage = new WalletPage(await context.newPage())
+    await walletPage.load()
+    await use(walletPage)
+  },
+  walletValidator: async ({ walletPage }, use) => {
+    const walletValidator = new WalletValidator(walletPage.page)
+    await use(walletValidator)
+  }
+})
+export const testMWSiwe = siwe.extend<ModalWalletFixture>({
   walletPage: async ({ context, browserName }, use) => {
     // WalletPage needs clipboard permissions with chromium to paste URI
     if (browserName === 'chromium') {
--- a/apps/laboratory/tests/shared/pages/ModalPage.ts
+++ b/apps/laboratory/tests/shared/pages/ModalPage.ts
@@ -5,20 +5,25 @@ export class ModalPage {
   private readonly baseURL = BASE_URL
 
   private readonly connectButton: Locator
+  private readonly url: string
 
   constructor(
     public readonly page: Page,
-    public readonly library: string
+    public readonly library: string,
+    public readonly siwe: boolean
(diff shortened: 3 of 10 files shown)


.github/workflows/ui_tests.yml
```diff
@@ -43,6 +43,7 @@ jobs:
       - name: Run Playwright tests
         env:
           NEXT_PUBLIC_PROJECT_ID: ${{ secrets.NEXT_PUBLIC_PROJECT_ID }}
+          NEXTAUTH_SECRET: ${{ secrets.TESTS_NEXTAUTH_SECRET }}
```
chris13524, line 46: Does this need to be a securely random secret value or would something arbitrary like "password" suffice? Since this server is only accessible inside the runner and it's not protecting any private info

apps/laboratory/tests/shared/pages/ModalPage.ts
```diff
@@ -5,20 +5,25 @@ export class ModalPage {
   private readonly baseURL = BASE_URL
 
   private readonly connectButton: Locator
+  private readonly url: string
 
   constructor(
     public readonly page: Page,
-    public readonly library: string
+    public readonly library: string,
+    public readonly siwe: boolean
```
chris13524, line 13: I'd advise against boolean parameters like this. Would prefer a string union type (e.g. `"siwe" | "regular"`) or better abstraction

apps/laboratory/tests/shared/validators/ModalValidator.ts
```diff
@@ -9,6 +9,10 @@ export class ModalValidator {
     await expect(this.page.getByTestId('account-button')).toBeVisible()
   }
 
+  async expectAuthenticated() {
+    await expect(this.page.getByTestId('w3m-authentication-status')).toContainText('authenticated')
```
chris13524, line 13: Blocker: `unauthenticated` contains `authenticated`. You may want to include the `Status:` prefix

apps/laboratory/tests/shared/fixtures/w3m-wallet-fixture.ts
```diff
@@ -26,4 +26,21 @@ export const testMW = base.extend<ModalWalletFixture>({
     await use(walletValidator)
   }
 })
+export const testMWSiwe = siwe.extend<ModalWalletFixture>({
+  walletPage: async ({ context, browserName }, use) => {
+    // WalletPage needs clipboard permissions with chromium to paste URI
+    if (browserName === 'chromium') {
```
chris13524, line 32: Seems like a lot of duplicate code here. Can there be a better abstraction? At least make the clipboard permission thing a function

chris13524 (changes requested):
1 blocker, rest are non-blocking

.github/workflows/ui_tests.yml
arein: as far as I understand it can be any random value. However, I don't see a problem with this approach. Dislike putting anything that's suffixed with `SECRET` in cleartext or a random value in code

apps/laboratory/tests/shared/pages/ModalPage.ts
arein: decided against inheritance/polymorphism for now. when `email` comes this needs to be refactored to accommodate this all better

apps/laboratory/tests/shared/fixtures/w3m-wallet-fixture.ts
arein: Once `email` comes this will deffo be too much copy/pasta. I sort of deliberately went for copy/pasta here so when `email` comes I can look at the right way of doing this. Other options to look at are polymorphism/inheritance.

chris13524: Agreed; don't want to over-abstract too soon

apps/laboratory/tests/shared/validators/WalletValidator.ts
```diff
@@ -15,7 +15,7 @@ export class WalletValidator {
   }
 
   async expectDisconnected() {
-    await this.page.waitForTimeout(1000)
+    await this.page.waitForTimeout(2000)
     await this.page.reload()
     await this.gotoSessions.click()
     await expect(this.page.getByTestId('session-card')).not.toBeVisible()
```
chris13524: Any way to do this without an artificial delay?

Does `session-card` update dynamically? If so, you should be able to just wait for the condition without a timeout or reload at all?

If not, you might be able to use https://playwright.dev/docs/test-assertions#expectpoll or another polling mechanism.

apps/laboratory/tests/shared/pages/ModalPage.ts
```diff
@@ -1,24 +1,32 @@
...
   private readonly baseURL = BASE_URL
 
   private readonly connectButton: Locator
+  private readonly url: string
 
   constructor(
     public readonly page: Page,
-    public readonly library: string
+    public readonly library: string,
+    public readonly flavor: ModalFlavor
```
chris13524, line 15: Perfect!

chris13524 (approved):

Cali93 (approved):
Waw, ty so much for this 🙏

apps/laboratory/tests/shared/validators/WalletValidator.ts
arein: Yes this needs to be removed. background: I think this was related to the fact that using `ethers` the `await` on `disconnect` immediately resolves.

[I'll try your suggestion](https://github.com/WalletConnect/web3modal/issues/1582)

PR #1681, opened 2023-12-29
feat(ui-test): cache Playwright

arein:
Shoves ~30s off the test runtime by caching playwright.

# Changes

- feat(ui-test): cache Playwright

# Associated Issues

closes #...

### DIFF
--- a/.github/workflows/ui_tests.yml
+++ b/.github/workflows/ui_tests.yml
@@ -33,17 +33,29 @@ jobs:
           node-version: 18.x
           cache: 'npm'
           cache-dependency-path: 'package-lock.json'
+      - name: Get installed Playwright version
+        id: playwright-version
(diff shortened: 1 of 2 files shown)


.github/workflows/ui_tests.yml
```diff
@@ -33,17 +33,29 @@ jobs:
           node-version: 18.x
           cache: 'npm'
           cache-dependency-path: 'package-lock.json'
+      - name: Get installed Playwright version
+        id: playwright-version
+        run: echo "PLAYWRIGHT_VERSION=$(node -e "console.log(require('./package-lock.json').packages['apps/laboratory']['devDependencies']['@playwright/test'])")" >> $GITHUB_ENV
```
tomiir, line 38: Maybe we could abstract this to a script to mantain it more easily. Also, we could maybe simplify it by using `npm list` + grep. Something like `npm list @playwright/test | grep /^@/` or similar. Not sure if it works as flawlessly as this though

tomiir (approved):

PR #1689, opened 2024-01-01
fix: dayjs locale format seconds is not working and is polluting the global dayjs locale

nikosrossolatos:
# Breaking Changes
N/A

# Changes

- fix: `getRelativeDateFromNow` will no longer return `%s sec` when seconds are provided.
  - dayjs@1.11.10 is expecting `"%d sec"` instead of `"%s sec"` ([here's why](https://github.com/iamkun/dayjs/blob/dev/src/plugin/relativeTime/index.js#L58)) to make the proper conversion to "10 sec"
- fix: do not pollute the global `en` locale 
  - We found out this bug while using this library on our app. It shouldn't be affecting _every_ app or library that uses this


packages/common/src/utils/DateUtil.ts
```diff
@@ -1,15 +1,18 @@
...
 
-dayjs.updateLocale('en', {
+const localeObject = {
+  ...englishLocale,
+  name: 'en-custom',
   relativeTime: {
     future: 'in %s',
     past: '%s ago',
-    s: '%s sec',
+    s: '%d sec',
```
nikosrossolatos, line 15: This is the main fix

packages/common/src/utils/DateUtil.ts
```diff
@@ -21,14 +24,16 @@ dayjs.updateLocale('en', {
...
+dayjs.locale('en-custom', localeObject)
 
 export const DateUtil = {
   getYear(date: string = new Date().toISOString()) {
     return dayjs(date).year()
   },
 
   getRelativeDateFromNow(date: string | number) {
-    return dayjs(date).fromNow(true)
+    return dayjs(date).locale('en-custom').fromNow(true)
```
nikosrossolatos, line 37: This is how I would propose to fix the global pollution of the dayjs en locale. Happy to try other solutions as well but I think this clean enough

packages/common/src/utils/DateUtil.ts
```diff
@@ -21,14 +24,16 @@ dayjs.updateLocale('en', {
...
+dayjs.locale('en-web3-modal', localeObject)
 
 export const DateUtil = {
   getYear(date: string = new Date().toISOString()) {
     return dayjs(date).year()
   },
 
   getRelativeDateFromNow(date: string | number) {
-    return dayjs(date).fromNow(true)
+    return dayjs(date).locale('en-web3-modal').fromNow(true)
```
nikosrossolatos, line 37: This is how I would propose to fix the global pollution issue. Happy to try other solutions but I think this is clean enough

enesozturk (approved):
LGTM

nikosrossolatos:
> LGTM

@enesozturk How can I run the vercel or the rest of the checks since I am missing authorisation so that we can merge this PR? :D

enesozturk:
@nikosrossolatos seems interesting, I'll investigate this and let you know. The issues are not expected so it shouldn't be a problem for people outside of the team, it's something we should fix unless you touch the GH workflow code.

nikosrossolatos:
> Vercel ones need to be authorized. I authorized labs, if this is good we can just merge it

Yeap that's what it saying for me too 😅 missing authorisation. I'll leave this to you then to merge if it's okay

tomiir (approved):

tomiir (approved):

PR #1691, opened 2024-01-02
fix: Deduplicate EIP6963 wallets from connectors

0xAsimetriq:
# Breaking Changes

N/A

# Changes

- fix: Deduplicate eip6963 wallets from connectors based on hardcoded map
- chore: Update dependencies

# Associated Issues

N/A


enesozturk (approved):
LGTM

PR #1698, opened 2024-01-03
Fix/wallet locked issue

svenvoskamp:
# Changes

- fix: Fix wallet locked autoconnect issue


packages/wagmi/src/connectors/EIP6963Connector.ts
```diff
@@ -74,10 +74,18 @@ export class EIP6963Connector extends InjectedConnector {
       if (!eip6963Wallet || connectedEIP6963Rdns !== eip6963Wallet.info.rdns) {
         return true
       }
+    }
+
+    if (
+      (eip6963Wallet && eip6963Wallet.provider._state?.isUnlocked === undefined) ||
+      (eip6963Wallet && eip6963Wallet.provider._state?.isUnlocked)
```
tomiir, line 81: wouldn't it be the same to check 
if (eip6963Wallet && eip6963Wallet.provider?._state?.isUnlocked === true) { return true }

glitch-txs (commented):
This is not a standard and some wallets might not have it, let's rather check for eth_account which should return an empty array when locked or disconnected.

packages/wagmi/src/connectors/EIP6963Connector.ts
```diff
@@ -74,10 +74,18 @@ export class EIP6963Connector extends InjectedConnector {
       if (!eip6963Wallet || connectedEIP6963Rdns !== eip6963Wallet.info.rdns) {
```
glitch-txs, line 74: shouldn't this be the other way around?

glitch-txs:
```suggestion
      if (eip6963Wallet && connectedEIP6963Rdns === eip6963Wallet.info.rdns) {
        const accounts = (await getProvider()).request({ method: 'eth_accounts' })
        if(accounts.length) return true
      }
```

could you try this?

PR #1702, opened 2024-01-04
Fix ethers loading hanging state

svenvoskamp:
# Changes

- fix: Fix ethers loading hanging state


arein (commented):
Can you provide an explainer of what the bug was? There's no test updates so it's hard to assess what the issue is?

svenvoskamp:
> Can you provide an explainer of what the bug was? There's no test updates so it's hard to assess what the issue is?

There was a state when `getLoginEmailUsed()` returns `true`, but connection couldn't succeed, because of `Magic RPC Error: [-32603] Internal error: User denied account access.` which kept button in loading state.

tomiir (approved):
LGTM!

Issue #1707, opened 2024-01-04
Peanut : User Avatar is stretched


tomiir:
**Issue**
 - User Avatar on Peanut is stretched
![Image](https://github.com/WalletConnect/web3modal/assets/84375331/67e2bcf5-d5fe-4a77-926a-26c2d23c0311)

**Suggested Solution from Rocky**
> Seems they are setting --w3m-border-radius-master to 0 which removes the radius from everything in the modal. I think the avatar is generated to be a circle in this case, if that's the case then we could either change the generated avatar or remove the dependency to the global variable and keep it circled always (this means reducing customization a bit)

Lets go with this ^^

PR #1708, opened 2024-01-04
fix: wui-avatar gradient algorithm

tomiir:
# Breaking Changes

N/A

# Changes

- feat:
- fix: avatar looking weird when border-radius was customized via `w3m-border-radius-master` as it's intended to be round.
- chore:

# Associated Issues

closes [#1707](https://github.com/WalletConnect/web3modal/issues/1707)


0xAsimetriq (commented):
We ideally want to keep this connected to theme's border radius as dapps might have brutalist / squarish design needs. Plus this will make ens avatars rounded all the time as well. What if you were to adjust algorithm that produces default avatar image instead or "scale it" up proportionally to border radius so it doesn't have those weird edges in square frame?

tomiir:
Came up with something that might work! 
https://github.com/WalletConnect/web3modal/assets/12738697/6e353279-56c2-41bd-bc54-298512304712

arein (dismissed):
LGTM but I don't know CSS enough and it seems the commented out stuff can be removed?

packages/ui/src/composites/wui-avatar/styles.ts
```diff
@@ -52,7 +52,7 @@ export default css`
   :host([data-variant='generated']) {
     box-shadow: 0 0 0 8px var(--wui-gray-glass-005);
     background: radial-gradient(
-      75.29% 75.29% at 64.96% 24.36%,
+      /* 75.29% 75.29% at 64.96% 24.36%, */ var(--local-radial-circle),
```
0xAsimetriq, line 55: If this comment is no longer needed, let's remove

0xAsimetriq (approved):

arein (approved):
LGTM!

PR #1725, opened 2024-01-08
feat: solana integration

gooxpf:
# Implement Solana

N/A

# Changes
Add 
- feat:
- Created Solana package  based on ethers and wagmi packages
- Added Solana utils in the scaffold-utils based on ethers

# Associated Issues

closes #...


packages/solana/src/connectors/WalletConnectConnector.ts
```diff
@@ -0,0 +1,311 @@
...
+   * If `qrcode = false`, this will return the pairing URI used to generate the
+   * QRCode.
+   */
+  public async connect(useURI?: boolean) {
+    const chainsNamespaces = [
+      'solana:' + solana.chainId, 'solana:' + solanaTestnet.chainId, 'solana:' + solanaDevnet.chainId
+    ]
+    const rpcMap = {
+      ['solana:' + solana.chainId]: solana.rpcUrl,
+      ['solana:' + solanaTestnet.chainId]: solanaTestnet.rpcUrl,
```
vcoolish, line 269: could you move testnets to optional namespaces?

apps/laboratory/tests/wallet.spec.ts
```diff
@@ -19,10 +19,17 @@ testConnectedMW.afterEach(async ({ modalPage, modalValidator, walletValidator, b
...
+      await modalPage.sign()
+      await walletValidator.expectReceivedSign({ chainName: 'Solana' })
+      await walletPage.handleRequest({ accept: true })
+      await modalValidator.expectAcceptedSign()
+    } else {
+      await modalPage.sign()
+      await walletValidator.expectReceivedSign({ chainName: DEFAULT_CHAIN_NAME })
+      await walletPage.handleRequest({ accept: true })
+      await modalValidator.expectAcceptedSign()
+    }
```
tomiir, line 32: I think this can be reduced to 
```
      const chainName = modalPage.library === 'solana' ? 'Solana' : DEFAULTCHAIN_NAME
      await modalPage.sign()
      await walletValidator.expectReceivedSign({ chainName })
      await walletPage.handleRequest({ accept: true })
      await modalValidator.expectAcceptedSign()
      ```

File: packages/core/src/controllers/NetworkController.ts [hunk 3]
```diff
@@ -94,6 +95,9 @@ export const NetworkController = {
   async switchActiveNetwork(network: NetworkControllerState['caipNetwork']) {
     await this._getClient().switchCaipNetwork(network)
     state.caipNetwork = network
+    if (ApiController && CoreHelperUtil.checkCaipNetwork(network, 'solana')) {
+      await ApiController.reFetchWallets()
+    }
```
tomiir, line 100: curious: why is a reset required here for solana specifically?

packages/scaffold-utils/src/SolanaHelpersUtils.ts
```diff
@@ -0,0 +1,71 @@
...
+import type { Chain, Provider } from './SolanaTypesUtil.js'
+
+export const SolHelpersUtil = {
+  getChain(chains: Chain[], chainId: string | null) {
+    const chain = chains.find(lChain => lChain.chainId === chainId)
+    if (chain) {
+      return chain
+    }
+
+    return chains[0]
```
tomiir, line 13: wouldn't this cause unexpected behaviors? should we maybe have a hardcoded default chain?

packages/scaffold-utils/src/SolanaHelpersUtils.ts
```diff
@@ -0,0 +1,71 @@
...
+
+    const selectedChain = chains.find(
+      chain => chain.chainId === chainId && chain.name === chainName
+    )
+
+    if (selectedChain) {
+      return selectedChain
+    }
+
+    return chains[0]
```
tomiir, line 27: same

packages/scaffold-utils/src/SolanaStoreUtil.ts
```diff
@@ -0,0 +1,122 @@
...
+import { Connection } from '@solana/web3.js'
+
+import UniversalProvider from '@walletconnect/universal-provider'
+
+import type { Chain, CombinedProvider, Provider } from './SolanaTypesUtil.js'
+
+type StateKey = keyof SolStoreUtilState
+
+export interface SolStoreUtilState {
+  projectId: string
```
tomiir, line 12: is this required? projectId is available in `OptionsController`

packages/scaffold-utils/src/SolanaStoreUtil.ts
```diff
@@ -0,0 +1,122 @@
...
+  provider?: Provider | CombinedProvider | UniversalProvider
+  providerType?: 'walletConnect' | `injected_${string}` | 'coinbaseWallet' | 'eip6963' | 'w3mEmail'
```
tomiir, line 14: don't think email is supported on solana atm

packages/scaffold-utils/src/SolanaTypesUtil.ts
```diff
@@ -0,0 +1,335 @@
...
+  data: string
+  programIdIndex: number
+}
+
+export interface TransactionElement {
+  meta: Meta
+  transaction: TransactionTransaction
+}
+
+export interface TransactionTransaction {
```
tomiir, line 118: nit: I think `TransactionData` would be a better name

packages/scaffold-utils/src/SolanaTypesUtil.ts
```diff
@@ -0,0 +1,335 @@
...
+  programId: string
+  data: string
+  keys: {
+    isSigner: boolean
+    isWritable: boolean
+    pubkey: string
+  }[]
+}
+
+interface VersionedInstractionRq {
```
tomiir, line 190: what does `Rq` stand for? Request? I'd leave full name even if long

packages/scaffold/src/views/w3m-networks-view/index.ts
```diff
@@ -97,6 +97,9 @@ export class W3mNetworksView extends LitElement {
         RouterController.push('SwitchNetwork', { ...data, network })
       }
     } else if (!isConnected) {
+      if (network.id.toLocaleLowerCase().includes('solana')) {
+        await NetworkController.switchActiveNetwork(network)
+      }
       NetworkController.setCaipNetwork(network)
```
tomiir, line 103: @svenvoskamp could we just use `switchActiveNetwork`  here? wyt? 
I think this might be some current redundant code.

packages/solana/exports/react.tsx
```diff
@@ -0,0 +1,70 @@
...
+  return modal
+}
+
+// -- Hooks -------------------------------------------------------------------
+export function useWeb3ModalProvider() {
+  const { provider, providerType, connection } = useSnapshot(SolStoreUtil.state)
+
+  const walletProvider = provider as Provider
+
+  const walletProviderType = providerType
```
tomiir, line 33: nit: I'd remove intermediate variables 

```
 return {
    walletProvider: provider as Provider,
    walletProviderType: providerType,
    connection
  }

File: packages/solana/package.json [hunk 12]
```diff
@@ -0,0 +1,98 @@
...
+    "react": {
+      "optional": true
+    },
+    "react-dom": {
+      "optional": true
+    },
+    "@web3modal/siwe": {
+      "optional": true
+    }
+  },
```
tomiir, line 74: @glitch-txs I think we could remove these correct?

packages/solana/src/client.ts
```diff
@@ -0,0 +1,482 @@
...
+        })
+        const address = await this.WalletConnectConnector.connect()
+        this.setWalletConnectProvider(address)
+      },
+
+      connectExternal: async ({ id }) => {
+        const adapter =
+          id.toLocaleLowerCase() === 'Trust'
+            ? 'trustWallet'
+            : (id.toLocaleLowerCase() as AdapterKey)
```
tomiir, line 107: do we have any control over this data? If so, could we add a `name` property to avoid manually mapping special cases?

packages/solana/src/client.ts
```diff
@@ -0,0 +1,482 @@
...
+        await this.walletAdapters[adapter].connect()
+        const address = this.walletAdapters[adapter].publicKey?.toString()
+        this.setInjectedProvider(
+          this.walletAdapters[adapter] as unknown as Provider,
+          adapter,
+          address
+        )
+      },
+
+      checkInstalled(ids) {
```
tomiir, line 118: what would the purpose of `ids` be? Don't see it being used other than defaulting to window.solana

packages/solana/src/client.ts
```diff
@@ -0,0 +1,482 @@
...
+
+    SolStoreUtil.subscribeKey('caipChainId', () => {
+      this.syncNetwork(chainImages)
+    })
+
+    if (typeof window === 'object') {
+      setTimeout(() => {
+        this.checkActiveProviders()
+        this.syncConnectors()
+      }, 500)
```
tomiir, line 213: any specific reason for this timeout? I'd add a comment on why if it's required.

packages/solana/src/client.ts
```diff
@@ -0,0 +1,482 @@
...
+    const walletId = localStorage.getItem(SolConstantsUtil.WALLET_ID)
+
+    try {
+      if (walletId === ConstantsUtil.WALLET_CONNECT_CONNECTOR_ID) {
+        await this.WalletConnectConnector.connect(true)
+        const provider = await this.WalletConnectConnector.getProvider()
+        const accounts = await provider.enable()
+        this.setWalletConnectProvider(accounts[0])
+      } else {
+        const wallet = walletId?.split('_')[1] as AdapterKey
```
tomiir, line 245: do walletIds ALWAYS have `_` ? I'd add optional chaining just in case to prevent this to crash in case it doesn't

packages/solana/src/client.ts
```diff
@@ -0,0 +1,482 @@
...
+    if (this.chains) {
+      if (chain) {
+        SolStoreUtil.setCaipChainId(`${chain.name}:${chain.chainId}`)
+        SolStoreUtil.setCurrentChain(chain)
+        localStorage.setItem(SolConstantsUtil.CAIP_CHAIN_ID, `${chain.name}:${chain.chainId}`)
+        if (providerType?.includes(ConstantsUtil.INJECTED_CONNECTOR_ID)) {
+          const wallet =
+            providerType.split('_')[1] === 'Trust'
+              ? 'trustWallet'
+              : (providerType.split('_')[1] as AdapterKey)
```
tomiir, line 382: same as before, if not possible to manipulate the name, would add some util function to prevent duplicated code.

packages/solana/src/client.ts
```diff
@@ -0,0 +1,482 @@
...
+          await this.syncAccount()
+
+          return
+        }
+        if (providerType === ConstantsUtil.WALLET_CONNECT_CONNECTOR_ID) {
+          const universalProvider = await this.WalletConnectConnector.getProvider()
+
+          const namespaces = this.WalletConnectConnector.generateNamespaces(chain.chainId)
+          SolStoreUtil.setConnection(
+            new Connection(chain.rpcUrl ?? 'https://api.devnet.solana.com', 'recent')
```
tomiir, line 396: is it ok to default to devnet?

packages/solana/src/connectors/WalletConnectConnector.ts
```diff
@@ -0,0 +1,254 @@
...
+  currency: 'SOL',
+  explorerUrl: 'https://solscan.io',
+  rpcUrl: 'https://api.testnet.solana.com'
+}
+
+export const solanaDevnet = {
+  chainId: '8E9rvCKLFQia2Y35HXjjpWzj8weVo44K',
+  name: 'Solana Devnet',
+  currency: 'SOL',
+  explorerUrl: 'https://solscan.io',
```
tomiir, line 32: out of ignorance, are explorer urls for testnet and devnet the same as mainnet?

packages/solana/src/connectors/WalletConnectConnector.ts
```diff
@@ -0,0 +1,254 @@
...
+      pubkey: address
+    })
+    const { signature } = signedMessage
+
+    return signature
+  }
+
+  public async signVersionedTransaction(transaction: VersionedTransaction) {
+    const transactionParams = {
+      feePayer: new PublicKey(SolStoreUtil.state.address ?? '').toBase58(),
```
tomiir, line 131: What happens when doing `PublicKey('')`? Will it throw / show weird behaviors?

packages/solana/src/connectors/universalProvider.ts
```diff
@@ -0,0 +1,78 @@
...
+     *
+     * // Subscribe to session update
+     *UniversalProviderFactory.provider.on(
+     *'session_update',
+     *({ topic, params }: { topic: string, params: unknown }) => {
+     *  console.log(topic, params)
+     *}
+     *)
+     *
+     */
```
tomiir, line 70: can we remove these or are these placeholders for upcoming actual implementations?

tomiir (changes requested):
Really good job! 
Really long PR too so a lot of comments. A mix of actual rfc and some questions given my lack of expertise on Solana. 
Nevertheless impressive feat 👏

.gitignore
```diff
@@ -2,11 +2,11 @@ dist
 node_modules
 .DS_Store
 .env
+.idea
 .next
 out
 lerna-debug.log
 .env.local
-.env
```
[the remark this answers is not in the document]
gooxpf: I removed duplication

packages/core/src/controllers/ApiController.ts
```diff
@@ -138,6 +138,16 @@ export const ApiController = {
     }
   },
 
+  getChains() {
+    const { caipNetwork } = NetworkController.state
```
enesozturk, line 142: Let's separate the preperation-execution-conclusion parts of the function body with a new line. This will help us to keep them in a good pattern to make them more readable.

```suggestion
    const { caipNetwork } = NetworkController.state

```

packages/scaffold-utils/src/SolanaConstantsUtil.ts
```diff
@@ -0,0 +1,28 @@
+import { PublicKey } from '@solana/web3.js'
+
+export const SolConstantsUtil = {
+  HASH_PREFIX: 'SPL Name Service',
+  NAME_OFFERS_ID: new PublicKey('85iDfUvr3HJyLM2zcq5BXSiDvUWfw6cSE1FfNBo8Ap29'),
```
enesozturk, line 5: What are these keys? How do we create those? Should we move them to environment variables? Preferably a JSDoc for `SolConstantsUtil` would be nice with short descriptions about the keys.

packages/scaffold-utils/src/SolanaHelpersUtils.ts
```diff
@@ -0,0 +1,71 @@
+import { PresetsUtil } from './PresetsUtil.js'
+
+import type { CaipNetwork } from '@web3modal/core'
+import type { Chain, Provider } from './SolanaTypesUtil.js'
+
+export const SolHelpersUtil = {
+  getChain(chains: Chain[], chainId: string | null) {
+    const chain = chains.find(lChain => lChain.chainId === chainId)
```
enesozturk, line 8: Let's separate the preperation-execution-conclusion parts of the function body with a new line. This will help us to keep them in a good pattern to make them more readable.

```suggestion
    const chain = chains.find(lChain => lChain.chainId === chainId)
    
```

packages/scaffold-utils/src/SolanaHelpersUtils.ts
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    return {
+      id: `${chain.name}:${chain.chainId}`,
+      name: chain.name,
+      imageId: PresetsUtil.EIP155NetworkImageIds[chain.chainId]
+    } as CaipNetwork
+  },
+  hexStringToNumber(value: string) {
+    const string = value.startsWith('0x') ? value.slice(2) : value
```
enesozturk, line 41: We should avoid using reserved keywords of the language for the variable names and they should be more descriptive. This will help us to avoid potential issues while bundling our code and improve our code's maintainability.

```suggestion
    const hexString = value.startsWith('0x') ? value.slice(2) : value
    const decimalValue = parseInt(string, 16)
    
    return decimalValue
```

packages/scaffold-utils/src/SolanaHelpersUtils.ts
```diff
@@ -0,0 +1,71 @@
...
+    const number = parseInt(string, 16)
+
+    return number
+  },
```
enesozturk, line 45: Let's separate functions with new lines as we do in all other utils and controllers.
```suggestion
  },
  
```

packages/scaffold-utils/src/SolanaTypesUtil.ts
```diff
@@ -0,0 +1,335 @@
...
+  data: string
+  programIdIndex: number
+  accountKeyIndexes: number[]
+}
+
+export interface RequestMethods {
```
enesozturk, line 196: Curious: do we manually create these types or copy over from somewhere? If we copy from somewhere, could you add a source to there within a JSDoc on top?

packages/solana/src/connectors/walletAdapters.ts
```diff
@@ -0,0 +1,39 @@
...
+  supportedWallets.forEach(wallet => {
+    if (window[wallet as keyof Window]) {
+      w3mConnectors.push({
+        id: adapters[wallet].name,
+        type: 'ANNOUNCED',
+        imageUrl: adapters[wallet].icon,
+        name: adapters[wallet].name,
+        provider: adapters[wallet],
+        info: {
+          rdns: `app.${wallet}`
```
enesozturk, line 34: Template literals for constructing values like `app.${wallet}` dynamically can raise concerns, especially in contexts where the input (in this case, `wallet`) is user-controlled or not strictly validated.

Template literals do not inherently validate the input. Using property access (`adapters[wallet].rdns`) ensures that only predefined keys in the adapters object can be accessed.

Direct property access also makes the code more readable and easier to understand. It's clear that you are accessing a specific property of an object, which makes the codebase easier to maintain and debug.

```suggestion
          rdns: adapters[wallet].rdns
```

packages/solana/src/utils/hash.ts
```diff
@@ -0,0 +1,42 @@
...
+  // @ts-expect-error this
+  window.getHashedName = getHashedName
+}
+
+export async function getNameAccountKey(
+  hashed_name: Buffer,
+  nameClass?: PublicKey,
+  nameParent?: PublicKey
+): Promise<PublicKey> {
+  const seeds = [hashed_name]
```
enesozturk, line 23: Let's separate the preperation-execution-conclusion parts of the function body with a new line. This will help us to keep them in a good pattern to make them more readable.

```suggestion
  const seeds = [hashed_name]
  
```

packages/ui/src/composites/wui-account-button/index.ts
```diff
@@ -74,7 +75,7 @@ export class WuiAccountButton extends LitElement {
     }
     if (this.balance) {
       const networkElement = this.networkSrc
-        ? html`<wui-image src=${this.networkSrc}></wui-image>`
+        ? html` <wui-image src=${this.networkSrc}></wui-image>`
```
enesozturk, line 78: This is possibly a result of a Prettier formatting process. Let's remove and check if it's happening again. I generally prefer avoiding empty spaces in renders. 

```suggestion
        ? html`<wui-image src=${this.networkSrc}></wui-image>`
```

packages/ui/src/composites/wui-account-button/index.ts
```diff
@@ -86,11 +87,11 @@ export class WuiAccountButton extends LitElement {
 
       return html`
         ${networkElement}
-        <wui-text variant="paragraph-600" color="inherit"> ${this.balance} </wui-text>
+        <wui-text variant="paragraph-600" color="inherit"> ${this.balance}</wui-text>
       `
     }
 
-    return null
+    return html` <wui-text variant="paragraph-600" color="inherit">No balance info</wui-text>`
```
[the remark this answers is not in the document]
enesozturk: Yes, let's double-check this. If we need a design improvement or collaboration, we could let our designers know and proceed with the UI changes when approved.

packages/solana/package.json
glitch-txs: yes, SIWE should be removed

packages/solana/src/client.ts
```diff
@@ -0,0 +1,482 @@
...
+    if (chain) {
+      SolStoreUtil.setCurrentChain(chain)
+      SolStoreUtil.setCaipChainId(`${chain.name}:${chain.chainId}`)
+    }
+    this.syncNetwork(chainImages)
+
+    this.walletAdapters = createWalletAdapters()
+    this.WalletConnectConnector = new WalletConnectConnector({
+      relayerRegion: 'wss://relay.walletconnect.com',
+      metadata: {
```
glitch-txs, line 188: metadata should be an argument from devs, url is required in it.

packages/solana/src/connectors/WalletConnectConnector.ts
```diff
@@ -0,0 +1,254 @@
...
+    return new Promise<string>((resolve, reject) => {
+      provider.on('display_uri', (uri: string) => {
+        if (!(this.qrcode && !useURI)) {
+          resolve(uri)
+        }
+      })
+      provider
+        .connect({
+          pairingTopic: undefined,
+          namespaces: solanaNamespace
```
glitch-txs, line 235:
```suggestion
          optionalNamespaces: solanaNamespace
```
let's use optionalNamespaces instead, should be the default

packages/solana/src/client.ts
```diff
@@ -0,0 +1,482 @@
...
+        description: 'Solana in Wallet Connect',
+        name: 'Wallet Connect',
+        icons: ['https://avatars.githubusercontent.com/u/37784886'],
+        url: 'https://web3modal.com'
+      },
+      autoconnect: false,
```
glitch-txs, line 194: why is this `false`?

packages/scaffold-utils/src/SolanaStoreUtil.ts
```diff
@@ -0,0 +1,122 @@
...
+
+  reset() {
+    state.provider = undefined
+    state.address = undefined
+    state.chainId = undefined
+    state.providerType = undefined
+    state.isConnected = false
+    state.error = undefined
+  }
+}
```
glitch-txs, line 122: this file seems like a duplicate from the core package, can we remove it?

packages/core/src/controllers/ApiController.ts
glitch-txs, line 142: `caipNetwork` should only refer to a current connected chain and we should not rely on it for static configuration, we should use `requestedCaipNetworks` instead

packages/core/src/controllers/ApiController.ts
```diff
@@ -138,6 +138,16 @@ export const ApiController = {
...
+    if (CoreHelperUtil.checkCaipNetwork(caipNetwork, 'solana')) {
```
glitch-txs, line 143: this seems to be unnecessary, doesn't it return the same string as `caipNetwork.id`?

packages/solana/src/connectors/websocket.ts
```diff
@@ -0,0 +1,18 @@
+export async function waitForOpenConnection(socket: WebSocket) {
```
glitch-txs, line 1: this seems to be more for utils instead of the connectors folder

glitch-txs (changes requested):
Great work! Let's try please moving as much logic as possible to the Solana package to avoid bundle size increase for EVM only packages.

Eventually we'd like to be able to support both types of chains at the same time: EVM chains + Solana

packages/scaffold-utils/src/SolanaConstantsUtil.ts
```diff
@@ -0,0 +1,28 @@
+import { PublicKey } from '@solana/web3.js'
```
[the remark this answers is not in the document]
gooxpf: Moved to solana/utils/scaffold

packages/core/src/controllers/NetworkController.ts
gooxpf: Without refetching here, when disconnected user change networks, it shows wrong wallets count in the modal all wallets.

.vscode/settings.json
```diff
@@ -1,25 +0,0 @@
...
-    "**/.turbo": true,
-    "**/playwright-report": true,
-    "**/coverage": true
-  },
-  "javascript.updateImportsOnFileMove.enabled": "always",
-  "typescript.tsdk": "node_modules/typescript/lib",
-  "typescript.updateImportsOnFileMove.enabled": "always",
-  "css.lint.unknownAtRules": "ignore",
-  "css.validate": true
-}
```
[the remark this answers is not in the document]
glitch-txs: Protocol reasons, let's keep this branch for Solana related changes only. Let's keep all the monorepo config as it is please.

packages/ui/src/composites/wui-account-button/index.ts
```diff
@@ -49,7 +49,8 @@ export class WuiAccountButton extends LitElement {
             address=${this.address}
           ></wui-avatar>
           <wui-text variant="paragraph-600" color="inherit">
-            ${UiHelperUtil.getTruncateString({
+            ${this.address &&
```
enesozturk, line 52: Let's not use the `&&` (AND) operator and instead use the `?` (ternary) operator. The AND operator might render the left-hand side values to the DOM, when it's `truty` but looks like "falsy". 
See: https://kentcdodds.com/blog/use-ternaries-rather-than-and-and-in-jsx

Even though we could be sure about the type definition of the `this.address`, the developer shouldn't be responsible for checking if it's not causing this issue or not.

packages/solana/src/utils/scaffold/SolanaConstantsUtil.ts
```diff
@@ -0,0 +1,38 @@
...
+   * Mainnet program ID
+   */
+  NAME_TOKENIZER_ID: new PublicKey('nftD3vbNkNqfj2Sd3HZwbpw4BxxKWr4AjGb9X38JeZk'),
+  MINT_PREFIX: Buffer.from('tokenized_name'),
+  WALLET_ID: '@w3m/solana_wallet',
+  CAIP_CHAIN_ID: '@w3m/solana_caip_chain',
+  ERROR_CODE_UNRECOGNIZED_CHAIN_ID: 4902,
+  ERROR_CODE_DEFAULT: 5000,
+  DEFAULT_CHAIN: {
+    chainId: '5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp',
```
enesozturk, line 32: Are these values up to date?

packages/solana/src/utils/scaffold/SolanaConstantsUtil.ts
```diff
@@ -0,0 +1,38 @@
...
+  /**
+   * The reverse look up class
+   */
+  REVERSE_LOOKUP_CLASS: new PublicKey('33m47vH6Eav6jr5Ry86XjhRft2jRBLDnDgPSHoquXi2Z'),
+  TOKEN_PROGRAM_ID: new PublicKey('TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA'),
+  /**
+   * Mainnet program ID
+   */
+  NAME_TOKENIZER_ID: new PublicKey('nftD3vbNkNqfj2Sd3HZwbpw4BxxKWr4AjGb9X38JeZk'),
+  MINT_PREFIX: Buffer.from('tokenized_name'),
```
enesozturk, line 26: Where this value is used? Should we update is something more understandable and seems related to Solana?

packages/solana/src/utils/clusterFactory.ts
```diff
@@ -0,0 +1,104 @@
...
+      id: number
+      result: number
+      params: {
+        subscription: number
+        result: number
+      }
+    }
+
+    /*
+     * If request is a subscribtion init notification
```
enesozturk, line 38:
```suggestion
     * If the request is a subscription init notification
```

packages/solana/src/connectors/walletAdapters.ts
```diff
@@ -0,0 +1,39 @@
+import { PhantomWalletAdapter } from '@solana/wallet-adapter-phantom'
```
enesozturk, line 1: In this folder (`/connectors`), this has `camelCase` while some others have `KebabCase`, let's stick to same naming convention (in case there is something different with those, but as I see both contain a class)

packages/solana/src/connectors/WalletConnectConnector.ts
enesozturk: @valentyn-solana If this `''` check is in responsible of the end user, we should definitely disable them from running unexpected situations. There might be a check inside the function like the following:

```js
  public async signVersionedTransaction(transaction: VersionedTransaction) {
     if(!SolStoreUtil.state.address) { ... take any action or throw error } 
   ...
   }
```

This has better readability because the engineers know what happens if the address is empty.

packages/solana/src/connectors/WalletConnectConnector.ts
```diff
@@ -0,0 +1,235 @@
...
+
+  public static readonly connectorName = 'walletconnect'
+
+  public async disconnect() {
+    const provider = await UniversalProviderFactory.getProvider()
+
+    try {
+      await provider.disconnect()
+    } finally {
+      // (TODO update typing for provider)
```
enesozturk, line 77: @valentyn-solana @gooxpf  Todo here

packages/solana/src/connectors/BaseConnector.ts
```diff
@@ -0,0 +1,392 @@
...
+    }).compileToV0Message()
+
+    // Make a versioned transaction
+    const transactionV0 = new VersionedTransaction(messageV0)
+
+    return transactionV0
+  }
+
+  /*
+   * Public async sendTransaction(encodedTransaction: string) {
```
enesozturk, line 163: @gooxpf @valentyn-solana Commented out, should we keep it?

packages/solana/src/connectors/BaseConnector.ts
```diff
@@ -0,0 +1,392 @@
...
+  }
+
+  public async watchTransaction(
+    transactionSignature: string,
+    callback: (params: Transaction | number) => void
+  ) {
+    return this.subscribeToCluster('signatureSubscribe', [transactionSignature], callback)
+  }
+
+  public async getBalance(requestedAddress: string, currency: 'lamports' | 'sol' = 'sol') {
```
enesozturk, line 185:
```suggestion
// on top, in the type definitions group
type Currency = 'lamports' | 'sol'
..
  public async getBalance(requestedAddress: string, currency: Currency = 'sol') {
```

packages/solana/src/connectors/BaseConnector.ts
```diff
@@ -0,0 +1,392 @@
...
+    try {
+      const address = requestedAddress ?? SolStoreUtil.state.address
+      const balance = await this.requestCluster('getBalance', [
+        address,
+        { commitment: 'processed' }
+      ])
+      const formatted =
+        currency === 'lamports'
+          ? `${balance?.value || 0} lamports`
+          : `${(balance?.value || 0) / 1000000000} sol`
```
enesozturk, line 195: Why this is `1000000000` and would it be better to define this variable on the top with a meaningful name?

```
// or something else name
const BALANCE_VALUE_DECIMAL_DIVIDER = 1000000000
```

apps/laboratory/src/components/Solana/SolanaSendTransactionTest.tsx
```diff
@@ -0,0 +1,163 @@
...
+  return (
+    <Stack direction={['column', 'column', 'row']}>
+      <Button
+        data-test-id="sign-transaction-button"
+        onClick={onSendTransaction}
+        isDisabled={loading}
+      >
+        Sign and Send Transaction
+      </Button>
+      {supportV0Transactions && (
```
enesozturk, line 145: While this will work as expected, let's stick to using ternary operator as I explained in the previous comment

enesozturk (dismissed):
LETS GOOOO 🚀 Left a few comments, I think they are not blockers but good to have a check and make updates. Appreciate the big work @gooxpf @valentyn-solana 🎉

packages/solana/src/connectors/BaseConnector.ts
valentyn-solana: Leftover, will remove

packages/solana/src/connectors/BaseConnector.ts
```diff
@@ -0,0 +1,392 @@
...
+    const hashedReverseLookup = getHashedName(address)
+    const reverseLookupAccount = await getNameAccountKey(
+      hashedReverseLookup,
+      SolConstantsUtil.REVERSE_LOOKUP_CLASS
+    )
+
+    const account = await this.getAccount(reverseLookupAccount.toBase58(), 'base64')
+
+    if (account) {
+      const dataBuffer = Buffer.from(String(account.data[0]), 'base64')
```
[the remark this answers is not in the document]
valentyn-solana: Since our functionality doesn't utilize getSolDomainsFromPublicKey and performReverseLookup we can get rid of them

packages/solana/src/utils/scaffold/SolanaConstantsUtil.ts
gooxpf: Yeah, its updated id.

valentyn-solana (dismissed):

packages/solana/src/connectors/walletAdapters.ts
enesozturk: @valentyn-solana Sorry I meant about the file names 🙏

enesozturk (approved):

packages/solana/src/connectors/walletAdapters.ts
```diff
@@ -0,0 +1,39 @@
...
+  supportedWallets.forEach(wallet => {
+    if (window[wallet as keyof Window]) {
+      w3mConnectors.push({
+        id: adapters[wallet].name,
+        type: 'ANNOUNCED',
+        imageUrl: adapters[wallet].icon,
+        name: adapters[wallet].name,
+        provider: adapters[wallet],
+        info: {
+          rdns: adapters[wallet].publicKey?.toString()
```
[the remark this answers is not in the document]
tomiir: hm, rdns is tied to the provider rather than the address itself. 
Technically should be a name for which you can do a reverse lookup for info on it but if it doesn't really exist in solana 🤷 

I'd use the wallet name or similar instead of the address though, since it's more inline with evm providers.

cc @glitch-txs what do you think?

apps/laboratory/src/utils/DataUtil.ts
```diff
@@ -22,35 +22,45 @@ export const wagmiSdkOptions = [
...
     title: 'Default',
     link: '/library/wagmi',
-    description: 'Basic configuration using wagmi as a driving library'
+    description: 'Basic configuration usign wagmi as a driving library'
   },
   {
     title: 'SIWE',
     link: '/library/wagmi-siwe/',
-    description: 'Configuration using wagmi and implementing sign in with ethereum'
+    description: 'Configuration usign wagmi and implementing sign in with ethereum'
```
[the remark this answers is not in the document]
gooxpf: Yeah,   I changed it for solana, forget about wagmi )

packages/solana/src/client.ts
valentyn-solana: Removed, cause it's redundant

valentyn-solana (dismissed):

packages/solana/src/client.ts
valentyn-solana: yeah, all injected wallets have name in id (like injected_phantom, injected_solflare) to distinguish them

packages/solana/src/client.ts
```diff
@@ -0,0 +1,482 @@
...
+    } else if (!isConnected && this.hasSyncedConnectedAccount) {
+      this.resetWcConnection()
+      this.resetNetwork()
+    }
+  }
+
+  private syncProfile(_address: string) {
+    this.setProfileName(null)
+    this.setProfileImage(null)
+  }
```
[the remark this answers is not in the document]
valentyn-solana: I think we will utilize it when adding socials sign, it has same method in ethers, so it is TODO to receive some profile info later

valentyn-solana (approved):

packages/core/src/controllers/NetworkController.ts
enesozturk: @gooxpf In this case, shouldn't it be refetching when switch from Solana -> Ethereum? Thise code condition, it'll only refetch when switch from X -> Solana right?

packages/scaffold/src/views/w3m-networks-view/index.ts
```diff
@@ -97,6 +97,9 @@ export class W3mNetworksView extends LitElement {
         RouterController.push('SwitchNetwork', { ...data, network })
       }
     } else if (!isConnected) {
+      if (network.id.toLocaleLowerCase().includes('solana')) {
```
enesozturk, line 100: Can we use `CoreHelperUtil.checkCaipNetwork(network, 'solana')` here? @gooxpf

packages/scaffold/src/views/w3m-networks-view/index.ts
svenvoskamp: We don't want library-specific code within our scaffold package (solana check here). What is the problem you are running into? @valentyn-solana @gooxpf

svenvoskamp: You shouldn't need `switchActiveNetwork` here, because you're not connected yet.

packages/core/src/controllers/NetworkController.ts
valentyn-solana: Yes, moved it out to solana listener when chain is changed, also fixed w3m-connect-view, added subscription to  wallets count

valentyn-solana (approved):

packages/scaffold-utils/src/SolanaStoreUtil.ts
phoenixVS: Yes, moved it out and using OptionsController now

phoenixVS (dismissed):

enesozturk (approved):

enesozturk (dismissed):

glitch-txs (approved):

tomiir (approved):

enesozturk (approved):

Issue #1729, opened 2024-01-09
[bug] Unsafe property access on `retryButton`


bkrem:
### Link to minimal reproducible example

https://walletconnect.sentry.io/issues/4724380976/events/004a68a322d84893904c63f7658c359a/

### Summary

* The `animate` method on the `retryButton` element seems to be called optimistically without proper access guards in place
* This leads to continuous Sentry errors across multiple of our projects consuming Web3Modal (and presumably externally too)
* Full stack trace attached below along with screenshot of exact lines affected. See also linked Sentry issue.

```
TypeError: Cannot read properties of null (reading 'animate')
  at W3mConnectingWidget.onShowRetry(./node_modules/@web3modal/scaffold/dist/esm/src/utils/w3m-connecting-widget/index.js:128:1)
  at W3mConnectingWidget.render(./node_modules/@web3modal/scaffold/dist/esm/src/utils/w3m-connecting-widget/index.js:58:1)
  at lit_element_s.update(./node_modules/lit-element/lit-element.js:6:222)
  at b.performUpdate(./node_modules/lit-element/node_modules/@lit/reactive-element/reactive-element.js:6:4810)
  at b.scheduleUpdate(./node_modules/lit-element/node_modules/@lit/reactive-element/reactive-element.js:6:4426)
  at b._$EP(./node_modules/lit-element/node_modules/@lit/reactive-element/reactive-element.js:6:4333)
```

<img width="936" alt="Screenshot 2024-01-09 at 10 37 15" src="https://github.com/WalletConnect/web3modal/assets/8663099/642ad1bf-4f5b-4426-9254-b78b4b4d92ef">

### List of related npm package versions

* `@web3modal/scaffold

CUManiar:
Fixed  with MR: #1734

bkrem:
> Fixed with MR: #1734

Thank you for patching this @CUManiar! 💯

PR #1734, opened 2024-01-10
Retry button `null` fix

CUManiar:
# Breaking Changes
Unnecessarily tries to `animate` on the retryButton whereas the button doesn't exist

N/A

# Changes
Adding optional chaining in order to avoid `.animate` call over `retryButton`. If it exist then it'll work accordingly otherwise it won't be bothered


CUManiar:
fix #1729

tomiir (approved):

PR #1736, opened 2024-01-10
Fix/w3m button loading

svenvoskamp:
# Changes

Prevent use case that you can open modal while button is in loading state


tomiir (approved):
lgtm. 
Also, should we diable the button as well when loading?

PR #1737, opened 2024-01-10
feat(ui-test): support vercel preview

arein:
Allows the wallet page to be in Vercel preview.

# Breaking Changes

N/A

# Changes

- feat(ui-test): support vercel preview
- fix:
- chore:

# Associated Issues

closes #...


apps/laboratory/tests/shared/pages/WalletPage.ts
```diff
@@ -7,18 +7,24 @@ export class WalletPage {
...
   async load() {
     await this.page.goto(this.baseURL)
   }
 
   async connect() {
+    const isVercelPreview = (await this.vercelPreview.count()) > 0
+    if (isVercelPreview) {
+      // eslint-disable-next-line  @typescript-eslint/no-explicit-any
+      await this.vercelPreview.evaluate((iframe: any) => iframe.remove())
+    }
```
tomiir, line 26: seems like repeated code, could this be abstracted to a util/helper ?
also, doesn't `HTMLIFrameElement` work?
something like 

```ts
export const removeVercelPreview = async (page: Page) => {
    const vercelPreviewCount = await this.vercelPreview.count()
    if (vercelPreviewCount > 0) {
      await this.vercelPreview.evaluate((iframe: HTMLIFrameElement) => iframe.remove())
    }
}```

arein: I'm copying it twice in test code. So my rule is you can copy twice then you need to refactor. Hope this is cool since it's tests and they will need some love anyways when web3wallet fixes are live.

Thanks on `HTMLIFrameElement`. Added!

tomiir (approved):

PR #1739, opened 2024-01-10
feat: lab contract write example

tomiir:
# Breaking Changes

N/A

# Changes

- feat: adds contract write tests by calling a Donut Vending Machine contract deployed on sepolia 
- fix:
- chore:


svenvoskamp (approved):
Approved if merge conflicts are fixed

.vscode/settings.json
```diff
@@ -10,9 +10,9 @@
   "files.exclude": {
     "**/.DS_Store": true,
     "**/.git": true,
-    "**/node_modules": true,
-    "**/out": true,
-    "**/dist": true,
+    "**/node_modules": false,
+    "**/out": false,
+    "**/dist": false,
```
0xAsimetriq, line 15: Do we need to update this, or can this be kept as local vscode override? Did you need to access these folders (they should never be tweaked as they are products of build).

0xAsimetriq (approved):

.vscode/settings.json
tomiir: added .vscode folder to gitignore since it's not supposed to change 🙏

arein (approved):
we need to spin up a repo for the donut contract later and CI etc

PR #1754, opened 2024-01-15
test: web wallet deeplink

ganchoradkov:


arein (dismissed):

tomiir (dismissed):

arein:
Screenshot 2024-01-15 at 9 01 35 AM

@ganchoradkov seems like there's a small edge case

ganchoradkov:
@arein it's not edge case, just need to wait a bit more the request modal to show up. I will still add a loader for it though

arein:
> @arein it's not edge case, just need to wait a bit more the request modal to show up. I will still add a loader for it though

Can you actually try this? When I sign it's all successful but then goes back into that state

PR #1765, opened 2024-01-16
feat: enableOnramp flag

tomiir:
# Breaking Changes

N/A

# Changes

- feat: adds onrampEnabled flag on Web3Modal and use it to display/hide onramp options in modal.
- fix:
- chore:


packages/scaffold/src/views/w3m-account-view/index.ts
```diff
@@ -162,6 +155,25 @@ export class W3mAccountView extends LitElement {
   }
 
   // -- Private ------------------------------------------- //
+  private onrampTemplate() {
+    const { enabled } = OnRampController.state
```
enesozturk, line 159: Maybe new lines between variables and return statements? 
```suggestion
    const { enabled } = OnRampController.state
    
```

enesozturk (approved):

packages/core/src/controllers/OnRampController.ts
```diff
@@ -16,6 +16,7 @@ export interface OnRampControllerState {
   providers: OnRampProvider[]
   selectedProvider: OnRampProvider | null
   error: string | null
+  enabled: boolean
```
xzilja, line 19: Should this perhaps live in `OptionsController` instead? (to keep all feature flags in one place)

PR #1775, opened 2024-01-18
feat: parallelizes ui tests

ganchoradkov:
# Breaking Changes

N/A

# Changes

- feat: updated react-wallet tests to use deeplinks allowing for parallelization

### DIFF
--- a/apps/laboratory/playwright.config.ts
+++ b/apps/laboratory/playwright.config.ts
@@ -10,7 +10,7 @@ export default defineConfig<ModalFixture>({
 
   fullyParallel: true,
   retries: 0,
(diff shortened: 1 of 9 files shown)


arein (changes requested):
LGTM  - epic!!!!! - but can you keep the old test and rename it to `qrCode.connect.ts` and then do some `test.skip()` such that it only runs on one browser/library so we don't have the race conditions.

xzilja (commented):
Don't see anything that jumps out so looks good, but actual test is failing atm.

lukaisailovic:
Closed by #1936

PR #1777, opened 2024-01-19
fix: wagmi 2 reconnect

tomiir:
# Breaking Changes

N/A

# Changes

- feat: adds `reconnect` flag
- fix: wagmi reconnection on vanilla and vue
- chore:

### DIFF
--- a/apps/laboratory/src/pages/library/wagmi-email.tsx
+++ b/apps/laboratory/src/pages/library/wagmi-email.tsx
@@ -1,6 +1,5 @@
 import { createWeb3Modal, defaultWagmiConfig } from '@web3modal/wagmi/react'
 import { QueryClient, QueryClientProvider } from '@tanstack/react-query'
-import { useEffect, useState } from 'react'
 import { WagmiProvider } from 'wagmi'
 import { Web3ModalButtons } from '../../components/Web3ModalButtons'
 import { WagmiTests } from '../../components/Wagmi/WagmiTests'
@@ -14,6 +13,7 @@ export const wagmiConfig = defaultWagmiConfig({
   chains: WagmiConstantsUtil.chains,
   projectId: ConstantsUtil.ProjectId,
   metadata: ConstantsUtil.Metadata,
+  ssr: true,
   enableEmail: true
 })
 
@@ -30,18 +30,12 @@ const modal = createWeb3Modal({
 ThemeStore.setModal(modal)
 
 export default function Wagmi() {
-  const [ready, setReady] = useState(false)
-
-  useEffect(() => {
-    setReady(true)
-  }, [])
-
-  return ready ? (
+  return (
     <WagmiProvider config={wagmiConfig}>
       <QueryClientProvider client={queryClient}>
         <Web3ModalButtons />
         <WagmiTests />
       </QueryClientProvider>
     </WagmiProvider>
-  ) : null
+  )
 }
--- a/examples/html-wagmi/src/main.js
+++ b/examples/html-wagmi/src/main.js
@@ -21,7 +21,13 @@ const wagmiConfig = defaultWagmiConfig({
 })
 
 // 3. Create modal
-const modal = createWeb3Modal({ wagmiConfig, projectId, chains, themeMode: 'light' })
+const modal = createWeb3Modal({
+  wagmiConfig,
+  projectId,
+  chains,
+  themeMode: 'light',
+  reconnect: true
+})
 
 // 4. Trigger modal programaticaly
 const openConnectModalBtn = document.getElementById('open-connect-modal')
--- a/examples/vue-wagmi/src/App.vue
+++ b/examples/vue-wagmi/src/App.vue
@@ -1,6 +1,5 @@
 <script lang="ts" setup>
 import { arbitrum, mainnet } from 'viem/chains'
-import type { Chain } from 'viem/chains'
 import {
   createWeb3Modal,
   defaultWagmiConfig,
@@ -17,7 +16,7 @@ if (!projectId) {
 }
 
 // 2. Create wagmiConfig
-const chains = [mainnet, arbitrum] as [Chain, ...Chain[]]
+const chains = [mainnet, arbitrum]
 const wagmiConfig = defaultWagmiConfig({
   chains,
   projectId,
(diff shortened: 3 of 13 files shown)


apps/laboratory/src/pages/library/wagmi-email.tsx
```diff
@@ -14,6 +13,7 @@ export const wagmiConfig = defaultWagmiConfig({
   chains: WagmiConstantsUtil.chains,
   projectId: ConstantsUtil.ProjectId,
   metadata: ConstantsUtil.Metadata,
+  ssr: true,
```
xzilja, line 16: Not a big fan of this, can we instead check internally for something like `window === 'undefined'` instead and pass that on to wagmi?

examples/vue-wagmi/src/App.vue
```diff
@@ -1,5 +1,5 @@
 <script lang="ts" setup>
-import { arbitrum, mainnet } from '@wagmi/core/chains'
+import { arbitrum, mainnet, type Chain } from 'viem/chains'
```
xzilja, line 2:
```suggestion
import { arbitrum, mainnet } from 'viem/chains'
import type { Chain } from 'viem/chains'
```

examples/html-wagmi/src/main.js
```diff
@@ -21,7 +21,13 @@ const wagmiConfig = defaultWagmiConfig({
 })
 
 // 3. Create modal
-const modal = createWeb3Modal({ wagmiConfig, projectId, chains, themeMode: 'light' })
+const modal = createWeb3Modal({
+  wagmiConfig,
+  projectId,
+  chains,
+  themeMode: 'light',
+  reconnect: true
```
xzilja, line 29: Similar to above, should reconnect be something we set internally and don't expose to user? Or perhaps this is a new flag we set to true by default and allow user to override? Thinking here is, how do we make it as simple and quick for user to get going without the need for all these extra options?

If they need to be enabled in ssr env and it has to be done manually, maybe we can combine it all under one flag?

examples/vue-wagmi/src/App.vue
```diff
@@ -16,7 +16,7 @@ if (!projectId) {
 }
 
 // 2. Create wagmiConfig
-const chains = [mainnet, arbitrum]
+const chains = [mainnet, arbitrum] as [Chain, ...Chain[]]
```
xzilja, line 19: This looks very odd, why do we need to do it this way? Will users of typescript also need to add this type assertion?

apps/laboratory/src/pages/library/wagmi-email.tsx
tomiir: I'm not against it, but maybe some user needs to disable this for some reason and wouldn't be able to. 
Maybe the flag can be on by default, and we check for `window` on our side as an extra check and they can turn it off completely

examples/html-wagmi/src/main.js
tomiir: I feel setting this to true by default is better. 

Regarding merging the `ssr` flag, I'm pretty sure this would be required so merging would make sense, but tagging @glitch-txs jic

examples/vue-wagmi/src/App.vue
tomiir: This shouldn't really be necessary as `[Chain, ...Chain[]]` should only mean "a Chain array with at least an element" but seems there's something weird with wagmi forcing us to cast it.

I changed the way we type things for now for a `Chain[]` and added a check to force users to provide at least 1 chain to the config.

tomiir: On this note, why do we have `chains` parameter on Web3ModalOptions? Shouldn't we use the ones in wagmi config?

examples/html-wagmi/src/main.js
glitch-txs: this is not related to SSR but rather revalidation of the storage data on page load, this needs to be set manually by calling `reconnect(config)` function if you're using `@wagmi/core`, for those using React this is already handled internally (it's true by default).
This helps devs that use other frameworks than React to make revalidation more intuitive, for those using SSR should get this flag set to `false` and handle the `reconnect` function themselves. e.g. revalidating on the app "`onMount`" or similar.

Wagmi uses this to optimistically hydrate the storage with cookies and after the app is mounted reconnect is called to avoid hydration mismatches.

apps/laboratory/src/pages/library/wagmi-email.tsx
glitch-txs: I think this is great for the examples since this is handled by Wagmi already, we just need to add the `cookies` feature imo

examples/html-wagmi/src/main.js
glitch-txs: > Similar to above, should reconnect be something we set internally and don't expose to user?

To clarify the above configuration, we added `...wagmiConfig` in defaultWagmiConfig so devs can add any extra Wagmi config's option they want, this is why you could add `ssr` or any other config

```ts
export function defaultWagmiConfig({
  projectId,
  chains,
  metadata,
  enableInjected,
  enableCoinbase,
  enableWalletConnect,
  enableEmail,
  enableEIP6963,
  ...wagmiConfig // <-- here
}: ConfigOptions): Config {
```

To add some context, reconnect is related to the revalidation of the storage data on page load (+ reconnecting), this needs to be set manually by calling `reconnect(config)` function if you're using `@wagmi/core`, for those using React this is already handled internally (it's true by default).

----------

Thinking a lot about this:

1.
> If they need to be enabled in ssr env and it has to be done manually, maybe we can combine it all under one flag?

If we do this and handle everything ourselves for those that are not using React then it might be a bit over-engineering imo, because this is something that Wagmi should handle instead.
There is for example a [Vue Wagmi library](https://github.com/unicape/use-wagmi) that might handle this internally as well, in this case devs won't be using tthis feature.

2. (rethinking this) It would be weird for a dev using SSR setting the `reconnect` flag to false, because they would still be able to reconnect if this is handled somewhere else.

I'm a bit confused now

should we leave this one and I can add the `reconnect` function manually in the documentation?

IMO let's leave this to wagmi and I can explain farther in the docs. (I just changed my mind xD)

Issue #1772, opened 2024-01-17
W3M shows buy failed when it was successful


tomiir:
Repro:
 - submit a buy on CB Pay
 - CB pay confirms buy
 - Return to W3M
 - It shows spinner for a long time
 - Eventually it says that the Tx failed

**CB Confirmation**
![Image](https://github.com/WalletConnect/web3modal/assets/84375331/3f3eb8f5-9100-429d-bce8-4cf4c02fbef3)

**W3M Issue**
![Image](https://github.com/WalletConnect/web3modal/assets/84375331/ec1a7070-aec1-4077-9d06-b509a091981c)
https://pay.coinbase.com/buy/select-asset?appId=bf18c88d-495a-463b-b249-0b9d3656cf5e&defaultNetwork=polygon&destinationWallets=%5B%7B%22address%22%3A%220x366b5EB733ee0FDB44fE9fBD96dc755C1FC67b75%22%2C%22blockchains%22%3A%5B%22ethereum%22%2C%22arbitrum%22%2C%22polygon%22%2C%22avalanche-c-chain%22%2C%22optimism%22%2C%22celo%22%5D%2C%22assets%22%3A%5B%22USDC%22%5D%7D%5D&partnerUserId=0x366b5EB733ee0FDB44fE9fBD96dc755C1FC67b75

PR #1779, opened 2024-01-19
fix: coinbase onramp activity infinite loading

tomiir:
# Breaking Changes

N/A

# Changes

- feat:
- fix: issue where loading transactions would never be refreshed and would be stuck in an inifnite load.
- chore:

# Associated Issues

closes [#1772](https://github.com/WalletConnect/web3modal/issues/1772)


packages/scaffold/src/views/w3m-onramp-activity-view/index.ts
```diff
@@ -104,6 +106,33 @@ export class W3mOnRampActivityView extends LitElement {
...
+    )
+
+    if (loadingTransactions.length === 0) {
+      clearTimeout(this.refetchTimeout)
+
+      return
+    }
+
+    // Wait 2 seconds before refetching
+    this.refetchTimeout = setTimeout(async () => {
```
xzilja, line 124: This looks like it should have https://developer.mozilla.org/en-US/docs/Web/API/AbortController implemented to cancel previous request?

xzilja (approved):

PR #1781, opened 2024-01-19
feat(analytics-option): fetch analytics flag from API

Cali93:
# Breaking Changes

N/A

# Changes

- feat(analytics-option): fetch analytics flag from API


packages/core/src/controllers/ApiController.ts
```diff
@@ -184,14 +191,24 @@ export const ApiController = {
...
-      ]),
-      CoreHelperUtil.wait(3000)
-    ])
+    const promises = [
+      ApiController.fetchFeaturedWallets(),
+      ApiController.fetchRecommendedWallets(),
+      ApiController.fetchNetworkImages(),
+      ApiController.fetchConnectorImages()
+    ]
+    if (OptionsController.state.enableAnalytics === undefined) {
```
tomiir, line 200: is it necessary to check agains `undefined` directly?

tomiir:
Unless this is a fix, I'd base this off V4 branch and merge against it!

packages/core/src/controllers/ApiController.ts
Cali93: Yes because if the param is provided through the code (could be `true` or `false`) it takes precedence over the Cloud setting.

tomiir (approved):

packages/core/src/controllers/OptionsController.ts
```diff
@@ -68,7 +69,11 @@ export const OptionsController = {
   },
 
   setEnableAnalytics(enableAnalytics: OptionsControllerState['enableAnalytics']) {
```
xzilja, line 71: Since `setEnableAnalytics` is called very early on in web3modal lifecycle, wouldn't this always resolve to else clause while api is fetching the flag?

xzilja (commented):
Left small comment

packages/core/src/controllers/OptionsController.ts
Cali93: Good catch! Last commit fixes this 🙏

tomiir (approved):

PR #1803, opened 2024-01-23
flag for allWallets button

glitch-txs:
# Breaking Changes

N/A

# Changes

- feat: Add option to disable All Wallets button.

```ts
type AllWallets = 'SHOW' | 'HIDE' | 'ONLY_MOBILE'
```
  `'HIDE'` type will remove it completely, 'onlyMobile' will remove it only on desktop to keep support for WalletConnect protocol in both platforms (in Desktop we have the WalletConnect button but in mobile we only depend on the All Wallets button)
  
  
image

- fix:
- chore:

# Associated Issues

closes https://github.com/WalletConnect/web3modal/issues/1656 https://github.com/WalletConnect/web3modal/issues/1410 https://github.com/WalletConnect/web3modal/issues/1657


packages/core/src/controllers/OptionsController.ts
```diff
@@ -5,6 +5,7 @@ import type { CustomWallet, Metadata, ProjectId, SdkVersion, Tokens } from '../u
 // -- Types --------------------------------------------- //
 export interface OptionsControllerState {
   projectId: ProjectId
+  allWallets?: boolean | 'onlyMobile'
```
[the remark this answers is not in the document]
glitch-txs: 'show' won't take any effect, should we still add it? could easily be `'hide' | 'onlyMobile' ` anyways it's just a type xD

packages/core/src/controllers/OptionsController.ts
```diff
@@ -5,6 +5,7 @@ import type { CustomWallet, Metadata, ProjectId, SdkVersion, Tokens } from '../u
 // -- Types --------------------------------------------- //
 export interface OptionsControllerState {
   projectId: ProjectId
+  allWallets?: 'show' | 'hide' | 'onlyMobile'
```
xzilja, line 8: To stick with naming of other feature flags, could we rename this to `enableAllWallets`.  Perhaps constants can be uppercase as well i.e `SHOW`, `HIDE`, `ONLY_MOBILE` etc..

glitch-txs: Wouldn't it be a bit confusing?
The enable flag is currently only in the default wagmi config, while this is createWeb3Modal, also all the enable flags are booleans while this has string options?

tomiir (approved):

PR #1826, opened 2024-01-23
fix/feat: Wagmi V2 Type fixes, Email Update Flow Setup

xzilja:
# Changes

- feat: Setup in `wallet` package for new email update flow
- fix: Wagmi V2 types related issues, simplify email connector
- chore: Update to latest wagmi v2 dependencies


packages/scaffold/src/views/w3m-update-email-primary-otp-view/index.ts
```diff
@@ -0,0 +1,44 @@
...
+  public constructor() {
+    super()
+  }
+
+  // --  Private ------------------------------------------ //
+  override onOtpSubmit: OnOtpSubmitFn = async otp => {
+    try {
+      if (this.emailConnector) {
+        await this.emailConnector.provider.connectOtp({ otp })
+        EventsController.sendEvent({ type: 'track', event: 'EMAIL_VERIFICATION_CODE_PASS' })
```
tomiir, line 17: Do we care to distinguish between login top vs update'

packages/wagmi/src/utils/defaultWagmiCoreConfig.ts
```diff
@@ -1,32 +1,26 @@
...
+
 import { emailConnector } from '../connectors/EmailConnector.js'
 
-export interface ConfigOptions
-  extends Omit<
-    CreateConfigParameters,
-    'client' | 'chains' | 'connectors' | 'multiInjectedProviderDiscovery'
-  > {
+export interface ConfigOptions {
+  chains: CreateConfigParameters['chains']
```
tomiir, line 10: Doesn't this force users to use the `[Chain, ...Chain[]]` cast?

packages/wagmi/src/utils/defaultWagmiReactConfig.ts
```diff
@@ -35,12 +29,13 @@ export function defaultWagmiConfig({
...
   enableCoinbase,
-  enableWalletConnect,
   enableEmail,
-  enableEIP6963,
-  ...wagmiConfig
+  enableWalletConnect,
+  enableEIP6963
 }: ConfigOptions): Config {
   const connectors: CreateConnectorFn[] = []
+  const transportsArr = chains.map(chain => [chain.id, http()])
```
tomiir, line 37: I'm wondering if we want to use wc transport at all here? 
We used to have it as a custom transport in wagmi1 but I couldn't find a way to provide it as a new option rather than the only one thus I left the `http` one, but maybe we want it for tracking or similar

tomiir (approved):

PR #1828, opened 2024-01-23
fix: wallet chain change sync

tomiir:
# Breaking Changes

N/A

# Changes

- feat:
- fix: issue when user changes network from the wallet and network is not configured.
- chore:


glitch-txs (approved):

packages/wagmi/src/client.ts
```diff
@@ -315,11 +312,14 @@ export class Web3Modal extends Web3ModalScaffold {
     }
   }
 
-  private async syncBalance(address: Hex, chain: Chain) {
+  private async syncBalance(address: Hex, chainId: number) {
+    const chain = this.wagmiConfig.chains.find((c: Chain) => c.id === chainId)
+
+    const id = chain?.id || this.options?.defaultChain?.id || this.wagmiConfig?.chains?.[0]?.id
```
tomiir, line 318: This keeps V3 functionality. But we should find a way to handle the case where the user switches to a non-configured chain, since wagmiv2 throws when trying to do something with it.

PR #1834, opened 2024-01-24
added filter transactions by chain

glitch-txs:
# Breaking Changes

- Activity Tab is filtered by chain

# Changes

- feat: Added filter transactions by chain


packages/core/src/controllers/TransactionsController.ts
```diff
@@ -12,6 +13,7 @@ type TransactionByYearMap = Record<number, TransactionByMonthMap>
 export interface TransactionsControllerState {
   transactions: Transaction[]
   transactionsByYear: TransactionByYearMap
+  lastNetworkInView: string | undefined
```
xzilja, line 16: Should this be that stricter `${string}:${string}` type by any chance? We might also have a type like CaipChain around that you could reuse for these.

packages/scaffold/src/views/w3m-transactions-view/index.ts
```diff
@@ -52,8 +51,10 @@ export class W3mTransactionsView extends LitElement {
             }
           }
         }),
+        NetworkController.subscribe(() => {
```
xzilja, line 54: This should probably be a subscribeKey to a specific field, I think caipChain, so it re-fetches only when that one change happens vs if anything else in controller changes.

xzilja (commented):
Looks like build is failing, so only leaving comments for now. As we are filtering data client side, will need to see how this affects pagination logic (we likely display less results than what we fetched).

Any chance this chain filtering can be done on blockchain api level vs on client?

packages/core/src/controllers/TransactionsController.ts
```diff
@@ -13,6 +15,7 @@ export interface TransactionsControllerState {
   transactions: Transaction[]
   coinbaseTransactions: TransactionByYearMap
   transactionsByYear: TransactionByYearMap
+  lastNetworkInView: CaipNetworkId | undefined
```
tomiir, line 18: What exactly is this for?

packages/core/src/controllers/TransactionsController.ts
```diff
@@ -124,9 +135,19 @@ export const TransactionsController = {
...
   },
 
+  filterByConnectedChain(transactions: Transaction[]) {
+    const chainId = NetworkController.state.caipNetwork?.id
+    const filteredTransactions = transactions.filter(
+      transaction => transaction.metadata.chain === chainId
+    )
+
+    return filteredTransactions
+  },
```
tomiir, line 145: I thought we were requesting the filtered txs directly from blockchain api 🤔 

wouldn't this mess up with pagination?

packages/core/src/controllers/TransactionsController.ts
glitch-txs: it tracks the connected network during the web component lifecycle, it helps to:

- Refetch the txs if the Activity view is open and the chain is changed from the wallet extension.
- Refetch the txs if the chain is changed when the Activity component is not mounted and then mounted/open.
- Prevent to refetch the txs every time the Activity view is opened.

packages/core/src/controllers/TransactionsController.ts
glitch-txs: > wouldn't this mess up with pagination?

Mmm I'm not sure and I'm not able to test this 🤔

tomiir: @geekbrother any way we can do the filterin directly on blockchain API side? I think there's an edge case to this where if we filter txs and don't have enough to fill the ui, we will never request more txs even if you have them

packages/scaffold-ui/src/partials/w3m-activity-list/index.ts
```diff
@@ -100,28 +105,50 @@ export class W3mActivityList extends LitElement {
...
         if (!transactions) {
           return null
         }
 
         return html`
-          <wui-flex flexDirection="column">
+          <wui-flex
+            flexDirection="column"
+            class="group-container"
+            last-group="${isLastGroup ? 'true' : 'false'}"
```
tomiir, line 150: why use strings? 
if neded, we can do `String(isLastGroup)`

enesozturk (approved):

PR #1847, opened 2024-01-28
chore: improve activity UI

tomiir:
# Breaking Changes

N/A

# Changes

- feat:
- fix: activity ui changes / fixes requested here: https://www.figma.com/file/gmX3t1YIFLvXdKEwbeuqEW/Web3Modal-v3?node-id=14782%3A463017&mode=dev
- chore:


xzilja (approved):

packages/core/tests/controllers/ApiController.test.ts
```diff
@@ -10,7 +10,8 @@ describe('ApiController', () => {
       featured: [],
       recommended: [],
       wallets: [],
-      search: []
+      search: [],
+      isAnalyticsEnabled: false
```
glitch-txs, line 14: yay! isn't this a breaking change?

glitch-txs (approved):

packages/core/tests/controllers/ApiController.test.ts
tomiir: this was already deployed, but the test was missing. It's not breaking cause the default is false.

glitch-txs: ah my bad, thought it was isActivityEnabled

PR #1848, opened 2024-01-28
fix: Onramp UI

tomiir:
# Breaking Changes

N/A

# Changes

- feat:
- fix: UI Review for Onramp flow
- chore:


xzilja (approved):
Looking good 👍 I can see this is going into a feature branch instead of V4. Perhaps it is time to set it up (hide it) behind a feature flag and start opening pr's against V4?

PR #1860, opened 2024-01-31
handle unsupported chain

svenvoskamp:
Added a solution for when you're connected with a chain that is not supported

* When `caipNetwork` changes, check if chain is supported 
* Added view to show when chain is not supported 
* Updated buttons to show a warning when chain is not supported


packages/scaffold/src/views/w3m-unsupported-chain-view/index.ts
```diff
@@ -0,0 +1,134 @@
...
+  // -- State & Properties --------------------------------- //
+  @state() private disconecting = false
+
+  // -- Render -------------------------------------------- //
+  public override render() {
+    return html`
+      <wui-flex
+        class="container"
+        flexDirection="column"
+        .padding=${['0', '0', '0', '0'] as const}
```
tomiir, line 34: is this required?

packages/ui/src/composites/wui-network-image/index.ts
```diff
@@ -25,17 +26,19 @@ export class WuiNetworkImage extends LitElement {
...
-      --local-icon-size: ${isLg ? '42px' : '24px'};
+      --local-path: var(--wui-path-network-${this.size});
+      --local-width:  var(--wui-width-network-${this.size});
+      --local-height:  var(--wui-height-network-${this.size});
+      --local-icon-size:  var(--wui-icon-size-network-${this.size});
     `
 
-    return html`${this.templateVisual()} ${isLg ? networkLgSvg : networkSvg}`
+    // eslint-disable-next-line no-nested-ternary
+    return html`${this.templateVisual()} ${isLg ? networkSvgLg : isMd ? networkSvgMd : networkSvgSm}`
```
tomiir, line 41: Just a suggestion, might be simpler to mantain something like:
```
const imagesBySize = { sm: smImage, md; mdImage, lg: lgImage }
...
return html`${this.templateVisual()} ${imagesBySize[size]}`
...

```

tomiir (dismissed):

tomiir (approved):

PR #1866, opened 2024-02-02
Refactor logic for which wallets to show in modal 

svenvoskamp:
# Changes

* Base injected connector (browser wallet) will always be in connector array on wagmi when enabled, even when there are now browser extension. I filtered this one out so we can show default wallets when browser extension is absent. 
* Custom wallets will show up again when configured 

# Associated Issues

closes https://github.com/WalletConnect/web3modal/issues/1863


glitch-txs (changes requested):
Let's not remove this one please, we need to ask wagmi to update their types

tomiir (approved):

glitch-txs (approved):

PR #1881, opened 2024-02-07
auto update constantsutil version

svenvoskamp:
# Changes

- feat: `Constantsutil.version` will match package.json version of scaffold-utils


tomiir (approved):
what a smart trick 🔥

PR #1886, opened 2024-02-08
docs(tests): improve instructions for local playwright setup

bkrem:
# Changes

- Improve setup instructions based on the issues I bumped into on my end setting up the playwright tests
- Annotate the `.env.example` with where the required values can be looked up

### DIFF
--- a/apps/laboratory/tests/README.md
+++ b/apps/laboratory/tests/README.md
@@ -5,9 +5,14 @@ We use Playwright as our functional test runner. It's configured to try multiple
 - Browsers: Chromium/Firefox
 - Modal flavors: default/SIWE/email
 
+## Setup
+
+- Make sure your `.env` is set up (see `.env.example` for reference)
+- Run `npm run playwright:install` to install the browsers required to run the tests
+
 ## Running Tests
 
-- `npx playwright test` to run in default mode (make sure your `.env` is set up)
+- `npx playwright test` to run in default mode.
 - `npm run playwright:debug` to step by step see what the tests are doing
 
 ## Debugging
--- a/apps/laboratory/.env.example
+++ b/apps/laboratory/.env.example
@@ -1,5 +1,6 @@
 # Obtain a project ID from https://cloud.walletconnect.com
 NEXT_PUBLIC_PROJECT_ID=""
+# See 1password for `NEXT_AUTH_SECRET`
 NEXTAUTH_SECRET=""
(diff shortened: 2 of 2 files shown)


apps/laboratory/tests/README.md
```diff
@@ -5,9 +5,14 @@ We use Playwright as our functional test runner. It's configured to try multiple
 - Browsers: Chromium/Firefox
 - Modal flavors: default/SIWE/email
 
+## Setup
+
+- Make sure your `.env` is set up (see `.env.example` for reference)
```
bkrem, line 10: @arein @svenvoskamp Something I'm currently still unclear on: where can I find the right values for local runs for the following:

```
NEXTAUTH_SECRET=""
NEXTAUTH_URL=""
MAILSAC_API_KEY=""
```

Without them the SIWE and Email tests seem to fail, and I wasn't able to find setup instructions for these values

glitch-txs: you can get them from Vercel / Labs deployment

arein: Mailsax: 1Password `Mailsac Web3Modal API Key`

arein (approved):

apps/laboratory/tests/README.md
bkrem: Thanks both, will update instructions accordingly for the next dev doing this 👍

bkrem: @glitch-txs do you know what `NEXTAUTH_URL` should be set to? Not defined in 1password or in w3m lab Vercel env variables and if I leave it blank/default I get `ERR_INVALID_URL`.

glitch-txs: I don't know but you can leave it as `NEXTAUTH_URL=""` I think

cc @Cali93

glitch-txs: it's from here: https://next-auth.js.org/configuration/options#nextauth_url
but only refers to prod in docs

bkrem: Ok so I figured out what the confusing thing was:
* `NEXTAUTH_URL` needs to point to the lab itself, so explicitly `http://localhost:3000` works but it's actually simpler to just omit it completely since it doesn't need to be set locally.
* Using this `.env.example` as-is was causing it to pick up the URL as empty string `""`

tomiir (approved):

PR #1897, opened 2024-02-09
feat: add workflow to clear gh cache on pr close

tomiir:
# Breaking Changes

N/A

# Changes

- feat: add workflow to clear gh cache on pr close
- fix: 
- chore:

### DIFF
--- a/.github/workflows/clear_cache.yml
+++ b/.github/workflows/clear_cache.yml
@@ -1,3 +1,9 @@
+# GitHub Caches are branch specific with non-default branches having access
+# to the `default` branch.
+# It appears that you cannot configure https://github.com/actions/cache to
+# only cache if `default` branch.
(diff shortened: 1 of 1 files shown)


.github/workflows/clear_cache.yml
```diff
@@ -0,0 +1,29 @@
+name: Post-merge Cache Cleanup
```
arein, line 1:
```suggestion
# GitHub Caches are branch specific with non-default branches having access
# to the `default` branch.
# It appears that you cannot configure https://github.com/actions/cache to
# only cache if `default` branch.
# This workflow helps with housekeeping the caches.
# See https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#force-deleting-cache-entries
name: Post-merge Cache Cleanup
```

arein (approved):

PR #1906, opened 2024-02-12
Release/4.0.3

xzilja:
# Breaking Changes

N/A

# Changes

- feat: Add theme sync every time when secure wallet app view is opened
- chore: Tag email as beta 
- chore: Revert to using manual version in scafolld utils constants

# Associated Issues

closes #...


svenvoskamp (approved):

packages/scaffold-utils/src/ConstantsUtil.ts
```diff
@@ -13,5 +11,5 @@ export const ConstantsUtil = {
   ADD_CHAIN_METHOD: 'wallet_addEthereumChain',
   EIP6963_ANNOUNCE_EVENT: 'eip6963:announceProvider',
   EIP6963_REQUEST_EVENT: 'eip6963:requestProvider',
-  VERSION: scaffoldUtilsPackageJson.version
+  VERSION: '4.0.2'
```
enesozturk, line 14: Should this be `4.0.3`? @xzilja

enesozturk (dismissed):
✨

packages/scaffold-utils/src/ConstantsUtil.ts
tomiir: @svenvoskamp did we merge the auto-updater ???

xzilja: 4.0.3 indeed. @tomiir I reverted auto updating as it was based on `assert {as json}` bit. It'll fail in quiet a few bundlers that have not implemented the standard yet.

PR #1921, opened 2024-02-14
Fix client.ts: reset EthersStoreUtil's error before new operation

dextfprotocol:
At the moment the **EthersStoreUtil**'s `error` is not reset before executing a new operation, therefore the value is carried over even for successful operations.

With this fix we reset `error` to `undefined` before every operation.

# Breaking Changes

N/A

# Changes

- fix: reset **EthersStoreUtil**'s `error` to `undefined` before every operation

# Associated Issues

closes #...


svenvoskamp (changes requested):
Nice catch! Can you please get CI to green, so we can approve 🙏

dextfprotocol (commented):
Done

glitch-txs (approved):
LGTM

svenvoskamp (approved):

PR #1925, opened 2024-02-14
fix: buy in progress interval being cleared prematurely

tomiir:
# Breaking Changes

N/A

# Changes

- feat:
- fix: buy in progress interval premature clear
- chore:


arein (approved):
You really need to add some regression tests here - this is ok without because it's a hotfix

PR #1933, opened 2024-02-16
fix: various email issues from w3i testing 

svenvoskamp:
# Changes

* Fix: In the signing message modal, the message area is not toggling when clicking on it. So only a small part of the message 
* Fix: When rejecting the device approving does nothing on w3m
* Fix: When typing the wrong code, need to remove the first character to paste another code
* Fix: Changing email address results in a back button that doesn't do anything
* Fix: Bad error styling on OTP view

# Associated Issues

closes #...


packages/core/src/controllers/RouterController.ts
```diff
@@ -72,8 +72,15 @@ export const RouterController = {
...
   },
 
-  replace(view: RouterControllerState['view'], data?: RouterControllerState['data']) {
-    if (state.history.length > 1 && state.history.at(-1) !== view) {
+  replace(
+    view: RouterControllerState['view'],
+    data?: RouterControllerState['data'],
+    reset?: boolean
+  ) {
+    if (reset) {
```
xzilja, line 80: instead of complicating this method, we could make a new `reset()` one perhaps?

xzilja (approved):
Looks good, left small suggestion

xzilja (approved):

tomiir (approved):

PR #1936, opened 2024-02-16
feat: test improvements

lukaisailovic:
# Changes
- Use deeplinks for connecting instead of copy pasting URI
- Added separate test for QR code connection
- General improvements and optimizations to tests

# Closes
#1775


svenvoskamp (approved):
LGTM ✨

PR #1937, opened 2024-02-19
fix: infinite loader when user is not connected

lukaisailovic:
When we receive `FRAME_IS_CONNECTED_SUCCESS` with value isConnected set to `false` instead of `true`, we are stuck in infinite loading state

### DIFF
--- a/packages/wagmi/src/client.ts
+++ b/packages/wagmi/src/client.ts
@@ -387,6 +387,7 @@ export class Web3Modal extends Web3ModalScaffold {
 
       const provider = (await connector.getProvider()) as W3mFrameProvider
       const isLoginEmailUsed = provider.getLoginEmailUsed()
+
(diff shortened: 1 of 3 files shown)


xzilja (dismissed):
Nice 👌

packages/wagmi/src/client.ts
```diff
@@ -397,9 +398,16 @@ export class Web3Modal extends Web3ModalScaffold {
           super.open({ view: 'ApproveTransaction' })
         }
       })
+
       provider.onRpcResponse(() => {
         super.close()
       })
+
+      provider.onNotConnected(() => {
```
svenvoskamp, line 406: Could you also add this to ethers client in `watchEmail` method?

svenvoskamp (changes requested):

svenvoskamp (approved):

PR #1957, opened 2024-02-22
Siwe package

glitch-txs:
# Breaking Changes

N/A

# Changes

- feat:
- fix:

  - Removed unused SIWE code from Scaffold
  - Moved SIWE code to its own package -> Core/ui/Scaffold/Client
  - Remove ts-expect-error declarations in Wagmi package
  - SIWE types were being exported from `@web3modal/core` which is a sub-dependency, types were moved to the SIWE package

- chore:

# Associated Issues

closes #...


xzilja (commented):
Good stuff, a lot of siwe changes here, I'd ping Cali just in case to also review. Left few suggestions. Ideally we should test this via installing deps using yarn / pnpm and npm in say next / vite projects, just to verify everything works correctly at build and run time.

packages/scaffold/src/partials/w3m-header/index.ts
```diff
@@ -107,8 +107,11 @@ export class W3mHeader extends LitElement {
   }
 
   private async onClose() {
-    if (SIWEController.state.isSiweEnabled && SIWEController.state.status !== 'success') {
-      await ConnectionController.disconnect()
+    if (OptionsController.state.isSiweEnabled) {
+      const { SIWEController } = await import('@web3modal/siwe')
```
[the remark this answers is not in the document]
glitch-txs: I don't think we can, we need to dynamic import everywhere to avoid peer issues

packages/siwe/exports/index.ts
```diff
@@ -1,8 +1,25 @@
-import type { SIWEConfig, SIWESession } from '@web3modal/core'
+import type {
+  SIWEConfig,
+  SIWESession,
+  SIWECreateMessageArgs,
+  SIWEVerifyMessageArgs,
+  SIWEClientMethods
+} from '../core/utils/TypeUtils.js'
```
tomiir, line 7: can't you do directly `export type { ... } from '../core/utils/TypeUtils.js`?

packages/siwe/package.json
```diff
@@ -19,26 +19,13 @@
...
-  "peerDependenciesMeta": {
-    "react": {
-      "optional": true
-    },
-    "react-dom": {
-      "optional": true
-    },
-    "vue": {
-      "optional": true
-    }
```
tomiir, line 36: Any consequence removing these?

turbo.json
```diff
@@ -37,12 +37,12 @@
...
     "build:scaffold-react": {
       "outputs": ["dist/**"],
-      "dependsOn": ["^build:scaffold-utils"],
+      "dependsOn": ["^build:scaffold-utils", "^build:scaffold", "^build:siwe"],
       "cache": false
     },
     "build:scaffold-vue": {
       "outputs": ["dist/**"],
-      "dependsOn": ["^build:scaffold-utils"],
+      "dependsOn": ["^build:scaffold-utils", "^build:scaffold", "^build:siwe"],
```
tomiir, line 45: I think `"^build:scaffold"` would not be needed given it already depends on `"^build:scaffold-utils"`

packages/siwe/package.json
glitch-txs: they are unused

packages/siwe/exports/index.ts
glitch-txs: Some are from different origin and some are used in the module so this would add extra lines of code + repeated path declarations

turbo.json
glitch-txs: if I remove them it breaks the build, seems like I need to explicitly declared all the packages is dependent on

tomiir (commented):
Lgtm ser 🚀

tomiir (approved):

Cali93 (commented):
LGTM, I tried it on labs and it still works as expected 💪 
Did it have an impact on bundle size ?

glitch-txs:
Thanks @Cali93 ! Haven't tested its bundle size yet, we'll see 👀

PR #1958, opened 2024-02-22
hotfix: better handling for rpc request 

svenvoskamp:
# Changes

fix: add multiple rpc methods to safe list for auto approval, so users don't end up with blank screen
fix: if rpc method is not handled by our frame provider, we show an error message. Request is automatically rejected
fix: rpc request will be rejected when you close the modal 
chore: bump version to `4.0.9`


packages/ethers/src/client.ts
```diff
@@ -755,8 +757,16 @@ export class Web3Modal extends Web3ModalScaffold {
...
+        if (W3mFrameHelpers.checkIfRequestExists(request)) {
+          if (!W3mFrameHelpers.checkIfRequestIsAllowed(request)) {
+            super.open({ view: 'ApproveTransaction' })
+          }
+        } else {
+          this.emailProvider?.rejectRpcRequest()
+          super.open()
+          setTimeout(() => {
+            SnackController.showError('This RPC method is not supported')
+          }, 300)
```
arein, line 769: So 
> so users don't end up with blank screen
is not correct right? They now see "This RPC Method is not supported"?

arein (approved):
LGTM but nitpick: I'd call it a hotfix because there are not tests in this PR.
Given the priority that's right - but maybe a hotfix in the follow up PR

packages/ethers/src/client.ts
svenvoskamp: If an RPC request is on our safe/not-safe list, we will handle it. Safe list requests will be automatically approved, and not-safe list requests will open the approve request view in the modal.

If an RPC request is not part of either of these lists, we will show "This RPC Method is not supported."

lukaisailovic: Since this is a user facing notification, can we change it to something in terms of "Requested action not supported" and and console.log for developers where we say rpc not supported with logging the exact method that is not supported?

PR #1994, opened 2024-02-27
fix: no retries for canary and refactor scripts

chris13524:
Removes retries from the canary since the purpose of the canary is to detect transient failures. [Slack DM from @arein](https://walletconnect.slack.com/archives/D04GF8MV5M3/p1708756971299939)

Also refactors the NPM scripts to avoid duplicate configs.


lukaisailovic (approved):

apps/laboratory/package.json
```diff
@@ -11,11 +11,11 @@
...
-    "playwright:test:canary": "npx playwright test --grep canary.spec.ts --project='Desktop Chrome/wagmi'",
-    "playwright:debug": "npx playwright test --debug",
-    "playwright:debug:wallet": "npx playwright test --debug 'connect-qr.spec.ts|wallet.spec.ts'",
-    "playwright:debug:siwe": "npx playwright test --debug --grep siwe.spec.ts",
-    "playwright:debug:canary": "npx playwright test --debug --grep canary.spec.ts --project='Desktop Chrome/wagmi'"
+    "playwright:test:canary": "npx playwright test --retries=0 --grep canary.spec.ts --project='Desktop Chrome/wagmi'",
+    "playwright:debug": "npm run playwright:test -- --debug",
+    "playwright:debug:wallet": "npm run playwright:test:wallet -- --debug",
+    "playwright:debug:siwe": "npm run playwright:test:siwe -- --debug",
+    "playwright:debug:canary": "npm run playwright:test:canary -- --debug"
```
arein, line 18: do you not need `--project='Desktop Chrome/wagmi'` anymore?

chris13524: We don't here, because now instead we are calling the other NPM script which does have it

PR #1997, opened 2024-02-28
feat: add support for more browsers in tests

lukaisailovic:
# Breaking Changes

N/A

# Changes

- feat:
- fix:
- chore:

# Associated Issues

closes #...

### DIFF
--- a/apps/laboratory/tests/shared/utils/device.ts
+++ b/apps/laboratory/tests/shared/utils/device.ts
@@ -1,9 +1,5 @@
 import { DEVICES } from '../constants/devices'
 
 export function getAvailableDevices(): string[] {
-  if (!process.env['CI']) {
-    return DEVICES
-  }
-
-  return DEVICES.filter(d => d !== 'Desktop Safari')
+  return DEVICES
 }


apps/laboratory/tests/shared/utils/project.ts
```diff
@@ -9,16 +11,35 @@ const PERMUTATIONS = availableDevices.flatMap(device =>
...
+  useOptions?: UseOptions
 }
 
 export type CustomProjectProperties = {
   [T in string]: CustomProperties
 }
 
+const braveOptions: UseOptions = {
+  launchOptions: {
+    executablePath: getValue(BRAVE_UBUNTU_PATH, BRAVE_MACOS_PATH)
```
tomiir, line 32: won't this break locally for windows devs? not sure we have any though but jic

tomiir (approved):

apps/laboratory/tests/shared/utils/project.ts
lukaisailovic: It will, but we are fine with that for now

apps/laboratory/tests/shared/constants/devices.ts
```diff
@@ -1 +1 @@
-export const DEVICES = ['Desktop Chrome', 'Desktop Firefox', 'Desktop Safari']
+export const DEVICES = ['Desktop Chrome', 'Desktop Brave', 'Desktop Firefox']
```
chris13524, line 1: Why did we remove Safari?

lukaisailovic: It was disabled for some time in the utility function, so disabled it here as well. Its not working on ubuntu runners. Will debug separately.

PR #2012, opened 2024-02-29
chore: refactor email tests

lukaisailovic:
Fix email tests and expand email testing suite


svenvoskamp (approved):
Nice! LGTM ✨

apps/laboratory/tests/shared/utils/email.ts
```diff
@@ -1,67 +1,87 @@
-import { Mailsac, type EmailMessage } from '@mailsac/api'
+import { Mailsac } from '@mailsac/api'
+const EMAIL_CHECK_TIMEOUT = 1000
+const MAX_EMAIL_CHECK = 16
```
bkrem, line 3: Is this limit based on something in particular or just something that works?

bkrem (approved):
💯

apps/laboratory/tests/shared/utils/email.ts
lukaisailovic: Ideally we want to keep that as low as possible to reduce the time it takes to run the tests. 10 was the initial limit, and slowly increased until I found one where almost every test would pass.

PR #2014, opened 2024-03-01
Feat: smart account init

tomiir:
# Breaking Changes

N/A

# Changes

- feat: add Smart Account initialization handlers and base upgrade screen in order to test it out.
- fix:
- chore:


apps/laboratory/src/components/Wagmi/WagmiTransactionTest.tsx
```diff
@@ -7,7 +7,7 @@ import { sepolia } from 'wagmi/chains'
 
 const TEST_TX = {
   to: vitalikEthAddress as Address,
-  value: parseGwei('0.0002')
+  value: parseGwei('0.3')
 }
```
lukaisailovic, line 11: nitpick: can we send back to faucet instead?

lukaisailovic (dismissed):
We should start thinking about how to test this with playwright as well

svenvoskamp (approved):
LFG ✨

PR #2017, opened 2024-03-03
Add a hook to return connected address

### LINKED ISSUE #1665: [feature] Hook for Returning Wallet Information

**Issue:**
I'm facing a challenge in identifying which wallet the user has chosen to connect. I've explored the documentation and codebase but couldn't find a solution.

**Description:**
In the context of our application, we use web3modal, and it would be immensely beneficial to retrieve information about the user's chosen wallet during the connection process. This information is crucial for improving user experience and our analytics.

**Background:**
During previous discussions here https://github.com/orgs/WalletConnect/discussions/4027, it was recommended that a dedicated hook be introduced to handle these connections and return relevant wallet information seamlessly.
--> This feature request aligns with the feedback from these discussions, aiming to address the identified gap by proposing the implementation of a specialized hook for retrieving wallet information within the Web3Modal repository.

Thank you for your assistance!

For now I am able to get that information when the user connects through WalletConnect.
But not when the user connects directly with a wallet, eg MetaMask.

I would like to implement the recommendation made by @glitch-txs in the following discussion: https://github.com/orgs/WalletConnect/discussions/4027#discussioncomment-7935554.

The suggestion is to introduce a dedicated hook in the Web3Modal repository for returning wallet information:
> I think we could build a hook for this, could you open an issue in Web3Modal repository for this? we could abstract WalletConnect & EIP-6963 connections into a hook that returns the wallet information or even in the events hook

Nazeeh21:
# Breaking Changes

N/A

# Changes
- feat: Added a hook `useAddress` that will return two things namely `address` that will be the currently connected address and `sessionInfo`.
- fix: 
- chore:

# Associated Issues

closes #1665


glitch-txs (changes requested):
Thanks! We're currently structuring the library differently.

- Stores are in the Core package, we can create a `walletInfo` state in [AccountConntroller](https://github.com/WalletConnect/web3modal/blob/V4/packages/core/src/controllers/AccountController.ts)
- In [Scaffold client](https://github.com/WalletConnect/web3modal/blob/V4/packages/scaffold/src/client.ts) we can add a `setWalletInfo` 
- in Wagmi and Ethers client's we can then add a syncWalletInfo, and update it when there's a change in the account state.
- We need to check if the connector is WalletConnect or Injected, Injected will also have this data if it was used with EIP-6963

This will keep the structure consistent and also add support for the JS export.

glitch-txs:
Closing in favor of https://github.com/WalletConnect/web3modal/pull/2099

Sorry as we needed this feature internally asap 🙏 🙏

PR #2033, opened 2024-03-06
feat: Sign 2.5

ganchoradkov:
PR adding Sign 2.5 in `web3modal/ethers5` & `web3modal/siwe`
- on siwe request, sign 2.5 authentication is used by default - if it's not supported by the wallet it fallbacks to legacy
- all `web3modal/siwe` behaviour preserved for backwards compatibility
- new `getMessageParams` client method is added to `createSIWEConfig` so the params that compose siwe message can be passed to Sign for multi-chain authentication
- replaced `siwe` package with `@walletconnect/utils` as recaps in siwe statement was not supported

## to test
wallet (on sign 2.5) - https://react-wallet-v2-kmnfc86q1-walletconnect1.vercel.app/walletconnect
legacy - any other supporting v2 protocol


packages/core/src/controllers/OptionsController.ts
```diff
@@ -19,6 +19,9 @@ export interface OptionsControllerState {
   enableAnalytics?: boolean
   metadata?: Metadata
   enableOnramp?: boolean
+  siweSession?: {
+    accounts: string[]
+  }
```
Cali93, line 24: Not sure if this should live in core.
@glitch-txs recently moved anything related to SIWE into its own package.

packages/ethers5/src/client.ts
```diff
@@ -168,8 +170,36 @@ export class Web3Modal extends Web3ModalScaffold {
...
+
+        if (siweConfig?.options?.enabled) {
+          const result = await WalletConnectProvider.authenticate({
+            nonce: await siweConfig.getNonce(),
+            methods: OPTIONAL_METHODS,
+            ...siweConfig.options.messageParams
+          })
+          if (result?.auths?.[0]) {
+            const chainId = parseInt(result.auths[0].p.chainId?.split(':')?.[1] || '', 10)
+            const address = result.auths[0]?.p.iss.split(':')[2] || ''
```
Cali93, line 184: Can we use formatting util functions here ?

packages/ethers5/src/client.ts
```diff
@@ -168,8 +170,36 @@ export class Web3Modal extends Web3ModalScaffold {
...
+            SIWEController.setSession({
+              address,
+              chainId
+            })
 
-        await WalletConnectProvider.connect()
+            // Kicks off verifyMessage and populates external states
+            const message = WalletConnectProvider.signer.client.formatAuthMessage({
+              request: result.auths[0].p,
```
Cali93, line 192: What does` result.auths[0]` represent ? Can we use a variable with an explicit name ?

apps/laboratory/src/utils/SiweUtils.ts
```diff
@@ -4,17 +4,17 @@ import type { SIWEVerifyMessageArgs, SIWECreateMessageArgs, SIWESession } from '
...
-      uri: window.location.origin,
-      address,
-      chainId,
-      nonce,
-      // Human-readable ASCII assertion that the user will sign, and it must not contain `\n`.
-      statement: 'Sign in With Ethereum.'
-    }).prepareMessage(),
+  messageParams: {
+    domain: 'lab.web3modal.com',
+    uri: 'https://lab.web3modal.com',
```
Cali93, line 9: Why is it hardcoded ?
```suggestion
     domain: window.location.host,
     uri: window.location.origin,
```

apps/laboratory/src/utils/SiweUtils.ts
```diff
@@ -4,17 +4,17 @@ import type { SIWEVerifyMessageArgs, SIWECreateMessageArgs, SIWESession } from '
...
+    chains: [1, 56],
```
Cali93, line 10: Any chance we can make this dynamic ?

apps/laboratory/src/utils/SiweUtils.ts
```diff
@@ -4,17 +4,17 @@ import type { SIWEVerifyMessageArgs, SIWECreateMessageArgs, SIWESession } from '
...
+    resources: [
+      'https://siwe.dev',
+      'urn:recap:eyJhdHQiOnsiaHR0cHM6Ly93ZWIzaW5ib3guY29tIjp7InB1c2gvYWxlcnRzIjpbe31dLCJwdXNoL25vdGlmaWNhdGlvbnMiOlt7fV19fX0='
+    ],
+    statement: 'Test statement'
```
Cali93, line 15: Could you add context around resources and the `urn` ? Perhaps by using a variable with an explicit name ?

apps/laboratory/src/pages/api/auth/[...nextauth].ts
```diff
@@ -47,22 +45,19 @@ export default async function auth(req: NextApiRequest, res: NextApiResponse) {
...
-          )
+          const { message, signature } = credentials
+          const address = message.match(/0x[a-fA-F0-9]{40}/u)?.[0] || ''
+          const chainId = `eip155:${
+            credentials.message.match(/Chain ID: (?<temp1>\d+)/u)?.[1] || 1
+          }`
+          let isValid = isValidEip191Signature(address, message, signature)
+          if (!isValid) {
+            isValid = await isValidEip1271Signature(address, message, signature, chainId, projectId)
+          }
```
Comment (Cali93) [inline apps/laboratory/src/pages/api/auth/[...nextauth].ts:56]: What's the reason to stop relying on `siwe` for signature verification ? Is it because the message doesn't match the SIWE spec anymore ?

apps/laboratory/src/pages/api/auth/[...nextauth].ts
```diff
@@ -47,22 +45,19 @@ export default async function auth(req: NextApiRequest, res: NextApiResponse) {
...
-            {
-              signature: credentials?.signature || '',
-              nonce
-            },
-            { provider }
-          )
+          const { message, signature } = credentials
+          const address = message.match(/0x[a-fA-F0-9]{40}/u)?.[0] || ''
+          const chainId = `eip155:${
+            credentials.message.match(/Chain ID: (?<temp1>\d+)/u)?.[1] || 1
```
Comment (Cali93) [inline apps/laboratory/src/pages/api/auth/[...nextauth].ts:51]: If we can get the message to match the SiweMessage spec you can parse the message.
```js
const parsedMessage = new SiweMessage(message)
const chainId = `eip155:${parsedMessage.chainId}`
```

apps/laboratory/src/pages/library/ethers-siwe.tsx
```diff
@@ -5,26 +5,28 @@ import { createWeb3Modal, defaultConfig } from '@web3modal/ethers5/react'
...
+      projectId: ConstantsUtil.ProjectId,
+      enableAnalytics: true,
+      metadata: ConstantsUtil.Metadata,
+      siweConfig,
+      enableOnramp: true,
+      customWallets: ConstantsUtil.CustomWallets
+    })
+    ThemeStore.setModal(modal)
+  }, [])
+
```
Cali93, line 29: Curious why are we moving this into the component ?

apps/laboratory/src/utils/SiweUtils.ts
```diff
@@ -1,20 +1,17 @@
...
-      // Human-readable ASCII assertion that the user will sign, and it must not contain `\n`.
-      statement: 'Sign in With Ethereum.'
-    }).prepareMessage(),
+  messageParams: {
+    domain: window.location.host,
+    uri: window.location.origin,
+    chains: WagmiConstantsUtil.chains.map(chain => chain.id),
+    statement: 'Please sign with your account'
+  },
+  createMessage: (args: SIWECreateMessageArgs) => formatMessage({ ...args }, args.address),
```
Cali93, line 14: Should this omit the address for the first param ?
```suggestion
  createMessage: ({ address, ...args }: SIWECreateMessageArgs) => formatMessage(args, address),
```

packages/ethers5/src/client.ts
```diff
@@ -168,8 +171,39 @@ export class Web3Modal extends Web3ModalScaffold {
...
+
+        if (siweConfig?.options?.enabled) {
+          const result = await WalletConnectProvider.authenticate({
+            nonce: await siweConfig.getNonce(),
+            methods: OPTIONAL_METHODS,
+            ...siweConfig.options.messageParams
+          })
+          // Auths is an array of signed CACAO objects https://github.com/ChainAgnostic/CAIPs/blob/main/CAIPs/caip-74.md
+          if (result?.auths?.[0]) {
+            const { p, s } = result.auths[0]
```
Cali93, line 185: Please use a variable for better readability. Btw don't we need to iterate over the signed CACAOs and do this for each of them ?
```suggestion
          const signedCacaos = result?.auths?.[0]
          if (signedCacaos) {
            const { p, s } = signedCacaos
```

apps/laboratory/src/pages/api/auth/[...nextauth].ts
Comment (ganchoradkov) [reply apps/laboratory/src/pages/api/auth/[...nextauth].ts:56]: `siwe` doesn't support `ReCaps` thus failing validation

apps/laboratory/src/pages/api/auth/[...nextauth].ts
Comment (ganchoradkov) [reply apps/laboratory/src/pages/api/auth/[...nextauth].ts:51]: `siwe` doesn't support `ReCaps` thus failing validation

apps/laboratory/src/pages/library/ethers-siwe.tsx
ganchoradkov: https://github.com/WalletConnect/web3modal/blob/435505ecc37ba3d06aeb3a34bdb03972e233e8d2/apps/laboratory/src/utils/SiweUtils.ts#L9

results in `Window is not defined` Next.js exception so we have to execute this only client side

packages/ethers5/src/client.ts
```diff
@@ -11,9 +12,10 @@ import type {
   Token
 } from '@web3modal/scaffold'
 import { Web3ModalScaffold } from '@web3modal/scaffold'
-import type { Web3ModalSIWEClient } from '@web3modal/siwe'
+import { SIWEController, type Web3ModalSIWEClient } from '@web3modal/siwe'
```
glitch-txs, line 15: SIWEController should be dynamically imported
```suggestion
import type { Web3ModalSIWEClient } from '@web3modal/siwe'
```

packages/ethers5/src/client.ts
```diff
@@ -138,6 +140,7 @@ export class Web3Modal extends Web3ModalScaffold {
               )
             }
             const ns = provider.signer?.session?.namespaces
+            console.log('WalletConnectProvider ns', ns)
```
glitch-txs, line 143:
```suggestion
```

packages/ethers5/src/client.ts
```diff
@@ -168,8 +171,40 @@ export class Web3Modal extends Web3ModalScaffold {
         WalletConnectProvider.on('display_uri', (uri: string) => {
           onUri(uri)
         })
-
-        await WalletConnectProvider.connect()
+        console.log('SIWEController', SIWEController, siweConfig)
+        console.log('WalletConnectProvider init opts', options)
```
glitch-txs, line 175:
```suggestion
```

packages/ethers5/src/client.ts
```diff
@@ -168,8 +171,40 @@ export class Web3Modal extends Web3ModalScaffold {
         WalletConnectProvider.on('display_uri', (uri: string) => {
           onUri(uri)
         })
-
-        await WalletConnectProvider.connect()
+        console.log('SIWEController', SIWEController, siweConfig)
```
glitch-txs, line 174:
```suggestion
```

packages/ethers5/src/client.ts
```diff
@@ -168,8 +171,40 @@ export class Web3Modal extends Web3ModalScaffold {
...
+        console.log('WalletConnectProvider init opts', options)
+
+        if (siweConfig?.options?.enabled) {
+          const result = await WalletConnectProvider.authenticate({
```
glitch-txs, line 178:
```suggestion
          const { SIWEController } = await import('@web3modal/siwe')
          
          const result = await WalletConnectProvider.authenticate({
```

glitch-txs: should we move the `authenticate` function inside `WalletConnectProvider`?

ganchoradkov: I don't understand this point. It's already this way this way - `await WalletConnectProvider.authenticate`

glitch-txs: you're right nvm lol

apps/laboratory/src/pages/api/auth/[...nextauth].ts
```diff
@@ -47,22 +45,19 @@ export default async function auth(req: NextApiRequest, res: NextApiResponse) {
...
+          }`
+          let isValid = isValidEip191Signature(address, message, signature)
+          if (!isValid) {
+            isValid = await isValidEip1271Signature(address, message, signature, chainId, projectId)
+          }
 
-          if (result.success) {
+          if (isValid) {
```
Comment (peculiarity) [inline apps/laboratory/src/pages/api/auth/[...nextauth].ts:58]: What is your opinion of the following code:

```javascript
if (
   isValidEip191Signature(address, message, signature)
   || await isValidEip1271Signature(address, message, signature, chainId, projectId)
) {
     return {
         id: `${chainId}:${address}`
     }
}

File: packages/ethers5/package.json [hunk 18]
```diff
@@ -43,8 +43,8 @@
     "lint": "eslint . --ext .js,.jsx,.ts,.tsx"
   },
   "dependencies": {
+    "@walletconnect/ethereum-provider": "2.11.2-rc-sign2.5-1",
```
tomiir, line 46: does it have a stable release?

packages/siwe/src/client.ts
```diff
@@ -90,7 +94,14 @@ export class Web3ModalSIWEClient {
     if (!chainId) {
       throw new Error('A chainId is required to create a SIWE message.')
     }
-    const message = this.methods.createMessage({ address, nonce, chainId })
+    console.log('chainId', chainId, address)
```
tomiir, line 97: same

packages/ethers5/src/client.ts
```diff
@@ -169,7 +171,38 @@ export class Web3Modal extends Web3ModalScaffold {
...
+            // Kicks off verifyMessage and populates external states
+            const message = WalletConnectProvider.signer.client.formatAuthMessage({
+              request: p,
+              iss: p.iss
+            })
+            SIWEController.verifyMessage({
+              message,
+              signature: s.s
+            })
+          }
```
tomiir, line 202: nit: Might be good to abstract this to a utils function like `verifyCacao` or similar

tomiir (approved):
Looks good, just needs logs removed and proper provider versions 🚀

apps/laboratory/src/pages/library/ethers-siwe.tsx
```diff
@@ -5,26 +5,28 @@ import { createWeb3Modal, defaultConfig } from '@web3modal/ethers5/react'
...
-  enableOnramp: true,
-  customWallets: ConstantsUtil.CustomWallets
-})
-
-ThemeStore.setModal(modal)
+import { useMemo } from 'react'
 
 export default function EthersSiwe() {
+  useMemo(async () => {
+    const { siweConfig } = await import('../../utils/SiweUtils')
```
glitch-txs, line 12: why is this dynamically imported?

glitch-txs (approved):
seems like we're using an RC version, should we add a stable one?

Cali93 (commented):
LGTM, the only concern I have is around the SIWE integration.
SIWE was designed to be a session per account (`address`+`chainId`).
What we end up with here is multiple signatures for the different chain Ids and a single session that defaults to mainnet.

I think in the short term, the user should be able to switch between chains without signing a new message if he has a valid signature for the given chain.
And switching to a chain that hasn't been approved yet should sign out the user -> connect and sign the message for the new chain or something along those lines.

For the mid and long term, we should look into having multiple cookie sessions for each signature.

apps/laboratory/src/pages/api/auth/[...nextauth].ts
```diff
@@ -47,22 +45,19 @@ export default async function auth(req: NextApiRequest, res: NextApiResponse) {
...
-          const nonce = await getCsrfToken({ req: { headers: req.headers } })
-          const result = await siwe.verify(
-            {
-              signature: credentials?.signature || '',
-              nonce
-            },
-            { provider }
-          )
+          const { message, signature } = credentials
+          const address = message.match(/0x[a-fA-F0-9]{40}/u)?.[0] || ''
```
Comment (enesozturk) [inline apps/laboratory/src/pages/api/auth/[...nextauth].ts:49]: I looks to me that these kind of regex values should be stored in a constant with a clear name so it could improve the readability and maintainability. 

There could be a map like the following:
```js
const regexPatterns = {
   address: /0x[a-fA-F0-9]{40}/u,
   chainId: /Chain ID: (?<temp1>\d+)/u,
   ...
}
```

packages/ethers5/src/client.ts
```diff
@@ -238,8 +271,13 @@ export class Web3Modal extends Web3ModalScaffold {
         const providerType = EthersStoreUtil.state.providerType
         localStorage.removeItem(EthersConstantsUtil.WALLET_ID)
         EthersStoreUtil.reset()
+        console.log(
```
enesozturk, line 274: I think Rocky already mentioned but a log, if we missed 🙏

apps/laboratory/src/utils/SiweUtils.ts
```diff
@@ -1,20 +1,19 @@
...
-      version: '1',
-      domain: window.location.host,
-      uri: window.location.origin,
-      address,
-      chainId,
-      nonce,
-      // Human-readable ASCII assertion that the user will sign, and it must not contain `\n`.
-      statement: 'Sign in With Ethereum.'
-    }).prepareMessage(),
+  // We don't require any async action to populate params but other apps might
```
enesozturk, line 8: Can we explain it a bit? In which situations the apps might require await the `getMessageParams` function? My concern is that users shouldn't confuse about how to call it and make sure about the type definitions since when we make it async, the return type will be also `Promise<..>`

enesozturk (approved):
Just left a few minor comments, LGTM!

apps/laboratory/src/utils/SiweUtils.ts
ganchoradkov: Any case where you might want to build dynamic `statement` or `uri` or `resources` with recaps or even provide `nonce` with the params instead of using the `getNonce` client method like
```
getMessageParams: async () => ({
    domain: window.location.host,
    uri: `${window.location.origin}/login/${await getSomeDynamicToken()}`,
    chains: WagmiConstantsUtil.chains.map(chain => chain.id),
    statement: 'Please sign with your account',
    nonce: await getCsrfToken()
  }),
```

glitch-txs (approved):

packages/siwe/scaffold/views/w3m-connecting-siwe-view/index.ts
```diff
@@ -69,6 +71,13 @@ export class W3mConnectingSiweView extends LitElement {
   }
 
   // -- Private ------------------------------------------- //
+
+  private onRender() {
+    if (SIWEController.state.session) {
+      ModalController.close()
+    }
+  }
```
tomiir, line 79: why do we need to close here?

tomiir (approved):

packages/siwe/scaffold/views/w3m-connecting-siwe-view/index.ts
ganchoradkov: there are two flows in Auth
- new 1-click auth flow that resolves after QR scan
- legacy personal_sign request after session is established

this conditions checks if session is already established with the new connect flow, or it should redirect to the legacy `ConnectSiwe` page

apps/laboratory/src/utils/SiweUtils.ts
```diff
@@ -1,20 +1,19 @@
-import { SiweMessage } from 'siwe'
 import { getCsrfToken, signIn, signOut, getSession } from 'next-auth/react'
 import type { SIWEVerifyMessageArgs, SIWECreateMessageArgs, SIWESession } from '@web3modal/siwe'
 import { createSIWEConfig } from '@web3modal/siwe'
+import { formatMessage } from '@walletconnect/utils'
```
glitch-txs, line 4: mmm I see we are importing a lot of functions and utilies from @walletconnect/utils, what if we export them from `@web3modal/siwe`? so devs would only need to install siwe instead of siwe + utils?

apps/laboratory/src/pages/api/auth/[...nextauth].ts
```diff
@@ -47,22 +48,19 @@ export default async function auth(req: NextApiRequest, res: NextApiResponse) {
...
-              nonce
-            },
-            { provider }
-          )
+          const { message, signature } = credentials
+          const address = message.match(ETH_ADDRESS_PATTERN)?.[0] || ''
+          const chainId = `eip155:${
+            credentials.message.match(ETH_CHAIN_ID_IN_SIWE_PATTERN)?.[1] || 1
+          }`
+          let isValid = isValidEip191Signature(address, message, signature)
```
Comment (glitch-txs) [inline apps/laboratory/src/pages/api/auth/[...nextauth].ts:56]: should we abstract this into a single function inside the SIWE package?

glitch-txs (changes requested):
let's add wagmi and abstract `@walletconnect/utils` inside `@web3modal/siwe`

apps/laboratory/src/pages/api/auth/[...nextauth].ts
Comment (glitch-txs) [reply apps/laboratory/src/pages/api/auth/[...nextauth].ts:56]: Following CAIP-74 shouldn't we get the type of signature from the header?

glitch-txs (approved):

apps/laboratory/src/utils/SiweUtils.ts
glitch-txs: @ganchoradkov I forgot to move this to the SIWE package. We could import it from there too

apps/laboratory/src/utils/SiweUtils.ts
glitch-txs: I think this could be of type `()=>(Promise<MessageParams> | MessageParams)`

So both sync and async fn could be allowed.

packages/siwe/core/utils/TypeUtils.ts
```diff
@@ -26,10 +26,36 @@ export type SIWEMessageArgs = {
   chains: number[]
   methods?: string[]
 } & Omit<SIWECreateMessageArgs, 'address' | 'chainId' | 'nonce' | 'version'>
+// Signed Cacao (CAIP-74)
+interface CacaoPayload {
```
devceline, line 30: Can't we use `CacaoPayload` from `@walletconnect/cacao`

UTkzhang:
@ganchoradkov @glitch-txs @tomiir @enesozturk 

This PR poses a few issues.

First, it is not backwards compatible with previous SIWE implementations. The new function requirement of getMessageParams should be optional instead.

Second, it makes current latest deployed docs outdated, one example being, that the new EIP155 standard ```eip155:chainId:address``` format is not compatible with latest siwe standard (and hence the siwe recommended npm lib for react, for e.g.). See https://eips.ethereum.org/EIPS/eip-4361 under "address". This should not be the default behaviour, and would cause another breaking change for many users (and one that isn't actually easily fixable, unless you force migration to eip155 address). 

IMO, these behaviours should be corrected in a future release.

PR #2038, opened 2024-03-11
:feat add send wallet feature

svenvoskamp:
* Add `wui-input-amount` component
* Add `wui-preview-item` component
* Add `wui-token-button` component 
* Add `w3m-input-token` partial
* Add `w3m-input-address` partial
* Add `w3m-wallet-send-details` partial
* Add `w3m-wallet-send-view` view
* Add `w3m-wallet-send-select-token` view
* Add `w3m-wallet-send-preview` view
* Add `SendController`


packages/scaffold/src/views/w3m-wallet-send-view/index.ts
```diff
@@ -0,0 +1,106 @@
...
+            : 'Invalid Address'
+        } else {
+          this.message = 'Add Address'
+        }
+      } else {
+        this.message = 'Add Amount'
+      }
+    } else {
+      this.message = 'Select Token'
+    }
```
tomiir, line 64:
```suggestion
    if(!this.token) {
      this.message = 'Select Token'
    }

    if(!this.sendTokenAmount) {
      this.message = 'Add Amount'
    }

    if(this.sendTokenAmount > Number(this.token.quantity.numeric)) {
      this.message = 'Insufficient Funds'
    }

    if(!this.receiverAddress) {
      this.message = 'Add Address'
    }

    if(!CoreHelperUtil.isAddress(this.receiverAddress)) {
      this.message = 'Invalid Address'
    }

    this.message = 'Preview Send'
```

tomiir: I think in this case using the ifs separately is much clearer. I'd in fact add an util method like `getMessage` or similar

packages/ui/src/composites/wui-tag/styles.ts
```diff
@@ -33,4 +33,12 @@ export default css`
     background-color: var(--wui-icon-box-bg-error-100);
     color: var(--wui-color-error-100);
   }
+
+  :host([data-size='lg']) {
+    padding: 9px 5px !important;
```
tomiir, line 38: is this `!important` ok?

tomiir (commented):
Some minor comments but looks good overall!

packages/scaffold/src/partials/w3m-input-token/index.ts
```diff
@@ -0,0 +1,122 @@
...
+  private handleSelectButtonClick() {
+    RouterController.push('WalletSendSelectToken')
+  }
+
+  private sendValueTemplate() {
+    if (this.token && this.sendTokenAmount) {
+      const price = this.token.price
+      const totalValue = price * this.sendTokenAmount
+
+      return html`<wui-text variant="small-400" color="fg-200">$${totalValue.toFixed(2)}</wui-text>`
```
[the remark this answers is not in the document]
svenvoskamp: This is not about the amount of the token, but the value in USD. Then we only need 2 decimals right?

packages/ui/src/composites/wui-tag/styles.ts
svenvoskamp: Yes we need it here

tomiir (approved):
LGTM 🚀

PR #2039, opened 2024-03-11
fix: filter connectors based on rdns

tomiir:
# Breaking Changes

N/A

# Changes

- feat:
- fix: filter duplicated connectors
- chore:

# Associated Issues
Closes https://linear.app/walletconnect/issue/W3M-81/metamask-is-showing-twice-in-web3modal
Closes https://linear.app/walletconnect/issue/W3M-80/coinbase-is-showing-twice

### DIFF
--- a/apps/laboratory/src/pages/library/ethers.tsx
+++ b/apps/laboratory/src/pages/library/ethers.tsx
@@ -18,9 +18,7 @@ const modal = createWeb3Modal({
   termsConditionsUrl: 'https://walletconnect.com/terms',
   privacyPolicyUrl: 'https://walletconnect.com/privacy',
   customWallets: ConstantsUtil.CustomWallets,
-  enableOnramp: true,
-  // Metamask wallet id
-  featuredWalletIds: ['c57ca95b47569778a828d19178114f4db188b89b763c899ba0be274e97267d96']
+  enableOnramp: true
 })
 
 ThemeStore.setModal(modal)
--- a/apps/laboratory/src/pages/library/ethers.tsx
+++ b/apps/laboratory/src/pages/library/ethers.tsx
@@ -19,6 +19,7 @@ const modal = createWeb3Modal({
   privacyPolicyUrl: 'https://walletconnect.com/privacy',
   customWallets: ConstantsUtil.CustomWallets,
   enableOnramp: true,
+  // Metamask wallet id
   featuredWalletIds: ['c57ca95b47569778a828d19178114f4db188b89b763c899ba0be274e97267d96']
 })
 
--- a/apps/laboratory/src/pages/library/ethers.tsx
+++ b/apps/laboratory/src/pages/library/ethers.tsx
@@ -18,7 +18,8 @@ const modal = createWeb3Modal({
   termsConditionsUrl: 'https://walletconnect.com/terms',
   privacyPolicyUrl: 'https://walletconnect.com/privacy',
   customWallets: ConstantsUtil.CustomWallets,
-  enableOnramp: true
+  enableOnramp: true,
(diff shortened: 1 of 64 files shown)


apps/laboratory/src/pages/library/ethers.tsx
```diff
@@ -18,7 +18,8 @@ const modal = createWeb3Modal({
   termsConditionsUrl: 'https://walletconnect.com/terms',
   privacyPolicyUrl: 'https://walletconnect.com/privacy',
   customWallets: ConstantsUtil.CustomWallets,
-  enableOnramp: true
+  enableOnramp: true,
+  featuredWalletIds: ['c57ca95b47569778a828d19178114f4db188b89b763c899ba0be274e97267d96']
```
lukaisailovic, line 22: Do we wan't to commit this?

packages/scaffold/src/partials/w3m-all-wallets-list/index.ts
```diff
@@ -145,7 +145,11 @@ export class W3mAllWalletsList extends LitElement {
 
   private onConnectWallet(wallet: WcWallet) {
     const { connectors } = ConnectorController.state
-    const connector = connectors.find(({ explorerId }) => explorerId === wallet.id)
+
+    const connector = connectors.find(
+      ({ explorerId, info, id }) =>
+        explorerId === wallet.id || info?.rdns === wallet.rdns || wallet.rdns === id
+    )
```
lukaisailovic, line 152: Can we add a comment here explaining whats happening

packages/wagmi/src/client.ts
```diff
@@ -342,13 +342,13 @@ export class Web3Modal extends Web3ModalScaffold {
...
-      c => c.id === CoreConstants.CONNECTOR_RDNS_MAP[coinbaseSDKId]
+      c => c.id === CoreConstants.CONNECTOR_RDNS_MAP[ConstantsUtil.COINBASE_CONNECTOR_ID]
     )
 
     filteredConnectors.forEach(({ id, name, type, icon }) => {
       // If coinbase injected connector is present, skip coinbase sdk connector.
-      const shouldSkip =
-        (coinbaseConnector && id === coinbaseSDKId) || ConstantsUtil.EMAIL_CONNECTOR_ID === id
+      const isCoinbaseRepeated = coinbaseConnector && id === coinbaseSDKId
+      const shouldSkip = isCoinbaseRepeated || ConstantsUtil.EMAIL_CONNECTOR_ID === id
```
lukaisailovic, line 351: Is this only for Coinbase and MM, or can we make `isCoinbaseRepeated` more modular so it doesn't happen in the future?

apps/laboratory/src/pages/library/ethers.tsx
tomiir: I wanted to use it to test this on lab preview.

packages/wagmi/src/client.ts
tomiir: The issue here is that wagmi specifically injects the coinbase SDK connector, and when the actual extension wallet is installed, it shows both. 
So if the `coinbaseConnector` exists, then we skip the sdk one to prevent duplicates. This doesn't happen on ethers as we instantiate everything ourselves and our implementation recognizes the installation. 

Seems like we could potentially do a PR to wagmi directly as well.

packages/scaffold/src/partials/w3m-all-wallets-list/index.ts
tomiir: I refactored the code to make it simpler but added a comment. Basically either matches by explorer ID or RDNS 🖖

lukaisailovic (approved):

lukaisailovic (approved):

PR #2046, opened 2024-03-13
chore(get-nonce-optional-address): add optional address param to getNonce

Cali93:
# Breaking Changes

N/A

# Changes

- chore: adds optional address argument to the `getNonce` function

# Associated Issues

Request from Cyberconnect


arein (approved):
Tests?

Cali93:
> Tests?
It is a non breaking change.
Adding E2E tests for SIWE should be done in a separate PR.

PR #2053, opened 2024-03-16
fix: missing call to SIWE `onSignOut`

tien:
# Breaking Changes

N/A

# Changes

- fix: missing call to SIWE `onSignOut`

### DIFF
--- a/packages/siwe/src/client.ts
+++ b/packages/siwe/src/client.ts
@@ -111,6 +111,7 @@ export class Web3ModalSIWEClient {
   }
 
   async signOut() {
+    this.methods.onSignOut?.()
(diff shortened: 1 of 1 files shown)


glitch-txs:
The function is already called. Please open an issue if you are facing any problem with SIWE so we can discuss how to fix it. Thanks!

tien:
> The function is already called. Please open an issue if you are facing any problem with SIWE so we can discuss how to fix it. Thanks!

No it's not you can check the signIn function up too has to manually call it

glitch-txs:
Could you open an issue describing the problem and adding steps to reproduce it?

PR #2054, opened 2024-03-16
fix: missing call to SIWE `onSignOut`

tien:
The function is not being called on sign out. See here where `onSignIn` function has to be manually called: https://github.com/WalletConnect/web3modal/blob/581658c308efe7a4a32274cb29000a36d671cda3/packages/siwe/src/client.ts#L105

I've experience this issue with the latest version, `onSignOut` doesn't get called on sign out.

Here's a quick reproduction: https://codesandbox.io/p/devbox/zmlwkg?migrateFrom=vrf9dq

Alert is triggered on sign in, but not on sign out.


tomiir (approved):

glitch-txs (approved):
I misread your prev PR, thought you were calling signOut twice 😄 

LGTM

Cali93 (approved):
Good catch, thanks for the PR!

PR #2060, opened 2024-03-21
fix: reset cursor when opening transaction history views

tomiir:
# Breaking Changes

N/A

# Changes

- feat:
- fix: clear cursor when opening views as they should always show most recent transactions on open
- chore:


arein (dismissed):
Test?

svenvoskamp (approved):

PR #2063, opened 2024-03-22
fix, fix checkInstalled of @web3modal/ethers5 and @web3modal/ethers (…

toccata-yy-lista:
…WalletConnect#2061)

# Breaking Changes

N/A

# Changes

- feat:
- fix:
- chore:

# Associated Issues

closes #...


tomiir:
Looks good! But is missing the wagmi part as well 🙏

PR #2072, opened 2024-03-25
refactor: remove modal open animation delay

enesozturk:
# Breaking Changes

Web3Modal is opening with a 300ms delay, which feels that the app is slow. This PR introduces removing delays and improving the social login approve transaction (iframe) animations as well. 

Also, in CSS, we are using transition property with unnecessary way or without performance optimizations.

# Associated Issues

[Slack thread](https://walletconnect.slack.com/archives/C03RVH94K5K/p1711371083539719)


tomiir:
This works for most cases, but we'd need to do the same thing on the secure-site side. 
If you try to execute a request now, there's a small lag until the secure-site pages actually render and I suspect is because it has it's own animation delay.

tomiir (approved):

tomiir (approved):

PR #2075, opened 2024-03-27
feat(canary): use playwright for duration calculation

arein:
# Breaking Changes

N/A

# Changes

- feat(canary): use playwright for duration calculation
- fix:
- chore:

# Associated Issues

closes #...


lukaisailovic (approved):
I'm not sure if this will also count the duration of the setup as well (fixture). If that is the case, our average duration will be ~20s. Canary will be able to detect if it works/doesn't work but no latency increase. Few seconds will be considered acceptable error.

PR #2080, opened 2024-03-27
chore: update donut contract

lukaisailovic:
Update donut contract so it doesn't have stock. Allows unlimited purchasing of donuts.


tomiir (approved):

enesozturk (approved):
🙏

PR #2081, opened 2024-03-27
chore: log canary times

chris13524:
Logs how long each step takes to help debugging long times.

### DIFF
--- a/apps/laboratory/tests/canary.spec.ts
+++ b/apps/laboratory/tests/canary.spec.ts
@@ -1,3 +1,5 @@
+// eslint-disable-file no-console
+
 import { testConnectedMW } from './shared/fixtures/w3m-wallet-fixture'
 import { uploadCanaryResultsToCloudWatch } from './shared/utils/metrics'
 import { expectConnection } from './shared/utils/validation'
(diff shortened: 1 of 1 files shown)


arein (dismissed):

arein (approved):

apps/laboratory/tests/canary.spec.ts
```diff
@@ -6,7 +6,11 @@ const ENV = process.env['ENVIRONMENT'] || 'dev'
 const REGION = process.env['REGION'] || 'eu-central-1'
 
 testConnectedMW.beforeEach(async ({ modalValidator, walletValidator }) => {
+  // eslint-disable-next-line no-console
```
bkrem, line 9: FYI you can use `eslint-disable-file no-console` at the top here to avoid needing it on each `console.time` call explicitly: https://stackoverflow.com/questions/34764287/turning-off-eslint-rule-for-a-specific-file

bkrem (approved):

apps/laboratory/tests/canary.spec.ts
chris13524: Updated to do this

bkrem (approved):
🏃 ⏲️

Issue #2074, opened 2024-03-26
[bug] WalletConnect Popup Window Transparency Issue When Triggered by Transparent Button


Phillweston:
### Link to minimal reproducible example

[https://lotso.org](https://lotso.org)

### Title:

WalletConnect Popup Window Transparency Issue When Triggered by Transparent Button

### Environment:

* **Operating System**: Windows 10
* **Browser**: Chrome
* **WalletConnect Version**: V4

### Description:

**Summary**: The WalletConnect popup window becomes transparent and unclickable when triggered by a button with transparent styling.

**Details**: When I attempt to activate the WalletConnect modal through a button on my site that has transparent styling in the navigation panel, the popup window inherits this transparency. This makes the popup window content nearly invisible and unclickable, impeding user interaction.

### Steps to Reproduce:

1. Visit [My Website](https://lotso.org).
2. Click on the \['Connect'\] button.
3. Observe that the WalletConnect popup window appears with transparency.

### Expected Behavior:

The WalletConnect popup should display with its default or specified non-transparent styling, ensuring all elements are visible and interactable.

### Actual Behavior:

The WalletConnect popup inherits the transparency of the triggering button, resulting in an almost fully transparent and unclickable popup window.

### Possible Causes/Suggestions:

* CSS inheritance issue where the modal is inheriting the button's transparent styling.
* Lack of isolated styling for the popup modal in the WalletConnect integration.

### Screenshots/Attachments:

![chrome_5xzJZEwHZq](https://uploads.linear.app/53158a94-cce0-42f0-a185-428db7e671e5/54dfc432-7ba0-4574-9a98-1a3b66b69afd/67c2e9ae-45c4-47c1-afdb-6c77d53a50c9?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzUzMTU4YTk0LWNjZTAtNDJmMC1hMTg1LTQyOGRiN2U2NzFlNS81NGRmYzQzMi03YmEwLTQ1NzQtOWE5OC0xYTNiNjZiNjlhZmQvNjdjMmU5YWUtNDVjNC00N2MxLWFmZGItNmM3N2Q1M2E1MGM5IiwiaWF0IjoxNzExNDY4OTkwLCJleHAiOjMzMjgyMDI4OTkwfQ.HTz3dh832KqUEexFkk56AcTjfAEL9oUJoOwkef2m_Ps)

### Additional Information:

[...]

PR #2082, opened 2024-03-27
Force the z-index attribute to avoid display error

Phillweston:
# Breaking Changes

```css
w3m-modal {
        pointer-events: auto;
        z-index: 999 !important;
}
```

# Associated Issues

Related to #2074


glitch-txs:
We have this option as a theming option, I guess we could default it though. In that case this should be added somewhere else in the code.

https://docs.walletconnect.com/web3modal/react/theming

glitch-txs (changes requested):
We have this option as a theming option, I guess we could default it though. In that case this should be added somewhere else in the code.

https://docs.walletconnect.com/web3modal/react/theming

Phillweston:
Therefore we can revise the default value of the z-index from 100 to 999, to ensure it always be shown in the front layer:
```javascript
--w3m-z-index: ${unsafeCSS(themeVariables?.['--w3m-z-index'] || 999)};
```

glitch-txs:
Yes

PR #2090, opened 2024-03-28
imports

glitch-txs:
- remove optionalDependencies object (vue, react, react-dom, siwe)

- added `"@web3modal/siwe#build:siwe",` as dependsOn string for scaffold: peerDeps are not considered by turbo (see [comment](https://github.com/vercel/turbo/issues/1601#issuecomment-1227685519))


packages/core/src/controllers/AccountController.ts
```diff
@@ -1,4 +1,4 @@
-import { subscribeKey as subKey } from 'valtio/utils'
+import { subscribeKey as subKey } from 'valtio/vanilla/utils'
```
enesozturk, line 1: Hmm, @glitch-txs curious, is this the line that removes some dependencies? If yes, that's awesome.

turbo.json
```diff
@@ -27,22 +27,29 @@
     },
     "build:scaffold": {
       "outputs": ["dist/**"],
-      "dependsOn": ["^build:common", "^build:core", "^build:ui", "^build:wallet", "^build:siwe"],
+      "dependsOn": [
+        "^build:common",
+        "^build:core",
+        "^build:ui",
+        "^build:wallet",
+        "@web3modal/siwe#build:siwe",
```
enesozturk, line 35: @glitch-txs Is this because we only want to build `siwe`, not the dependencies of the `siwe`?

enesozturk (approved):

packages/core/src/controllers/AccountController.ts
glitch-txs: The other one requires React because it's also imported internally, so it was throwing an error with yarn because it wasn't installing React. This path works the same way but the React dep is not imported internally.

turbo.json
glitch-txs: This is related to my comment:

> added "@web3modal/siwe#build:siwe", as dependsOn string for scaffold: peerDeps are not considered by turbo (see https://github.com/vercel/turbo/issues/1601#issuecomment-1805589927)

During build turbo wasn't wating for SIWE to be build before starting to build Scaffold which was causing issues, it was because Turbo ignores peerDeps, this fixes it

PR #2093, opened 2024-03-29
chore: configurable secure url

lukaisailovic:
Add an option to configure secure site url so we can implement automated tests for the secure site


enesozturk (approved):

hazelnutcloud:
getting `ReferenceError: process is not defined` in browser because of this change

samueldlacruz:
> Entrando `ReferenceError: process is not defined`en el navegador debido a este cambio.

I have the same error

PR #2099, opened 2024-03-30
add wallet info

glitch-txs:
# Breaking Changes

N/A

# Changes

- feat: Added function and hook to get the connected wallet information (name, icon, etc)

```tsx
import { useWalletInfo } from '@web3modal/wagmi/react


export default App(){
  const { walletInfo } = useWalletInfo()

  console.log(walletInfo.name, walletInfo.icon)
  
  return <YourApp/>

}
```

```ts
modal.subscribeWalletInfo(handler)

const { name, icon } = modal.getWalletInfo()
```

# Associated Issues

closes https://github.com/WalletConnect/web3modal/issues/1665


packages/ethers/src/client.ts
```diff
@@ -924,6 +919,39 @@ export class Web3Modal extends Web3ModalScaffold {
     }
   }
 
+  private syncConnectedWalletInfo() {
+    const currentActiveWallet = window?.localStorage.getItem(EthersConstantsUtil.WALLET_ID)
+    const providerType = EthersStoreUtil.state.providerType
+
+    if (providerType === ConstantsUtil.EIP6963_CONNECTOR_ID) {
```
enesozturk, line 926: Seems like there is no else state for the inner condition;
```suggestion
    if (providerType === ConstantsUtil.EIP6963_CONNECTOR_ID && currentActiveWallet) {
```

packages/ethers5/src/client.ts
```diff
@@ -485,10 +480,10 @@ export class Web3Modal extends Web3ModalScaffold {
     const currentActiveWallet = window?.localStorage.getItem(EthersConstantsUtil.WALLET_ID)
     if (currentActiveWallet) {
       const currentProvider = this.EIP6963Providers.find(
-        provider => provider.name === currentActiveWallet
+        provider => provider.info.name === currentActiveWallet
```
tomiir, line 483: is this always the case? is there a case where info is not serialized? eg. non-6963 wallets?

tomiir (approved):

PR #2104, opened 2024-04-02
fix: reset sa enabled networks

tomiir:
# Breaking Changes

N/A

# Changes

- feat:
- fix: ethers `disconnect` not being awaited. Issue where a SA user could disconnect, connect with a non-enabled account and SA flow would be prompted due to network being enabled, ending in error.
- chore: add corresponding UI test and controller test. Add disconnect UI test.

### DIFF
--- a/apps/laboratory/tests/smart-account.spec.ts
+++ b/apps/laboratory/tests/smart-account.spec.ts
@@ -2,6 +2,8 @@ import { testModalSmartAccount } from './shared/fixtures/w3m-smart-account-fixtu
 import type { ModalWalletPage } from './shared/pages/ModalWalletPage'
 import type { ModalWalletValidator } from './shared/validators/ModalWalletValidator'
 
+const NOT_ENABLED_SMART_ACCOUNT_INDEX = 10
+
 testModalSmartAccount.beforeEach(async ({ modalValidator }) => {
   await modalValidator.expectConnected()
 })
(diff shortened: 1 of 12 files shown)


glitch-txs (approved):

apps/laboratory/tests/smart-account.spec.ts
```diff
@@ -45,8 +47,26 @@ testModalSmartAccount(
...
+
+testModalSmartAccount(
+  'it should use an eoa when disconnecting and connecting to a not enabled address',
+  async ({ modalPage, modalValidator, context }) => {
+    const walletModalPage = modalPage as ModalWalletPage
+    const walletModalValidator = modalValidator as ModalWalletValidator
+
+    await walletModalPage.togglePreferredAccountType()
+    await walletModalPage.disconnect()
+    await walletModalPage.page.waitForTimeout(2500)
```
lukaisailovic, line 64: Is this absolutely required?

apps/laboratory/tests/smart-account.spec.ts
```diff
@@ -45,8 +47,26 @@ testModalSmartAccount(
...
+
+    await walletModalPage.emailFlow('web3modal-smart-account@mailsac.com', context)
```
lukaisailovic, line 66: Extract to const

apps/laboratory/tests/smart-account.spec.ts
tomiir: it was preventing the magic disconnect promise from resolving, but shoould work without it.
will push fix in a diff pr if possible

PR #2110, opened 2024-04-03
chore: add conditional logs for canary

lukaisailovic:
Logs (console.time) are now optional and turned off by default

### DIFF
--- a/apps/laboratory/tests/shared/utils/logs.ts
+++ b/apps/laboratory/tests/shared/utils/logs.ts
@@ -1,0 +1,17 @@
+/* eslint no-console: 0 */
+
+const TIMING_LOGS_ENABLED = process.env['TIMING_LOGS'] === 'true' || false
+
+export function timeStart(label?: string | undefined) {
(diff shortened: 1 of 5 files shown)


apps/laboratory/tests/shared/utils/logs.ts
```diff
@@ -0,0 +1,17 @@
+/* eslint no-console: 0 */
+
+const TIMING_LOGS_ENABLED = process.env['TIMING_LOGS'] === 'true' || false
+
+export function timeStart(label?: string | undefined) {
```
chris13524, line 5: Why do you allow undefined? I feel like this should just be `label: string`

chris13524 (approved):

enesozturk (approved):

apps/laboratory/tests/shared/utils/logs.ts
lukaisailovic: Following the definition of `console.time`, you can actually just call it like that without any param

PR #2116, opened 2024-04-04
feat: convert

enesozturk:
# Changes

This PR introduces the convert feature along with many changes:

- Add new composites and stories
- Refactor composites
- Add convert views `w3m-convert-view`, `w3m-convert-preview-view`, and `w3m-convert-select-token-view`.
- `ConvertController` to handle all states of the convert feature: balances, tokens, creating convert call data, creating allowance, making calculations, etc.
- Refactor `RouterController` to handle the approve transaction view working with opening both from W3M and outside of the modal. 
- Add `NumberUtil` with a new bignumber package to handle calculations of the numbers which have big decimals.

# Associated Issues

- [Linear Project](https://linear.app/walletconnect/project/convert-swaps-c87ad3336fd7?layout=list&ordering=priority&grouping=workflowState&subGrouping=none&showCompletedIssues=all&showSubIssues=true)


packages/ethers/src/client.ts
```diff
@@ -273,6 +278,67 @@ export class Web3Modal extends Web3ModalScaffold {
...
+
+      formatUnits: (value: bigint, decimals: number) => formatUnits(value, decimals),
+
+      async getEstimatedGas(data) {
+        const chainId = EthersStoreUtil.state.chainId
+        const provider = EthersStoreUtil.state.provider
+        const address = EthersStoreUtil.state.address
+
+        if (!provider) {
+          throw new Error('connectionControllerClient:sendTransaction - provider is undefined')
```
lukaisailovic, line 293: Should we have custom errors?

packages/core/src/utils/ConstantsUtil.ts
```diff
@@ -61,7 +61,101 @@ export const ConstantsUtil = {
...
+    'BAL',
+    'BICO',
+    'CRV',
+    'ENS',
+    'MATIC',
+    'OP'
+  ],
+
+  POPULAR_TOKENS: [
+    'ETH',
```
enesozturk, line 97: Some list of tokens to show for the demo purpose. Waiting Lit's virtualize SSR support for listing all possible tokens in the performant way.

- https://github.com/lit/lit/pull/4564

enesozturk, line 97: Some list of tokens to show for the demo purpose. Waiting Lit's virtualize SSR support for listing all possible tokens in the performant way.

- https://github.com/lit/lit/pull/4564

packages/core/src/utils/ConvertApiUtil.ts
```diff
@@ -0,0 +1,330 @@
+import { CoreHelperUtil } from '../utils/CoreHelperUtil.js'
```
enesozturk, line 1: Once we have all endpoints ready in blockchain API, this Util will be removed.

packages/core/src/utils/CoreHelperUtil.ts
```diff
@@ -54,6 +54,16 @@ export const CoreHelperUtil = {
     return caipAddress.split(':')[2]
   },
 
+  getEvmChainId(caipNetworkId?: `${string}:${string}`) {
```
enesozturk, line 57: Replace with `common/NetworkUtil/caipNetworkIdToNumber`?

packages/ethers/src/client.ts
enesozturk: Wdym by 'custom errors'? Having better error messages? @lukaisailovic

packages/solana/src/client.ts
```diff
@@ -140,7 +140,19 @@ export class Web3Modal extends Web3ModalScaffold {
         })
 
         return signature as string
-      }
+      },
+
+      sendTransaction: async () => {
```
enesozturk, line 145: These are just to handle type errors. It's not available in Solana right now and won't be used. Maybe we could handle these better?

packages/wagmi/src/client.ts
```diff
@@ -168,7 +174,48 @@ export class Web3Modal extends Web3ModalScaffold {
...
+          to: data.to,
+          value: data.value,
+          gas: data.gas,
+          gasPrice: data.gasPrice,
+          data: data.data,
+          chainId,
+          type: 'legacy' as const
+        }
+
+        // eslint-disable-next-line @typescript-eslint/prefer-ts-expect-error, @typescript-eslint/ban-ts-comment
```
enesozturk, line 206: Handle the type errors and remove `ts-ignore`

packages/core/src/utils/ConstantsUtil.ts
tomiir: Would this be something that we could store in api.web3modal maybe?

tomiir: same

packages/core/src/utils/ConvertApiUtil.ts
```diff
@@ -0,0 +1,330 @@
...
+  async searchTokens(searchTerm: string) {
+    const { api, paths } = this.get1InchAPI()
+
+    return await api.get<TokenInfo[]>({
+      path: paths.search,
+      params: { query: searchTerm }
+    })
+  },
+
+  async getMyTokensWithBalance() {
```
tomiir, line 189: isn't this available in portfolio endpoint cc @geekbrother @svenvoskamp ?

packages/scaffold/src/views/w3m-convert-preview-view/index.ts
```diff
@@ -0,0 +1,228 @@
...
+    return html`
+      <w3m-convert-details
+        detailsOpen=${this.detailsOpen}
+        sourceTokenSymbol=${this.sourceToken?.symbol}
+        sourceTokenPrice=${this.sourceTokenPriceInUSD}
+        toTokenSymbol=${this.toToken?.symbol}
+        toTokenConvertedAmount=${toTokenConvertedAmount}
+        gasPriceInUSD=${formatNumberToLocalString(this.gasPriceInUSD, 3)}
+        .priceImpact=${this.priceImpact}
+        slippageRate=${0.5}
```
tomiir, line 197: is this ok? isn't this value in controller state?

packages/scaffold/src/views/w3m-convert-select-token-view/index.ts
```diff
@@ -0,0 +1,261 @@
...
+                      ?disabled=${selected}
+                      symbol=${tokenInfo.symbol}
+                      price=${tokenInfo?.price}
+                      amount=${tokenInfo?.balance}
+                      imageSrc=${tokenInfo.logoURI}
+                      @click=${() => {
+                        if (!selected) {
+                          this.onSelectToken(tokenInfo)
+                        }
+                      }}
```
tomiir, line 147: you could avoid inlining this by currying the parameter and adding selected there:

```
const onSelectToken = (tokenInfo: TokenInfo, selected: boolean) => () => {...}

@click={this.onSelectToken(tokenInfo, selected)}
```

packages/scaffold/src/views/w3m-convert-view/index.ts
```diff
@@ -0,0 +1,327 @@
...
+          if (!newRoute.includes('Convert')) {
+            ConvertController.resetValues()
+          }
+        }),
+        ConvertController.subscribeKey('sourceToken', newSourceToken => {
+          this.sourceToken = newSourceToken
+        }),
+        ConvertController.subscribeKey('toToken', newToToken => {
+          this.toToken = newToToken
+        }),
```
tomiir, line 89: why subscribe these separately?

packages/scaffold/src/views/w3m-convert-view/index.ts
```diff
@@ -0,0 +1,327 @@
...
+        ConvertController.subscribe(newState => {
+          this.initialized = newState.initialized
+          this.loading = newState.loading
+          this.loadingPrices = newState.loadingPrices
+          this.transactionLoading = newState.transactionLoading
+          this.sourceToken = newState.sourceToken
+          this.sourceTokenAmount = newState.sourceTokenAmount
+          this.sourceTokenPriceInUSD = newState.sourceTokenPriceInUSD
+          this.toToken = newState.toToken
```
tomiir, line 98: and why subscribe two times? isn't it already covered in `subscribeKey`?

packages/solana/src/client.ts
tomiir: I would at least leave a comment for future reviewers. 
Technically this shouldn't even be here, what type error is requiring it?

packages/wagmi/src/client.ts
tomiir: what is the error here?

packages/wagmi/src/client.ts
```diff
@@ -449,8 +502,33 @@ export class Web3Modal extends Web3ModalScaffold {
...
+          if (super.isTransactionStackEmpty()) {
+            super.close()
+          } else {
+            super.popTransactionStack(true)
+          }
+        }
+
+        const isPayloadString = typeof payload === 'string'
+        const isAddress = isPayloadString ? payload?.startsWith('0x') : false
+        const isCompleted = isAddress && payload?.length > 10
```
tomiir, line 523: what happens with responses that are not 0x strings?

packages/core/src/utils/ConvertApiUtil.ts
```diff
@@ -0,0 +1,330 @@
...
+export type GetConvertDataResponse = {
+  toAmount: string
+  tx: TransactionData
+}
+
+export type GetGasPricesResponse = {
+  baseFree: string
+  low: {
+    maxPriorityFeePerGas: string
+    maxFeePerGas: string
```
tomiir: can we abstract this to `type GasPrice`?

packages/core/src/utils/ConvertApiUtil.ts
```diff
@@ -0,0 +1,330 @@
...
+        approveAllowance: endpoints.approveAllowance,
+        gas: endpoints.gasPrice,
+        gasPrice: endpoints.gasPrice,
+        swap: endpoints.swap,
+        tokens: endpoints.tokens,
+        tokensCustom: endpoints.tokensCustom,
+        tokenPrices: endpoints.tokensPrices,
+        search: endpoints.search,
+        balance: endpoints.balance,
+        quote: endpoints.quote
```
tomiir: couldn't you do 
```
paths: { ...endpoints }

// or

paths: get1InchEndpoints(chainId, address)
```
?

packages/ethers/src/client.ts
tomiir: Custom error as in extending the error class eg.
```
class ConnectionControllerError extends Error {
...
}
```

I think it'd be a nice addition to standardize errors but I'd do it globally  (for all errors in w3m)

packages/scaffold/src/partials/w3m-account-tokens-widget/index.ts
```diff
@@ -39,6 +39,10 @@ export class W3mAccountTokensWidget extends LitElement {
   }
 
   // -- Private ------------------------------------------- //
+  private watchConvertValues() {
+    this.watchTokenBalance = setInterval(() => AccountController.fetchTokenBalance(), 1000)
+  }
```
tomiir: won't this spam the api causing a lot of requests to our balance provider? cc @geekbrother

packages/ui/src/utils/Math.ts
```diff
@@ -0,0 +1,30 @@
+/**
```
tomiir, line 1: any way to test this?

packages/ui/src/utils/NumberUtil.ts
```diff
@@ -0,0 +1,23 @@
+/**
```
tomiir, line 1: any way to test this? maybe already covered by the controller test?

packages/wagmi/src/client.ts
```diff
@@ -165,7 +171,48 @@ export class Web3Modal extends Web3ModalScaffold {
         await disconnect(this.wagmiConfig)
       },
 
-      signMessage: async message => signMessage(this.wagmiConfig, { message })
+      signMessage: async message => signMessage(this.wagmiConfig, { message }),
+
+      getEstimatedGas: async args => {
```
tomiir, line 176: I'd maintain naming with `estimateGas`

packages/scaffold/src/partials/w3m-account-tokens-widget/index.ts
geekbrother: Since we have a big TTL for the Zerion cache, we should increase this by at least 5 secs.

packages/core/src/controllers/ConvertController.ts
```diff
@@ -0,0 +1,687 @@
...
+
+  // Request values
+  slippage: ConstantsUtil.CONVERT_SLIPPAGE_TOLERANCE,
+
+  // Tokens
+  tokens: undefined,
+  popularTokens: undefined,
+  suggestedTokens: undefined,
+  foundTokens: undefined,
+  myTokensWithBalance: undefined,
```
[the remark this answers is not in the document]
enesozturk: I think yeah. When we don't define undefined values in the initial state, I feel like it might be confusing, it gives more confidence when we see all of the states in a place. I do think we should keep them.

packages/core/src/utils/ConstantsUtil.ts
tomiir: Can the list be fetched api side?

packages/core/src/controllers/ConvertController.ts
```diff
@@ -0,0 +1,687 @@
...
+    if (target === 'sourceToken') {
+      state.sourceTokenPriceInUSD = price
+    } else if (target === 'toToken') {
+      state.toTokenPriceInUSD = price
+    }
+  },
+
+  switchTokens() {
+    const newSourceToken = state.toToken ? { ...state.toToken } : undefined
+    const newToToken = state.sourceToken ? { ...state.sourceToken } : undefined
```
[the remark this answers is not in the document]
enesozturk: It would work, but with issues. Due to JavaScript's nature, assigning one variable which points to an object to another variable is not the same when assigning variables which references to primitive values like `string`, `number`, and `boolean`.

JavaScript working with pointers and when we assign one object to another, we are actually assigning a reference to the first object. This means both variables point to the same memory location. Consequently, if you modify the object through one variable, the changes will be reflected when accessing it through the other variable. This can lead to unintended side effects, especially in complex applications where it might be hard to track which parts of your code are modifying the object.

Destructuring, on the other hand, is useful because it allows us to create a new object with the same properties but with a separate reference. This avoids the issues associated with direct object references. When you destructure and assign the properties of one object to another new object, you are essentially copying the values. For primitives, this is straightforward, and for objects, if done carefully, it can help ensure that the objects don't share the same reference.

packages/core/src/utils/ConstantsUtil.ts
enesozturk: Here, I'd like to not filter the possible tokens, instead list all of them. But to do this performantly, we need to use `lit-virtualizer` which is a package to render lots of items without performance issues.

But we have some SSR issues on lit-virtualizer side which is resolved but not merge yet: https://github.com/lit/lit/pull/4564

packages/core/src/utils/ConvertApiUtil.ts
enesozturk: Yeah, we switched to the same balance endpoint now.

packages/scaffold/src/partials/w3m-account-tokens-widget/index.ts
```diff
@@ -29,10 +29,6 @@ export class W3mAccountTokensWidget extends LitElement {
     this.unsubscribe.forEach(unsubscribe => unsubscribe())
   }
 
-  public override firstUpdated() {
```
enesozturk, line 32: After completing convert. We should watch for the changes. I thought we could just refetch the balance after convert is done, but we are not sure if the balances will be updated right after the convert.

So instead, I wanted to fetch balances periodically and stop fetching when the screen is unmounted in `w3m-account-wallet-features-widget`

packages/solana/src/client.ts
enesozturk: If I remove them here, I need to make the `sendTransaction`, `parseUnits` and `formatUnits` methods as optional in the ConnectionController.

This would force us to make a type check everytime when we use these methods, like `ConnectionController?.sendTransaction(..)`, or even we need to me the check on the top level of the function we are using them:

```js
function createConvertTransaction() {
 const availableToSendTransaction = !!ConnectionController?.sencTransaction
...
}
````

etc. It depends on us that we could both make them optional methods or use empty functions on the Solana client to not raise optional property access things.

It's matter the answer of: 
- "Will we have the same features in the Solana?"
- "If yes, is the Solana methods working the same way EVM works?" (I guess the answer is no)
- "If not, is it worth make the other client's work harder? etc."

We should double check these with the Boosty tea, but I do believe it's too early to consider wallet features in Solana network for now.

File: packages/wagmi/src/client.ts (hunk 12 above)
Comment (svenvoskamp) [reply packages/wagmi/src/client.ts:516]: +1 on this. This feels a bit hacky and unreliable.

File: packages/solana/src/client.ts (hunk 5 above)
Comment (svenvoskamp) [reply packages/solana/src/client.ts]: I think this is because the client requires this right?

File: packages/scaffold/src/client.ts [hunk 22]
```diff
@@ -83,6 +85,22 @@ export class Web3ModalScaffold {
     ModalController.close()
   }
 
+  public redirect(route: RouterControllerState['view']) {
```
svenvoskamp, line 88: Why do these need to be public?

packages/ui/src/utils/NumberUtil.ts
enesozturk: Done!

packages/solana/src/client.ts
enesozturk: Totally. ConnectorController is the place we are using the same methods higher level on the UI side or in other controllers. So we are making the API same for all clients and use them in same place.

packages/wagmi/src/client.ts
enesozturk: @svenvoskamp @tomiir I think we could make this by checking the string if it's transaction has or not: https://github.com/WalletConnect/web3modal/pull/2116/commits/5f6f129610911ed4b2d1b24709fe63839aff9639

packages/ethers/src/client.ts
enesozturk: @tomiir would be a nice addition but since we are following the same pattern for all error messages in the clients, let's set this as a error management improvement task. Created: https://linear.app/walletconnect/issue/W3M-257/improve-error-management-in-the-w3m-clients

packages/core/src/utils/ConstantsUtil.ts
enesozturk: Created a post merge ticket for this: https://linear.app/walletconnect/issue/W3M-258/post-merge-improvements-for-convert

packages/core/src/controllers/ConvertController.ts
```diff
@@ -0,0 +1,687 @@
...
+    const prices = await ConvertApiUtil.getTokenPriceWithAddresses([
+      ConstantsUtil.NATIVE_TOKEN_ADDRESS
+    ])
+    const price = prices[ConstantsUtil.NATIVE_TOKEN_ADDRESS] || '0'
+    state.tokensPriceMap[ConstantsUtil.NATIVE_TOKEN_ADDRESS] = price
+    state.networkPrice = price
+  },
+
+  async getMyTokensWithBalance() {
+    const res = await ConvertApiUtil.getMyTokensWithBalance()
```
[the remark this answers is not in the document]
enesozturk: We are now using the same balance endpoint that we use on the home page

packages/wallet/src/W3mFrameHelpers.ts
```diff
@@ -67,6 +72,28 @@ export const W3mFrameHelpers = {
     return (request as { payload: W3mFrameTypes.RPCRequest })?.payload?.method
   },
 
+  getResponseType(response: unknown) {
```
enesozturk, line 75: Here I'm checking both type and payload and returning a string of possible RPC response types that I defined in `W3mFrameConstants`. This would allow us to use less logics in the client and take actions depending on the response type (error, tx hash, or any other thing)

packages/wagmi/src/client.ts
```diff
@@ -493,32 +493,32 @@ export class Web3Modal extends Web3ModalScaffold {
...
-            super.popTransactionStack(true)
-          }
-        }
+      provider.onRpcResponse(response => {
+        const responseType = W3mFrameHelpers.getResponseType(response)
 
-        const isPayloadString = typeof payload === 'string'
-        const isAddress = isPayloadString ? payload?.startsWith('0x') : false
-        const isCompleted = isAddress && payload?.match(RegexUtil.transactionHashRegex)
+        switch (responseType) {
```
enesozturk, line 499: Once we have more cases other than handling RPC request error or tx hash, we could have add to this switch

svenvoskamp (approved):
LGTM ✨

PR #2118, opened 2024-04-04
fix: get isConnected from provider

ignaciosantise:
# Breaking Changes
N/A

# Changes
- fix: using `isConnected` value from the provider

# Associated Issues

closes #...


tomiir (approved):

arein (approved):
Can we add a test to cover regressions?

glitch-txs (approved):
El que sabe sabe

tomiir:
@arein I don't think this can be tested in e2e consistently as it ends up excuting the flow and in correct state (tries to get user, fails, show unconnected).

Feels like this should be unit tested, we should unit test (all) clients during this tech debt week

PR #2119, opened 2024-04-05
feat: support custom connectors

chris13524:
# Breaking Changes

N/A

# Changes

- feat: support custom connectors
- chore: make return types consistent
- chore: missing and typo comments


glitch-txs (commented):
Mmm, the idea of defaultWagmiConfig is to save dev time, we intentionally prevent connectors to be overriden to make sure WalletConnectConnector was set correctly. For custom connector devs can always use the usual Wagmi configuration without using the default function.

Also this seem like it might add some connectors twice if the dev doesn't know what's internally

chris13524:
@glitch-txs There's a lot of stuff this function sets up which `defaultWagmiConfig()` drastically simplifies: WalletConnect, email wallet, RPC URLs, EIP-6963. Lot of lines of code for a dev to re-implement if they just want a new connector, and it wouldn't use our RPC out of the box.

To avoid duplicates we can detect if a connector ID already exists in the list. If it already exists, don't add. That should solve your concerns?

glitch-txs:
That'd be better I guess.

As a point aside, I think we should also abstract the Blockchain API implementation so it could be used without this function being required.

tomiir (approved):

tomiir:
I'd filter out duplicates and leave it as-is. But this should be included in docs @glitch-txs @boidushya to prevent confusion

chris13524:
[Slack conversation](https://walletconnect.slack.com/archives/C03RVH94K5K/p1712284368781579)

PR #2120, opened 2024-04-05
feat/convert wui components

enesozturk:
# Changes

This PR introduces the new `ui` components and refactors for some of the existing components along with some required utility changes.

- `common/NumberUtil` has been added with `bignumber.js` package to make calculations with numbers which have long decimals.
- `core/TypeUtils` has been updated to add new Convert feature related events.
- `ui/utils/NumberUtil` has been added to render numbers with a human readable format.

N/A

# Changes

- feat:
- fix:
- chore:

# Associated Issues

closes #...


packages/ui/src/composites/wui-convert-details/index.ts
svenvoskamp, line 1: This file is very large and specific to the `convert` feature. We might want to move this to partials as it is feature specific.

packages/ui/src/composites/wui-convert-input/index.ts
svenvoskamp, line 1: It is not allowed to use core in the ui package. We should move this one to partials as well as it is feature specific and needs logic.

svenvoskamp (changes requested):
Awesome job! But could you please move the 2 feature specific composites to partials as this makes more sense to me.

svenvoskamp (approved):

packages/ui/src/utils/UiHelperUtil.ts
```diff
@@ -119,5 +119,28 @@ export const UiHelperUtil = {
...
+      return value.toLocaleString('en-US', {
+        maximumFractionDigits: decimals,
+        minimumFractionDigits: decimals
+      })
+    }
+
+    return parseFloat(value).toLocaleString('en-US', {
+      maximumFractionDigits: decimals,
+      minimumFractionDigits: decimals
+    })
```
tomiir, line 144: isnt this repeated in `NumberUtils`?

enesozturk: It should be removed, let me check

packages/ui/src/components/wui-image/styles.ts
```diff
@@ -3,8 +3,8 @@ import { css } from 'lit'
 export default css`
   :host {
     display: block;
-    width: 100%;
-    height: 100%;
+    width: var(--local-width);
+    height: var(--local-height);
```
tomiir, line 7: just checking this is controlled by ` wui-image` directly and it doesn't break other things?

packages/ui/src/utils/UiHelperUtil.ts
enesozturk: Good catch, we have it more than one, removed the others!

packages/scaffold/src/partials/w3m-convert-input/index.ts
```diff
@@ -0,0 +1,226 @@
...
+        ? html`<wui-text variant="small-400" color="fg-200">
+            ${UiHelperUtil.formatNumberToLocalString(this.balance, 3)}
+          </wui-text>`
+        : null}
+      ${this.target === 'sourceToken' ? this.tokenActionButtonTemplate(haveBalance) : null}
+    `
+  }
+
+  private tokenActionButtonTemplate(_haveBalance: boolean) {
+    if (_haveBalance) {
```
tomiir, line 198: why the underscore arg? That usually means unused

packages/ui/src/components/wui-image/styles.ts
enesozturk: Exactly, all seem good. Thats good point that it might affect thousands of UIs :)

packages/common/src/utils/NumberUtil.ts
```diff
@@ -0,0 +1,49 @@
...
+      // eslint-disable-next-line new-cap
+      return BigNumber(0)
+    }
+
+    const aBigNumber = new BigNumber(a)
+    const bBigNumber = new BigNumber(b)
+
+    return aBigNumber.multipliedBy(bBigNumber)
+  }
+}
```
tomiir, line 25: I think I asked this in a diff pr but why not 
`BigInt(a) * BigInt(b)`?

packages/scaffold/src/partials/w3m-convert-input/index.ts
enesozturk: It was because of the conflict with a variable used with same name in the class. But for this one, we don't actually need it. Good call, removed

packages/common/src/utils/NumberUtil.ts
enesozturk: BigNumber is not the same as BigInt. We are using bignumber.js for calculations. This library is capable of making calculations with tiny numbers without messing with BigInt even though the number itself has more than 18 decimals. 

BigNumber have provides functional programming approach for running calculations. I think I might create a doc about why we need / why we choose it.

tomiir (approved):

enesozturk:
@svenvoskamp @tomiir The Dangerfile results with the error above. I double checked our exports and compared line by line. So it's a misleading error. Created an issue for it [here](https://linear.app/walletconnect/issue/W3M-180/[improvement]-dangerjs-returning-wrong-conclusion-at-some-checks).

Same for the e2e tests, doesn't related to this PR.

PR #2122, opened 2024-04-05
feat: WC Names integration

tomiir:
# Breaking Changes

N/A

# Changes

- feat: WC Names integration.
- fix:
- chore:

# Associated Issues

Closes https://linear.app/walletconnect/issue/W3M-39/profile-name-selection


svenvoskamp (changes requested):
Looks very nice! Left a few small comments.

packages/ui/src/composites/wui-ens-input/index.ts
```diff
@@ -0,0 +1,80 @@
...
+  @property({ type: Boolean }) public disabled = false
+
+  @property() public value?: string
+
+  @property({ type: Boolean }) public loading = false
+
+  // -- Render -------------------------------------------- //
+  public override render() {
+    return html`
+      <input
```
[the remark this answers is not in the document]
tomiir: needed some custom styling that was over complicated if using `wui-input-text` but trying to work around it 🖖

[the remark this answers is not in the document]
tomiir: Added an `inputRightPadding` prop to `wui-input-text` which I'm not convinced is the cleanest, but allowed me to remove the custom styling here

packages/core/tests/controllers/EnsController.test.ts
```diff
@@ -0,0 +1,45 @@
...
+
+        return Promise.resolve(TEST_NAME)
+      }
+    }
+  }
+})
+
+// -- Tests --------------------------------------------------------------------
+describe('EnsController', () => {
+  it('should have valid default state', () => {
```
enesozturk, line 39: Seems we don't have many tests covered. We have functions like  `validateName`,  `parseEnsApiError` that could be unit tested

packages/core/src/controllers/EnsController.ts
```diff
@@ -0,0 +1,161 @@
...
+
+      const coinType = EnsUtil.convertEVMChainIdToCoinType(networkId)
+      await BlockchainApiController.registerEnsName({
+        coinType,
+        address: address as `0x${string}`,
+        signature,
+        message
+      })
+
+      AccountController.setProfileName(`${name}.wcn.id`)
```
enesozturk, line 143: Will we still keep the `wcn.id` when we go to prod?

packages/core/src/controllers/EnsController.ts
```diff
@@ -0,0 +1,161 @@
...
+      RouterController.replace('RegisterAccountNameSuccess')
+      state.loading = false
```
enesozturk, line 145: Maybe we could move `state.loading = false` to `finally` to avoid duplicate?

packages/core/src/utils/FetchUtil.ts
```diff
@@ -14,6 +14,19 @@ interface PostArguments extends RequestArguments {
   body?: Record<string, unknown>
 }
 
+async function fetchData(...args: Parameters<typeof fetch>) {
+  const response = await fetch(...args)
+  if (!response.ok) {
+    // Create error object and reject if not a 2xx response code
+    const err = new Error(`HTTP status code: ${response.status}`, {
```
enesozturk, line 21: Why do we do this? Is it for formalizing the errors? Are we sure it's not breaking any error handling of blockchain calls?

packages/scaffold/src/partials/w3m-account-wallet-features-widget/index.ts
```diff
@@ -97,10 +97,11 @@ export class W3mAccountWalletFeaturesWidget extends LitElement {
       <wui-profile-button
         @click=${this.onProfileButtonClick.bind(this)}
         address=${ifDefined(this.address)}
+        profileName=${ifDefined(this.profileName)}
         networkSrc=${ifDefined(networkImage)}
         icon="chevronBottom"
         avatarSrc=${ifDefined(this.profileImage ? this.profileImage : undefined)}
-        ?isprofilename=${Boolean(this.profileName)}
+        ?isProfileName=${Boolean(this.profileName)}
```
enesozturk, line 104: This should be changed in https://github.com/WalletConnect/web3modal/pull/2276/files, `isProfileName` shouldn't be needed

packages/scaffold/src/views/w3m-register-account-name-view/index.ts
```diff
@@ -0,0 +1,188 @@
...
+          this.suggestions = val.suggestions
+          this.loading = val.loading
+        })
+      ]
+    )
+  }
+
+  // -- Lifecycle ----------------------------------------- //
+  public override firstUpdated() {
+    this.formRef.value?.addEventListener('keydown', event => {
```
enesozturk, line 43: Don't we need to remove listener when unmount (on `disconnect()`)?

packages/scaffold/src/views/w3m-register-account-name-view/index.ts
```diff
@@ -0,0 +1,188 @@
...
+        justifyContent="space-between"
+        class="suggestion"
+      >
+        <wui-text color="fg-100" variant="paragraph-400" class="suggested-name">
+          ${this.name}</wui-text
+        >${this.nameSuggestionTagTemplate()}
+      </wui-flex>
+      ${suggestions
+        .filter(s => s.name !== this.name)
+        .map(suggestion => this.availableNameTemplate(suggestion.name))}
```
enesozturk, line 150: I think even though it's a simple filtering, makings any filter/sort/reduce on outside of the render would be a nice approach to keep things more readable

```suggestion
const filteredNames = suggestions.filter(s => s.name !== this.name)

      ${filteredNames.map(suggestion => this.availableNameTemplate(suggestion.name))}
```

packages/scaffold/src/views/w3m-register-account-name-view/styles.ts
```diff
@@ -0,0 +1,34 @@
...
+    border-radius: var(--wui-border-radius-xs);
+  }
+
+  .suggestion:hover {
+    background-color: var(--wui-gray-glass-005);
+    cursor: pointer;
+  }
+
+  .suggested-name {
+    max-width: 220px;
```
enesozturk, line 19: Is this value also applicable for `mobile` or smaller screens? Is it possible to use relative units?

packages/scaffold/src/views/w3m-upgrade-to-smart-account-view/index.ts
```diff
@@ -27,7 +28,11 @@ export class W3mUpgradeToSmartAccountView extends LitElement {
         .padding=${['0', '0', 'l', '0'] as const}
       >
         ${this.onboardingTemplate()} ${this.buttonsTemplate()}
-        <wui-link>
+        <wui-link
+          @click=${() => {
+            CoreHelperUtil.openHref('https://walletconnect.com/faq', '_blank')
```
enesozturk, line 33: Could we create a `Navigation` unit/constant in `@common` and use it here like `Navigation.WEBSITE_FAQ` to avoid hard coded values? This would allow us to maintain all URLs or other sources/hard coded values in one place.

packages/ui/src/composites/wui-profile-button/index.ts
```diff
@@ -40,10 +42,10 @@ export class WuiProfileButton extends LitElement {
         <wui-flex gap="xs" alignItems="center">
           <wui-text variant="large-600" color="fg-100">
             ${UiHelperUtil.getTruncateString({
-              string: this.address,
-              charsStart: this.isProfileName ? 18 : 4,
-              charsEnd: this.isProfileName ? 0 : 4,
-              truncate: this.isProfileName ? 'end' : 'middle'
+              string: this.profileName ?? this.address,
```
enesozturk, line 45: These should be handled in https://github.com/WalletConnect/web3modal/pull/2276/files

packages/core/src/controllers/EnsController.ts
```diff
@@ -0,0 +1,161 @@
...
+    if (!network) {
+      throw new Error('Network not found')
+    }
+    const address = AccountController.state.address
+    const emailConnector = ConnectorController.getAuthConnector()
+    if (!address || !emailConnector) {
+      throw new Error('Address or auth connector not found')
+    }
+    state.loading = true
+    const message = JSON.stringify({
```
enesozturk, line 112:
```suggestion
    const network = NetworkController.state.caipNetwork
    const emailConnector = ConnectorController.getAuthConnector()
    
    if (!network) {
      throw new Error('Network not found')
    }
    
     if (!address || !emailConnector) {
      throw new Error('Address or auth connector not found')
    }
    
    const address = AccountController.state.address
    state.loading = true
    
    const message = JSON.stringify({
```

packages/core/src/controllers/BlockchainApiController.ts
```diff
@@ -242,6 +245,40 @@ export const BlockchainApiController = {
...
   },
 
+  async lookupEnsName(name: string) {
+    return api.get<BlockchainApiLookupEnsName>({
+      path: `/v1/profile/account/${name}.wcn.id?projectId=${OptionsController.state.projectId}`
+    })
+  },
+
+  async reverseLookupEnsName({ address }: { address: string }) {
+    // Should add /coinType subroute once ready in BlockchainApi
```
enesozturk, line 255: Should be removed as confirmed https://walletconnect.slack.com/archives/C06L6HUN1H6/p1715937712951339?thread_ts=1715932794.012259&cid=C06L6HUN1H6

enesozturk (commented):
Left some comments

packages/core/src/controllers/EnsController.ts
tomiir: don't really understand the suggested change, this code will break since address is not defined 🤔

packages/core/src/utils/FetchUtil.ts
tomiir: FetchUtil was not throwing when receiving failed requests (neither 500 nor 400)
Wondering how this was working for you on swaps controller for instance 🤔

packages/core/src/controllers/EnsController.ts
enesozturk: This was just to make it look better, yeah forgot to add address line - sorry

packages/core/src/controllers/BlockchainApiController.ts
```diff
@@ -242,6 +245,39 @@ export const BlockchainApiController = {
     })
   },
 
+  async lookupEnsName(name: string) {
+    return api.get<BlockchainApiLookupEnsName>({
+      path: `/v1/profile/account/${name}.wcn.id?projectId=${OptionsController.state.projectId}`
```
enesozturk, line 250: > ..${name}.wcn.id..
As I remember, we removed `wcn.id`  - or did we just change it? Is this okay to keep? When we wanted to change it to something else, is this the only place we need to update?

enesozturk: Can we move `wcn.id` to `core/src/utils/ConstantsUtil.ts`'s `CONSTANTS_UTIL` object so we could ref to it in both `BlockchainApiController` and `EnsController`?

packages/ethers/src/client.ts
```diff
@@ -421,21 +421,26 @@ export class Web3Modal extends Web3ModalScaffold {
...
-          const ensProvider = new InfuraProvider('mainnet')
+        try {
+          const chainId = NetworkUtil.caipNetworkIdToNumber(this.getCaipNetwork()?.id)
+          let ensName: string | null = null
+          let wcName: boolean | string = false
 
-          const name = await ensProvider.resolveName(value)
-          if (name) {
-            return name
+          if (value?.endsWith('wcn.id')) {
```
enesozturk, line 430: Same, seems like we are calling `wcn.id` in multiple places. Calling them like `CONSTANTS_UTIL.ENS_NAME` or something would be really nice

enesozturk (approved):
✅ 🚀

PR #2124, opened 2024-04-06
fix: identity resolution

chris13524:
# Breaking Changes

N/A

# Changes

- fix: identity API not using mainnet
- fix: identity API not being used for ethers SDKs

[Slack conversation](https://walletconnect.slack.com/archives/C03RVH94K5K/p1712418227203839?thread_ts=1712373328.103609&cid=C03RVH94K5K)


packages/ethers/src/client.ts
```diff
@@ -881,20 +881,28 @@ export class Web3Modal extends Web3ModalScaffold {
...
-      if (name) {
-        this.setProfileName(name)
-      }
-      if (avatar) {
-        this.setProfileImage(avatar)
+    try {
+      const { name, avatar } = await this.fetchIdentity({
+        address
+      })
+      this.setProfileName(name)
```
glitch-txs, line 888: show we use conditionals here too? e.g. some address can have `name` but not `avatar`. Not sure how this is handled internally

chris13524: The API returns `null` if there is no avatar so the conditional is redundant?

glitch-txs (approved):

packages/ethers/src/client.ts
```diff
@@ -881,20 +881,28 @@ export class Web3Modal extends Web3ModalScaffold {
...
+        if (name) {
+          this.setProfileName(name)
+        }
+        if (avatar) {
+          this.setProfileImage(avatar)
+        }
+      } else {
+        this.setProfileName(null)
+        this.setProfileImage(null)
       }
```
arein, line 905: can we just remove this code? It was for fallbacks if our API is down

arein (approved):

packages/ethers/src/client.ts
chris13524: Isn't it nice to have the fallback still?

tomiir (approved):

packages/ethers/src/client.ts
chris13524: I think it's not a bad thing to have. We also had downtime this week that this fallback handled.

PR #2127, opened 2024-04-08
fix: ethers fail to disconnect

tomiir:
# Breaking Changes

N/A

# Changes

- feat:
- fix: "Failed to disconnect" error on ethers + relay connections
- chore:

## Description
Email provider does not implement `emit` which  was being fired after disconnection. This caused the client to actually disconnect, but throw an error causing desync in the UI and an error popup.

Tests were not capturing this because the UI kept the modal open and the button in loading state so "account button" was never displayed. Updated it to expect to see the connect button instead


svenvoskamp (approved):

tomiir:
Updated description !

PR #2131, opened 2024-04-09
:feat add send logic 

svenvoskamp:
# Changes

* Add logic for sending native tokens
* Add logic for sending erc20 tokens


arein (commented):
what is this automated testing strategy for this?

I see there are no unit/functional tests here - will this be covered in an end-2-end test? Seems complicated because it requires gas

packages/core/src/controllers/SendController.ts
```diff
@@ -9,6 +9,8 @@ export interface SendControllerState {
   receiverAddress?: string
   receiverProfileName?: string
   receiverProfileImageUrl?: string
+  gasPric?: bigint
```
tomiir, line 12: `gasPrice`

svenvoskamp:
> what is this automated testing strategy for this?
> 
> I see there are no unit/functional tests here - will this be covered in an end-2-end test? Seems complicated because it requires gas

We need to do some research here for the testing strategy

PR #2136, opened 2024-04-10
fix: outdated link in readme

miffycs:
# Breaking Changes

N/A

# Changes

- feat:
- fix: outdated link in readme
- chore:

# Associated Issues

N/A


enesozturk (dismissed):

enesozturk:
@miffycs Thank you for the contribution! 

It's a good catch but I guess you might think it's deprecated since the `www-web3modal` is not public. We cannot do that for security reasons for not, but it's a good reminder to keep as open-source as possible. Maybe in the future? :) 

Closing for now. Best.

PR #2137, opened 2024-04-10
fix: solana tests universal provider issue

gooxpf:
## Description
Fix Universal Provider import errors


packages/solana/src/connectors/universalProvider.ts
```diff
@@ -48,4 +36,15 @@ export class UniversalProviderFactory {
       SolStoreUtil.setAddress('')
     })
   }
+
+  public static async getProvider() {
+    if (!UniversalProviderFactory.provider) {
```
enesozturk, line 41: I think we could wrap the condition in one condition at a level which would make it easier to read



```suggestion
public static async getProvider() {
  if (!UniversalProviderFactory.provider) {
    const provider = await UniversalProviderFactory.init()

    if (!provider) {
      throw new Error('Failed to initialize universal provider')
    }

    return provider
  }

  return UniversalProviderFactory.provider
}
```

packages/solana/src/connectors/universalProvider.ts
```diff
@@ -1,11 +1,10 @@
-import UniversalProvider from '@walletconnect/universal-provider'
+import { UniversalProvider as Provider } from '@walletconnect/universal-provider'
 
 import { SolStoreUtil } from '../utils/scaffold/SolanaStoreUtil.js'
 import type { WalletConnectAppMetadata } from './walletConnectConnector.js'
-
+import type UniversalProvider from '@walletconnect/universal-provider'
```
enesozturk, line 5: I guess this is the main fix? Was it because it's conflicting between the type definitions and the actual class while compiling?

enesozturk (approved):

packages/solana/src/connectors/universalProvider.ts
gooxpf: Main fix is ```import { UniversalProvider as Provider } from '@walletconnect/universal-provider'``` also split type import. [In the example ](https://www.npmjs.com/package/@walletconnect/universal-provider) you can see default import, but default import doesnot contain init function, i guess webpack issue. The problem was UniversalProvider imported from default did not contains init function.

PR #2139, opened 2024-04-10
chore: move preferredAccountType storage to secure-site

tomiir:
# Breaking Changes

N/A

# Changes
- chore: removed `preferredAccountType` from w3m storage
- Uses response from secure-site to update `AccountController` state.

## Closes
https://linear.app/walletconnect/issue/W3M-195/improve-preferredaccounttype-default


packages/ethers/src/client.ts
```diff
@@ -587,28 +588,22 @@ export class Web3Modal extends Web3ModalScaffold {
...
         EthersStoreUtil.setIsConnected(true)
         EthersStoreUtil.setAddress(address as Address)
+        EthersStoreUtil.setPreferredAccountType(preferredAccountType as W3mFrameTypes.AccountType)
         this.setSmartAccountDeployed(Boolean(smartAccountDeployed))
-        this.setSmartAccountEnabledNetworks(smartAccountEnabledNetworks)
 
         this.watchEmail()
         this.watchModal()
       }
+      super.setLoading(false)
```
enesozturk, line 606: Seems like a fix

enesozturk (approved):

packages/ethers/src/client.ts
tomiir: more like an improvement I think, makes it less jumpy

enesozturk (approved):

svenvoskamp (approved):

apps/laboratory/tests/smart-account.spec.ts
```diff
@@ -4,7 +4,9 @@ import { EOA, SMART_ACCOUNT } from './shared/validators/ModalWalletValidator'
 
 import type { ModalWalletValidator } from './shared/validators/ModalWalletValidator'
 
-const NOT_ENABLED_EMAIL = 'test@w3ma.msdc.co'
+function getNotEnabledEmail(parallelIndex: number) {
+  return `test-${parallelIndex}@w3ma.msdc.co`
+}
```
lukaisailovic, line 9: If we are able to check for arbitrary address, we can update `getEmailAddressToUse`  to accept optional secondary param that would override default domain with the other one

tomiir: done !

PR #2141, opened 2024-04-11
hotfix: w3m button incorrect state

svenvoskamp:
# Changes

* patch version to `4.1.7`
* Check if connected state is already present. If this is the case, don't update connected state when emailConnector returns connected state 

closes https://github.com/WalletConnect/web3modal/issues/2126


packages/wagmi/src/client.ts
```diff
@@ -454,8 +454,11 @@ export class Web3Modal extends Web3ModalScaffold {
       })
 
       provider.onNotConnected(() => {
-        this.setIsConnected(false)
-        super.setLoading(false)
+        const isConnected = this.getIsConnectedState()
```
enesozturk, line 457: @svenvoskamp Why are we required to check the `isConnected` state here?

> Check if connected state is already present

(from PR description) Can you explain this a bit more? In which situation it's being already presented?

lukaisailovic (approved):

packages/wagmi/src/client.ts
svenvoskamp: If you're previously connected with an extension wallet or mobile wallet, the isConnected state will be true. The emailProvider will override this with `false`, because it is not previously connected and this will show the incorrect button state.

PR #2146, opened 2024-04-12
chore: remove fixed addresses from SA tests

tomiir:
# Breaking Changes

N/A

# Changes

- chore: refactor SA tests to prevent using fixed ids which were causing issues when ran from different environments with different keys.
- chore: unpack `openSettings` and other utils into single actions to allow for more flexible testing

### DIFF
--- a/apps/laboratory/tests/shared/pages/ModalWalletPage.ts
+++ b/apps/laboratory/tests/shared/pages/ModalWalletPage.ts
@@ -34,10 +34,11 @@ export class ModalWalletPage extends ModalPage {
     await disconnectBtn.click({ force: true })
   }
 
-  async getAddress() {
+  async getAddress(): Promise<string> {
     const address = await this.page.getByTestId('account-settings-address').textContent()
     expect(address, 'Address should be present').toBeTruthy()
 
-    return address as string
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    return address!
   }
 }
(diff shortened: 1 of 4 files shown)


apps/laboratory/tests/shared/pages/ModalWalletPage.ts
```diff
@@ -12,23 +12,18 @@ export class ModalWalletPage extends ModalPage {
...
-    await this.page.getByTestId('w3m-header-close').click()
     await this.page.waitForTimeout(2000)
   }
 
   async togglePreferredAccountType() {
-    await this.openSettings()
     await this.page.getByTestId('account-toggle-preferred-account-type').click()
-    await this.page.getByTestId('w3m-header-close').click()
-    await this.page.waitForTimeout(2000)
+    await this.page.waitForTimeout(2500)
```
lukaisailovic, line 26: Why increased to `2500` instead of `2000`? Does it just work better or?

lukaisailovic (approved):

apps/laboratory/tests/shared/pages/ModalWalletPage.ts
tomiir: Made it less flickery, but will try reducing to 2000 again and see if it's stable enough

lukaisailovic: If its more stable lets keep it as is. Just wanted to know the reason 👌

PR #2102, opened 2024-04-02
refactor/lab ui update design system vars

enesozturk:
# Changes

- Refactored our laboratory UI to use Tailwind, Radix, and Shadcn UI approach
- Updated our w3m design system variable to have a new look.

# Associated Issues

closes #...

PR #2154, opened 2024-04-14
fix: test runner cannot find element when wrapped with another element

enesozturk:
# Issue

The Playwright tests are unable to find the element when another DOM element is on top of the queried element. This causes tests to not proceed and fail.

Screenshot 2024-04-14 at 17 31 05

# Changes

This PR was introduced to replace the toast components which is rendering the toast at the right-bottom of the screen. These changes are already introduced in #2102 but to be able to fix the test issues first, toast replacement is separated to be sure they are running as expected.

### Additional

We are also missing the `MAILSAC_API_KEY` for the ui-tests `Run canary with minimal environment config` step.

### DIFF
--- a/.github/workflows/ui_tests.yml
+++ b/.github/workflows/ui_tests.yml
@@ -95,6 +95,7 @@ jobs:
       - name: Run canary with minimal environment config
         env:
           NEXT_PUBLIC_PROJECT_ID: ${{ secrets.NEXT_PUBLIC_PROJECT_ID }}
+          MAILSAC_API_KEY: ${{ secrets.TESTS_MAILSEC_API_KEY }}
           CI: true
         working-directory: ./apps/laboratory/
         run: npm run playwright:test:canary
--- a/apps/laboratory/package.json
+++ b/apps/laboratory/package.json
@@ -13,7 +13,7 @@
     "playwright:test:email": "playwright test --grep 'email.spec.ts'",
     "playwright:test:siwe": "playwright test --grep siwe.spec.ts",
     "playwright:test:sa": "playwright test --grep smart-account.spec.ts",
-    "playwright:test:canary": "playwright test --retries=0 --grep canary.spec.ts --project='Desktop Chrome/wagmi'",
+    "playwright:test:canary": "playwright test --retries=0 --grep canary.spec.ts --project='Desktop Firefox/ethers'",
     "playwright:debug": "npm run playwright:test -- --debug",
     "playwright:debug:wallet": "npm run playwright:test:wallet -- --debug",
     "playwright:debug:email": "npm run playwright:test:email -- --debug",
@@ -39,7 +39,7 @@
     "next-auth": "4.24.5",
     "react-icons": "4.12.0",
     "siwe": "2.1.4",
-    "sonner": "^1.4.41",
+    "sonner": "1.4.41",
     "valtio": "1.11.2",
     "viem": "2.9.3",
     "wagmi": "2.5.7"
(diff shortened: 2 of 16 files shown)


apps/laboratory/package.json
```diff
@@ -13,7 +13,7 @@
     "playwright:test:email": "playwright test --grep 'email.spec.ts'",
     "playwright:test:siwe": "playwright test --grep siwe.spec.ts",
     "playwright:test:sa": "playwright test --grep smart-account.spec.ts",
-    "playwright:test:canary": "playwright test --retries=0 --grep canary.spec.ts --project='Desktop Chrome/wagmi'",
+    "playwright:test:canary": "playwright test --retries=0 --grep canary.spec.ts --project='Desktop Firefox/ethers'",
```
chris13524, line 16: What are you switching this?

.github/workflows/ui_tests.yml
```diff
@@ -95,6 +95,7 @@ jobs:
       - name: Run canary with minimal environment config
         env:
           NEXT_PUBLIC_PROJECT_ID: ${{ secrets.NEXT_PUBLIC_PROJECT_ID }}
+          MAILSAC_API_KEY: ${{ secrets.TESTS_MAILSEC_API_KEY }}
```
chris13524, line 98: The whole point of this minimal environment step is to have a minimum configuration. The canary doesn't need this API key and so it shouldn't be passed. This key isn't available when the canary runs.

chris13524 (changes requested):

.github/workflows/ui_tests.yml
enesozturk: I see. Then I think we should remove the requirement of using `w3m-email-fixture.ts` which is importing `MAILSAC_API_KEY ` and causing tests to fail. 

I think this is a matter of the PR we introduced minimal test step. I'll take a look how we could do this.

apps/laboratory/package.json
enesozturk: This was for local testing purpose, should have been missed. Will revert back.

enesozturk: Btw, why it's testing only on wagmi?

chris13524: Because the canary is supposed to test if the system is generally working together. E.g. is the relay down, and what is the overall latency of the system over-time. Chromium/Wagmi is the sort of best-case configuration/recommendation and is least likely to fail from test flakes

.github/workflows/ui_tests.yml
chris13524: This regression happened before, see [this PR](https://github.com/WalletConnect/web3modal/pull/2106) which @tomiir helped refactor

.github/workflows/ui_tests.yml
```diff
@@ -95,6 +95,7 @@ jobs:
       - name: Run canary with minimal environment config
```
chris13524, line 95:
```suggestion
      - name: Run canary with minimal environment config
        # Keep in-sync with https://github.com/WalletConnect/rs-relay/blob/master/terraform/canary/main.tf#L48
```

.github/workflows/ui_tests.yml
tomiir: Is the canary now running new tests? Why was it working before and not anymore?

enesozturk: @tomiir have no idea tbh. I think we have already open PR about it that I just saw: https://github.com/WalletConnect/web3modal/pull/2029

chris13524: The canary has been working, you can see it at the bottom of [this Grafana dashboard](https://g-e66b9a1a24.grafana-workspace.eu-central-1.amazonaws.com/d/prod-at-a-glance/prod-at-a-glance?orgId=1). Doesn't seem like #2029 affected it

PR #2158, opened 2024-04-15
fix: email update

ignaciosantise:
# Changes
- fix: use `updateEmail` response to decide whether to redirect to first or second OTP screen

Steps to reproduce issue in production:
1. Create a NEW email account using an unused email
2. After connecting, change your email to a new one fast
3. Magic will send an OTP code directly to the new email and not to the current one
4. This causes the view to hang waiting for UPDATE_EMAIL_SUCCESS event

# Associated Issues
https://linear.app/walletconnect/issue/W3M-171/%5Bbug%5D-freeze-update-email-flow-when-otp-is-sent-to-new-email-only

### Related
Secure site PR 120


enesozturk (approved):

tomiir (approved):
This looks OK, just wondering if zod will break when secure site starts returning new data and old versions can't handle it. 

Did you try new secure site with older versions of w3m?

ignaciosantise:
@tomiir yes, i tested the new secure site with an old version of w3m and it doesn't break because `zod.parse` removes the extra data. But i'll test everything again to be 100% sure.

ignaciosantise:
@arein @lukaisailovic added update email tests

apps/laboratory/tests/shared/pages/ModalPage.ts
```diff
@@ -226,4 +228,51 @@ export class ModalPage {
...
+
+    // Clear messages before putting new email
+    await email.deleteAllMessages(this.emailAddress)
+    await this.page.getByTestId('wui-email-input').locator('input').press('Enter')
+
+    // Wait until the next screen appears
+    await expect(this.page.getByText('Enter the code we sent')).toBeVisible({
+      timeout: 10_000
+    })
+    const confirmCurrentEmail = await this.page.getByText(this.emailAddress).isVisible()
```
arein, line 249: nitpick: would be nice to replace this with getById

ignaciosantise: in this case, i need to verify if the screen contains the current email or the new email. Getting the component by ID will always return true

apps/laboratory/tests/email.spec.ts
```diff
@@ -2,6 +2,11 @@ import { expect } from '@playwright/test'
 import { testMEmail } from './shared/fixtures/w3m-email-fixture'
 import { SECURE_WEBSITE_URL } from './shared/constants'
 
+const mailsacApiKey = process.env['MAILSAC_API_KEY']
+if (!mailsacApiKey) {
+  throw new Error('MAILSAC_API_KEY is not set')
+}
+
```
lukaisailovic, line 9: I don't  think we need this

ignaciosantise: you mean to remove the if condition? In that case, i should set a default value for the key

i need the mailsacApiKey

apps/laboratory/tests/shared/pages/ModalPage.ts
```diff
@@ -93,7 +96,8 @@ export class ModalPage {
     if (otp.replace(' ', '').length !== 6) {
       otp = email.getOtpCodeFromBody(emailBody)
     }
-    await this.enterOTP(otp)
+    // Remove empy space in OTP code 111 111
+    await this.enterOTP(otp.replace(' ', ''))
```
tomiir, line 100: maybe the `replace` could be in `getOtpCodeFromBody` to prevent repeating

ignaciosantise: done

tomiir (approved):

lukaisailovic (approved):

PR #2162, opened 2024-04-16
fix: tests race condition

lukaisailovic:
Fixing an issue that would introduce race condition in case tests are ran from multiple environments at the same time.


tomiir (approved):

arein (approved):
Do we need to reserve these addresses in Mailsac's UI?
The reason I used so before when I set it up was because I needed to register. But I assume that this is fine now due to the custom domain.

lukaisailovic:
> Do we need to reserve these addresses in Mailsac's UI? The reason I used so before when I set it up was because I needed to register. But I assume that this is fine now due to the custom domain.

@arein You are correct. Custom domain allows us to do {anything}@ourdomain and we will catch the messages with the API

PR #2164, opened 2024-04-17
remove siwe signOut function on high level packages

glitch-txs:
# Breaking Changes

N/A

# Changes

- feat:
- fix: Web3Modal was calling the SIWE's signOut function twice.
the signOut function from SIWE is already called internally in [`scaffold/src/modal/w3m-modal`](https://github.com/WalletConnect/web3modal/blob/0478feb5b420c2eefcadd3410f97e2d8bffbbecf/packages/scaffold/src/modal/w3m-modal/index.ts#L181)
image

Context: https://walletconnect.slack.com/archives/C03RVH94K5K/p1713282453076899
- chore:

# Associated Issues

closes #...


chris13524 (commented):
Not approving as I'm not intimately familiar, but thanks for the fix!!

tomiir (approved):

PR #2166, opened 2024-04-18
chore: added update email tests

ignaciosantise:
# Changes
- chore: added update email tests

# Associated feature
https://github.com/WalletConnect/web3modal/pull/2158


apps/laboratory/tests/email.spec.ts
```diff
@@ -2,6 +2,11 @@ import { expect } from '@playwright/test'
 import { testMEmail } from './shared/fixtures/w3m-email-fixture'
 import { SECURE_WEBSITE_URL } from './shared/constants'
 
+const mailsacApiKey = process.env['MAILSAC_API_KEY']
```
enesozturk, line 5: @ignaciosantise We'd better to be skeptical about calling API key here bc previously we had issues by the tests which doesn't need `MAILSAC_API_KEY` and moved them into the functions. 

See: https://github.com/WalletConnect/web3modal/pull/2029

Is this still intentional?

enesozturk (approved):
@ignaciosantise left a comment

apps/laboratory/tests/email.spec.ts
ignaciosantise: Ahh okey! didn't know that. Where should i get the api key? Here?

```
testMEmail('it should update email', async ({ modalPage }) => {
  const mailsacApiKey = process.env['MAILSAC_API_KEY']
  if (!mailsacApiKey) {
    throw new Error('MAILSAC_API_KEY is not set')
  }

  await modalPage.updateEmail(mailsacApiKey)
})
```

tomiir (approved):

PR #2167, opened 2024-04-19
chore: trim the query when searching for wallets

lukaisailovic:
When searching for wallets we would send any whitespace as well which would present us with "no wallets" screen. 

Made two changes:
- Sending `trimmedQuery` to ApiController instead of `query`
- Compare `trimmedQuery` with previous query that is also trimmed. This will not send network calls for whitespaces


lukaisailovic:
> Wondering if this should be inside of ApiController directly + test? 

@tomiir Thats a good point. I think I will have to trim on the UI layer to not trigger additional requests (changed if condition).

Do you think we should add trimming to the controller in addition to the current logic?

chris13524:
+1 on a test for this

packages/core/tests/controllers/ApiController.test.ts
```diff
@@ -383,6 +383,64 @@ describe('ApiController', () => {
...
+      }
+    ]
+    OptionsController.setIncludeWalletIds(includeWalletIds)
+    OptionsController.setExcludeWalletIds(excludeWalletIds)
+
+    let fetchSpy = vi.spyOn(api, 'get').mockResolvedValue({ data })
+    const fetchImageSpy = vi.spyOn(ApiController, '_fetchWalletImage').mockResolvedValue()
+
+    // Whitespace
+    await ApiController.searchWallet({ search: 'MetaMask    ' })
```
chris13524, line 404: Test with leading whitespace not just trailing would be helpful too

lukaisailovic: @chris13524 Done!

tomiir (approved):

PR #2169, opened 2024-04-19
chore: upload canary test results

chris13524:
# Breaking Changes

N/A

# Changes

- chore: add upload to S3 bucket

# Associated Issues

Support in canary deployment: https://github.com/WalletConnect/rs-relay/pull/1514

[Slack conversation](https://walletconnect.slack.com/archives/C058RS0MH38/p1713543851866829?thread_ts=1713537481.807059&cid=C058RS0MH38)


Dockerfile.canary
```diff
@@ -12,4 +13,4 @@ RUN npm run playwright:install
 
 ENV TIMING_LOGS=true
 
-CMD ["npm", "run", "playwright:test:canary"]
+CMD ["./docker-canary.sh"]
```
chris13524, line 16: Doing as script instead of in-script (like CloudWatch) because the trace.zip isn't generated until after the test completes.

enesozturk (approved):

tomiir (approved):

apps/laboratory/docker-canary.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/env/bin
+
+# Not adding `set -e` so that S3 upload happens regardless
+
+npm run playwright:canary:test
+
+destination="s3://$TEST_RESULTS_BUCKET/web3modal-canary/$(date --iso-8601=seconds)/test-results/"
+echo "Uploading test results to $destination"
+aws s3 cp ./test-results/ $destination --recursive
```
arein, line 9: minor nitpick: dis fails now if the container doesn't have AWS credis which breaks some Developer Experience use-cases

arein (approved):

apps/laboratory/docker-canary.sh
chris13524: What use case does it break?

I had a comment here explaining why I had to do it this way: https://github.com/WalletConnect/web3modal/pull/2169#discussion_r1572704269

chris13524: We can make it optional e.g. with an env var in another PR if we need it

PR #2170, opened 2024-04-19
fix: vitest gh actions doesn't throw error

enesozturk:
# Changes

### Refactor Scripts and Workflows

Since we are using monorepo and the Vitest coverage tools don't have a proper monorepo setup, we are merging all the generated coverage files from our packages and saving them to another file with a script. 

The `vitest-coverage-report-action` is using these merged coverage reports to generate comments on the PR.

> [!NOTE]  
> A two different scripts like the following prevent workflow to detect issues.
>
> `"test": "npm run test; node ./scripts/coverage.js`
>
> If the `test` script fails, when it run the `coverage.js` script, the CI thinks that the job is OK.

### Upgrade Vitest

Vitest has many improvements to the GH actions in the recent versions. Upgrading it might help us to raise the error in CI.

Example: https://github.com/vitest-dev/vitest/pull/5093
See recent releases: https://github.com/vitest-dev/vitest/releases

### Fix `PublicStateController` tests

Recently we added the `loading` state to the PublicStateController. The unit tests needed some updates regarding to this.

~~*Intentionally, one of the test are still broken to test the GH Action issue.~~

### Add `vite.config.ts`

The `vitest-coverage-report-action` requires the Vite config file in the root.


lukaisailovic (approved):
Awesome  👌

PR #2173, opened 2024-04-22
:feat add send logic 

svenvoskamp:
# Changes

* Add logic for sending native tokens
* Add logic for sending erc20 tokens


packages/ethers/src/client.ts
```diff
@@ -341,6 +343,32 @@ export class Web3Modal extends Web3ModalScaffold {
...
+
+        const browserProvider = new BrowserProvider(provider, chainId)
+        const signer = new JsonRpcSigner(browserProvider, address)
+        const contract = new Contract(data.tokenAddress, erc20ABI, signer)
+
+        if (!contract || typeof contract['transfer'] !== 'function') {
+          throw new Error('Contract or transfer method is undefined')
+        }
+
+        const tx = await contract['transfer'](data.receiverAddress, data.tokenAmount)
```
tomiir, line 369: Is this really needed? I'd think  `writeContract` would be generic, not tied to `transfer`. 
Any way we can make this method to execute any contract write and not just transfers?

packages/scaffold/src/views/w3m-wallet-send-preview-view/index.ts
```diff
@@ -114,11 +129,64 @@ export class W3mWalletSendPreviewView extends LitElement {
...
+
+      try {
+        await ConnectionController.sendTransaction({
+          to,
+          address,
+          data,
+          value,
+          gasPrice: this.gasPrice
+        })
+        RouterController.replace('Account')
```
tomiir, line 184: Is this required if we use the transaction stack?

packages/scaffold/src/views/w3m-wallet-send-preview-view/index.ts
```diff
@@ -114,11 +129,64 @@ export class W3mWalletSendPreviewView extends LitElement {
...
+    } else if (this.receiverAddress && this.sendTokenAmount && this.gasPrice) {
+      RouterController.pushTransactionStack({
+        view: null,
+        goBack: true
+      })
+
+      const to = this.receiverAddress as `0x${string}`
+      const address = AccountController.state.address as `0x${string}`
+      const value = ConnectionController.parseUnits(this.sendTokenAmount.toString(), 18)
+      const data = '0x'
```
tomiir, line 174: to my point before for a generic `writeContract` you could encode all of this call into callData here.

packages/core/src/controllers/ConvertController.ts
```diff
@@ -385,12 +385,15 @@ export const ConvertController = {
...
 
-    const value = res.instant
+    const value = res.standard
     const gasFee = BigInt(value)
     const gasLimit = BigInt(INITIAL_GAS_LIMIT)
     const gasPrice = this.calculateGasPriceInUSD(gasLimit, gasFee)
 
     state.gasPriceInUSD = gasPrice
+
+    // eslint-disable-next-line consistent-return
```
enesozturk, line 395: I guess rather than doing an empty return on top, we could handle this better way to give better feedback to the user. When the endpoint return an error, what would happen on the UI side?

I could tackle this in the improvements of the convert.

packages/core/src/utils/CoreHelperUtil.ts
```diff
@@ -54,6 +54,16 @@ export const CoreHelperUtil = {
     return caipAddress.split(':')[2]
   },
 
+  getEvmChainId(caipNetworkId?: `${string}:${string}`) {
```
enesozturk, line 57: There should already be a function for this? Could you check `common/NetworkUtil`'s `caipNetworkIdToNumber` function?

packages/core/src/utils/CoreHelperUtil.ts
```diff
@@ -275,5 +285,14 @@ export const CoreHelperUtil = {
     }
 
     return false
+  },
+
+  extractEthereumAddress(address: string) {
```
enesozturk, line 290: Same. You could use `CoreHelperUtil.getPlainAddress` function from `core` package.

packages/ethers/tsconfig.json
```diff
@@ -6,5 +6,5 @@
     "baseUrl": "../../"
   },
   "extends": "../../tsconfig.json",
-  "include": ["exports", "src"]
+  "include": ["exports", "src", "src/**/*.json"]
```
enesozturk, line 9: Do we need this even though we have `src` in the list?

packages/scaffold/src/views/w3m-wallet-send-preview-view/index.ts
enesozturk: Totally - we could use the same API as we do in ConvertController on top (before) of the `sendTransaction`;


```js
RouterController.pushTransactionStack({
      view: 'Account',
      goBack: false,
      onSuccess() {
        ...
      }
    })
```

packages/ui/src/utils/Math.ts
```diff
@@ -0,0 +1,30 @@
+/**
+ * Interpolates a value from one range to another
+ * @param inputRange - number array of length 2 that represents the original range
+ * @param outputRange - number array of length 2 that represents the new range
+ * @param value - the value to interpolation
+ * @returns
+ */
+export function interpolate(inputRange: number[], outputRange: number[], value: number) {
```
enesozturk, line 8: This should be now used under `ui` package: `MathUtil.interpolate`

packages/ui/src/utils/NumberUtil.ts
```diff
@@ -0,0 +1,23 @@
+/**
+ * Format the given number or string to human readable numbers with the given number of decimals
+ * @param value - The value to format. It could be a number or string. If it's a string, it will be parsed to a float then formatted.
+ * @param decimals - number of decimals after dot
+ * @returns
+ */
+export function formatNumberToLocalString(value: string | number | undefined, decimals = 2) {
```
enesozturk, line 7: This should be now used under `ui` package: `UiHelperUtil.formatNumberToLocalString`

packages/wagmi/tsconfig.json
```diff
@@ -6,5 +6,5 @@
     "baseUrl": "../../"
   },
   "extends": "../../tsconfig.json",
-  "include": ["exports", "src"]
+  "include": ["exports", "src", "src/**/*.json"]
```
enesozturk, line 9: Same, is this intentional? We already have `src` file listed in `include` array.

packages/scaffold/src/views/w3m-wallet-send-preview-view/index.ts
svenvoskamp: This is for native tokens for a `sendTransaction`, not a `writeContract`

packages/core/src/controllers/ConvertController.ts
tomiir: Instead of disabling the rule, wouldn't it make more sense to return something like 
`{ gasPrice: null, gasPriceInUsd: null }` on the earlt return?

packages/ethers/src/client.ts
```diff
@@ -341,6 +339,35 @@ export class Web3Modal extends Web3ModalScaffold {
...
+          throw new Error('connectionControllerClient:sendTransaction - provider is undefined')
+        }
+
+        if (!address) {
+          throw new Error('connectionControllerClient:sendTransaction - address is undefined')
+        }
+
+        const browserProvider = new BrowserProvider(provider, chainId)
+        const signer = new JsonRpcSigner(browserProvider, address)
+        const contract = new Contract(data.tokenAddress, erc20ABI, signer)
```
tomiir, line 357: we're still depending on the ABI here :| 
I think it'd be best for now to keep this a separate method like `sendERC20` so we can be as flexible as we want with implementation. 

@enesozturk how did you work around this with convert? I assume you need to execute contract writes as well?

svenvoskamp: We can send the abi as argument?

enesozturk: On the convert side, we don't do contract writing since we are using 1Inch's aggregator contract and send transaction with parameters of `from`, `to`, and a few others.

packages/scaffold/src/views/w3m-wallet-send-preview-view/index.ts
```diff
@@ -114,11 +130,64 @@ export class W3mWalletSendPreviewView extends LitElement {
     return null
   }
 
-  private onSendClick() {
-    RouterController.reset('Account')
-    setTimeout(() => {
-      SendController.resetSend()
-    }, 200)
+  private async onSendClick() {
```
enesozturk, line 133: I think moving business logic to the controllers might help us to keep the UI components and partials smaller. Since we are also using different controllers inside the function, this might also decrease the dependencies of these components. This would be a function on SendController which gets only parameters and does the things. Wdyt?

svenvoskamp: I agree, added business logic to `SendController`

packages/ethers/src/client.ts
```diff
@@ -342,6 +339,35 @@ export class Web3Modal extends Web3ModalScaffold {
         const txReceipt = await txResponse.wait()
 
         return (txReceipt?.hash as `0x${string}`) || null
+      },
+
+      writeContract: async (data: WriteContractArgs) => {
+        const { chainId, provider, address } = EthersStoreUtil.state
+
+        if (!provider) {
+          throw new Error('connectionControllerClient:sendTransaction - provider is undefined')
```
tomiir, line 348: Should this say Ethers client instead of connectionController?

packages/scaffold/src/partials/w3m-wallet-send-details/index.ts
```diff
@@ -14,11 +14,15 @@ export class W3mWalletSendDetails extends LitElement {
...
   @property({ type: Object }) public caipNetwork?: CaipNetwork
 
+  @property({ type: Number }) public networkFee?: number
+
   // -- Render -------------------------------------------- //
   public override render() {
     return html` <wui-text variant="small-400" color="fg-200">Details</wui-text>
       <wui-flex flexDirection="column" gap="xxs">
-        <wui-list-content textTitle="Network cost" textValue="$3.20"></wui-list-content>
+        <wui-list-content textTitle="Network cost" textValue="${'$'}${ifDefined(
```
tomiir, line 23: No need for {} here

packages/scaffold/src/views/w3m-account-settings-view/index.ts
```diff
@@ -27,6 +28,8 @@ export class W3mAccountSettingsView extends LitElement {
 
   private readonly networkImages = AssetController.state.networkImages
 
+  private text = ''
```
tomiir, line 31: Is this ok or should it be a stare variable?

packages/scaffold/src/views/w3m-wallet-send-view/index.ts
```diff
@@ -78,6 +89,16 @@ export class W3mWalletSendView extends LitElement {
   }
 
   // -- Private ------------------------------------------- //
+
+  private async fetchNetworkPrice() {
+    await ConvertController.getNetworkTokenPrice()
```
tomiir, line 94: Maybe this should be abstracted from convert controller to somewhere more generic?
Cc @enesozturk

tomiir (approved):
Looks good overall! Just some minor comments / questions

packages/scaffold/src/views/w3m-wallet-send-view/index.ts
svenvoskamp: Leaving this for now, but I guess we need to refactor this a bit in the future

packages/scaffold/src/partials/w3m-input-token/index.ts
```diff
@@ -66,7 +69,9 @@ export class W3mInputToken extends LitElement {
       const price = this.token.price
       const totalValue = price * this.sendTokenAmount
 
-      return html`<wui-text variant="small-400" color="fg-200">$${totalValue.toFixed(2)}</wui-text>`
+      return html`<wui-text class="totalValue" variant="small-400" color="fg-200"
+        >${totalValue ? `$${totalValue.toFixed(2)}` : 'Incorrect value'}</wui-text
```
enesozturk, line 73: We can use `UiHelperUtil.formatNumberToLocalString` for better price formatting

packages/scaffold/src/partials/w3m-input-token/index.ts
```diff
@@ -106,7 +111,19 @@ export class W3mInputToken extends LitElement {
 
   private onMaxClick() {
     if (this.token) {
-      SendController.setTokenAmount(Number(this.token?.quantity.numeric))
+      if (this.gasPriceInUSD) {
```
enesozturk, line 114: If we don't have else state, maybe we could use both in same if?

```suggestion
      if (this.token && this.gasPriceInUSD) {
```

packages/scaffold/src/partials/w3m-wallet-send-details/index.ts
```diff
@@ -14,11 +14,15 @@ export class W3mWalletSendDetails extends LitElement {
...
 
+  @property({ type: Number }) public networkFee?: number
+
   // -- Render -------------------------------------------- //
   public override render() {
     return html` <wui-text variant="small-400" color="fg-200">Details</wui-text>
       <wui-flex flexDirection="column" gap="xxs">
-        <wui-list-content textTitle="Network cost" textValue="$3.20"></wui-list-content>
+        <wui-list-content textTitle="Network cost" textValue="$${ifDefined(
+          this.networkFee?.toFixed(2).toString()
```
enesozturk, line 24: Same for this, `UiHelperUtil.formatNumberToLocalString` could be used

packages/scaffold/src/views/w3m-account-settings-view/index.ts
```diff
@@ -236,6 +243,13 @@ export class W3mAccountSettingsView extends LitElement {
 
     this.loading = true
     await emailConnector?.provider.setPreferredAccount(accountTypeTarget)
+    await ConnectionController.reconnectExternal(emailConnector)
+    this.switched = true
+    this.text =
+      this.text === 'Switch to your smart account'
```
enesozturk, line 249: Why don't we use `this.preferredAccountType === W3mFrameRpcConstants.ACCOUNT_TYPES.SMART_ACCOUNT` as we did on top? Using hard coded string for condition doesn't seem reliable. We should be avoiding as much as possible.

packages/scaffold/src/views/w3m-wallet-send-preview-view/index.ts
```diff
@@ -114,11 +137,27 @@ export class W3mWalletSendPreviewView extends LitElement {
...
   }
 
-  private onSendClick() {
-    RouterController.reset('Account')
-    setTimeout(() => {
-      SendController.resetSend()
-    }, 200)
+  onSendClick() {
+    if (this.token?.address && this.sendTokenAmount && this.receiverAddress) {
+      SendController.sendERC20Token({
```
enesozturk, line 142: I'd prefer to move this logic to `SendController` and on the UI, just call `SendController.sendToken()` which will do conditions inside that function. This would decrease the business logic on the UI side and move the maintainability to controllers (if we have all these states in `SendController`) - We could go this was as well, not a strong opinion.

packages/wagmi/src/client.ts
```diff
@@ -208,6 +223,54 @@ export class Web3Modal extends Web3ModalScaffold {
...
+        } catch (error) {
+          return false
+        }
+      },
+
+      getEnsAvatar: async (value: string) => {
+        try {
+          const avatar = await wagmiGetEnsAvatar(this.wagmiConfig, {
+            name: normalize(value),
+            chainId: 1
```
enesozturk, line 261: Is it using `chainId: 1` because ENS is only on Ethereum?

svenvoskamp: Added a check here if `chainId === mainnet.id`. I think we need a extra check on the blockchainAPI to resolve the wallet address from an ENS name. It is now only possible to resolve a name with the wallet address, not otherwise.

packages/wallet/src/RegexUtil.ts
```diff
@@ -2,5 +2,5 @@ export const RegexUtil = {
   address: /^0x(?:[A-Fa-f0-9]{40})$/u,
   transactionHash: /^0x(?:[A-Fa-f0-9]{64})$/u,
   // Signed message hash can be variable length
-  signedMessage: /^0x(?:[a-fA-F0-9]{6,})$/u
+  signedMessage: /^0x(?:[a-fA-F0-9]{62,})$/u
```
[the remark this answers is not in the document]
svenvoskamp: I think we made sure that all signatures have a minimum of 62 chars right? @tomiir

tomiir (approved):

packages/wallet/src/RegexUtil.ts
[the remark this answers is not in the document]
tomiir: It works for safe smart accounts for now. We need to refactor the RPC entirely to use request IDs and act accrodingly. We can define this on our next meet.

PR #2181, opened 2024-04-23
refactor: modal test regression issues

enesozturk:
# Changes

## Modal animation refactor

We are using JS-based animations which have some regression issues. When we click the account button too fast, this will try to open and close the modal too fast and after some point, the modal goes into a state where it's trying to run animation before the previous animation didn't end. 

This could be fixed in several ways. When it comes to animations, the CSS only solution is the best when it comes to performance. This PR introduces changes to our modal open/close animations and moves them to a CSS solution.

### Steps to reproduce

- Open any lab example 
- Press the account button several times too fast

## Switch network test logic

While we switch networks, we don't properly wait for the required changes. This is causing test runners to go into weird states or close the modal in unexpected situations. 

Now, we are waiting for the network switch button to have switched the network name to complete the switch network logic. This addresses most of the regression issues.


svenvoskamp (approved):

lukaisailovic (approved):

apps/laboratory/tests/shared/pages/ModalPage.ts
```diff
@@ -115,7 +115,7 @@ export class ModalPage {
       this.page.getByText(email),
       `Expected current email: ${email} to be visible on the notification screen`
     ).toBeVisible({
-      timeout: 10_000
+      timeout: 20_000
```
enesozturk, line 118: Revert

apps/laboratory/src/components/Ethers/EthersTests.tsx
```diff
@@ -6,8 +7,17 @@ import { EthersTransactionTest } from './EthersTransactionTest'
 import { EthersWriteContractTest } from './EthersWriteContractTest'
 
 export function EthersTests() {
+  const [ready, setReady] = React.useState(false)
   const { isConnected } = useWeb3ModalAccount()
 
+  React.useEffect(() => {
```
glitch-txs, line 13: let's avoid usingn this, is bad practice for SSR. If we have hydration errors we should fix it in the SDK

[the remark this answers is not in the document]
enesozturk: @glitch-txs totally agree. This is for testing purposes. A lot different reasons for failing tests going on. Making some discovery.

apps/laboratory/src/components/Ethers/EthersSignTypedDataTest.tsx
```diff
@@ -48,7 +48,12 @@ export function EthersSignTypedDataTest() {
 
       const signature = await signer?.signTypedData(domain, types, message)
 
-      toast({ title: 'Succcess', description: signature, status: 'success', isClosable: true })
+      toast({
+        title: 'Success',
+        description: `${signature.slice(0, 90)}...`,
```
chris13524, line 53: We are truncating the signature now? Could we add it to the page somewhere so we can copy the full signature? Useful for other things sometimes

apps/laboratory/tests/shared/pages/ModalPage.ts
```diff
@@ -154,11 +154,11 @@ export class ModalPage {
     const accountBtn = this.page.getByTestId('account-button')
     await expect(accountBtn, 'Account button should be visible').toBeVisible()
     await expect(accountBtn, 'Account button should be enabled').toBeEnabled()
-    await accountBtn.click({ force: true })
+    await accountBtn.click()
     const disconnectBtn = this.page.getByTestId('disconnect-button')
     await expect(disconnectBtn, 'Disconnect button should be visible').toBeVisible()
     await expect(disconnectBtn, 'Disconnect button should be enabled').toBeEnabled()
-    await disconnectBtn.click({ force: true })
+    await disconnectBtn.click()
```
chris13524, line 161: Nice; generally better to avoid force:true

apps/laboratory/tests/shared/pages/ModalWalletPage.ts
```diff
@@ -18,20 +18,26 @@ export class ModalWalletPage extends ModalPage {
...
 
   override async disconnect(): Promise<void> {
-    this.openSettings()
     const disconnectBtn = this.page.getByTestId('disconnect-button')
     await expect(disconnectBtn, 'Disconnect button should be visible').toBeVisible()
     await expect(disconnectBtn, 'Disconnect button should be enabled').toBeEnabled()
-    await disconnectBtn.click({ force: true })
+    await disconnectBtn.click()
+    await this.page.getByTestId('connect-button').waitFor({ state: 'visible', timeout: 5000 })
+    await this.page.waitForTimeout(2000)
```
chris13524, line 40: Why do we need this timeout?

apps/laboratory/tests/shared/validators/ModalValidator.ts
```diff
@@ -67,6 +67,10 @@ export class ModalValidator {
     await expect(this.page.getByText(ConstantsUtil.SigningSucceededToastTitle)).toBeVisible({
       timeout: 30 * 1000
     })
+    const closeButton = this.page.locator('button[aria-label="Close"]')
+    await expect(closeButton).toBeVisible()
+    await closeButton.click()
+    await this.page.waitForTimeout(200)
```
chris13524, line 73: Why do we need this timeout?

apps/laboratory/tests/shared/pages/ModalWalletPage.ts
```diff
@@ -18,20 +18,26 @@ export class ModalWalletPage extends ModalPage {
   override async switchNetwork(network: string) {
     await this.page.getByTestId('account-switch-network-button').click()
     await this.page.getByTestId(`w3m-network-switch-${network}`).click()
-    await this.page.waitForTimeout(2000)
+    const switchNetworkButton = this.page.getByTestId('account-switch-network-button')
+    await expect(switchNetworkButton).toBeVisible()
+    await expect(switchNetworkButton, `Switched network should include ${network}`).toContainText(
+      network,
+      { timeout: 10_000 }
+    )
```
lukaisailovic, line 26: Why was this logic moved from `ModalValidator`?

tomiir: I'd change this for a regular timeout (to wait for the switch) and move the expect back to the validator

enesozturk: When we use the switch network step, we should both:

- be sure it switched as expected before stepping on
- be not responsible of checking network switches in high level

I feel like making these inside the `switchNetwork` would be better in terms of encapsulation. @lukaisailovic

tomiir (approved):
🚀

svenvoskamp (approved):

apps/laboratory/src/components/Ethers/EthersTests.tsx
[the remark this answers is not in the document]
enesozturk: https://linear.app/walletconnect/issue/W3M-291/[ssr]-improve-ssr-support-for-w3m

apps/laboratory/tests/shared/pages/ModalWalletPage.ts
enesozturk: That's a good point. @lukaisailovic Could you review again please?

glitch-txs (approved):

PR #2184, opened 2024-04-23
add status

glitch-txs:
# Breaking Changes

N/A

# Changes

- feat: add status state with 'reconnecting', 'connected' and 'disconnected' values.
- fix:
- chore:

# Associated Issues

closes #...


svenvoskamp (changes requested):
LGTM, but I think there are some linter changes as well because you removed `plugin:prettier/recommended`. Could you resolve that?

turbo.json
```diff
@@ -32,7 +32,7 @@
         "^build:core",
         "^build:ui",
         "^build:wallet",
-        "@web3modal/siwe#build:siwe",
+        "^build:siwe",
```
tomiir, line 35: is this ok @glitch-txs ?

glitch-txs: yes

svenvoskamp (approved):

PR #2186, opened 2024-04-24
fix: don't use random emails in tests

lukaisailovic:
Reuse same set of emails for tests. Introduce offset for email change emails. Number of emails we consume is currently `numberOfWorkers * 2` (16 at the moment)


apps/laboratory/tests/shared/utils/email.ts
```diff
@@ -78,8 +77,6 @@ export class Email {
   }
 
   getEmailAddressToUse(index: number, domain = EMAIL_DOMAIN): string {
-    const prefix = randomBytes(12).toString('hex')
-
-    return `${prefix}-w${index}@${domain}`
+    return `w3m-w${index}@${domain}`
```
tomiir, line 80: Wouldn't this result in same runner executing in parallel in different contexts? 
eg. multiple prs here or on secure site, or local dev running tests

unless it's an accepted "risk" due to cost?

lukaisailovic: @tomiir It would. Previous change was implemented for that exact reason. And you're right, due to cost its reverted.

tomiir (approved):

PR #2190, opened 2024-04-25
Add signAndSendTransaction to Solflare, Backpack and Trust Wallet Ada…

gooxpf:
# Changes
Add signAndSendTransaction to BackPack, SolFlare and Trust wallet adaoters.

- fix: Program counter error description.


tomiir (approved):
Looks good but have a nitpick: 
They all look exacly the same right? 
Any way we could maybe abstract this so we don't repeat as much?

PR #2195, opened 2024-04-26
refactor: sync theme variables with secure site

enesozturk:
# Changes

The secure site uses the theme variables from the web3modal version it has. But this approach is problematic. When we make changes to our theme variables over time, and if users have a web3modal version that is not the same version as the secure site's web3modal. This would create UI's with a different look inside the transaction approval screens.

This PR introduces changes to share our w3m theme variables with a secure site with a `syncTheme` listener and render the UI with the same variables of w3m.

See the PR on the web3modal for the required changes: https://github.com/WalletConnect/secure-web3modal/pull/122


packages/scaffold/src/client.ts
```diff
@@ -38,6 +39,7 @@ export interface LibraryOptions {
   projectId: OptionsControllerState['projectId']
   themeMode?: ThemeMode
   themeVariables?: ThemeVariables
+  w3mThemeVariables?: W3mThemeVariables
```
svenvoskamp, line 42: `w3mThemeVariables` shouldn't be in options

packages/ui/src/utils/ThemeUtil.ts
```diff
@@ -40,7 +40,7 @@ export function setThemeVariables(themeVariables: ThemeVariables) {
   }
 }
 
-function createRootStyles(themeVariables?: ThemeVariables) {
+export function createRootStyles(themeVariables?: ThemeVariables) {
```
svenvoskamp, line 43: Do we need to export it with this solution?

enesozturk: Nope, let me revert

packages/wallet/src/W3mFrameSchema.ts
```diff
@@ -41,7 +41,8 @@ export const AppUpdateEmailPrimaryOtpRequest = z.object({ otp: z.string() })
 export const AppUpdateEmailSecondaryOtpRequest = z.object({ otp: z.string() })
 export const AppSyncThemeRequest = z.object({
   themeMode: z.optional(z.enum(['light', 'dark'])),
-  themeVariables: z.optional(z.record(z.string(), z.string().or(z.number())))
+  themeVariables: z.optional(z.record(z.string(), z.string().or(z.number()))),
+  w3mThemeVariables: z.record(z.string(), z.string())
```
tomiir, line 45: if this is not `optional` it will require installing a new v on secure-site to prevent zod crashes

enesozturk: Let's go with the new version way. This should be required when we call `syncTheme` function to keep our themes synced, wdyt? @tomiir

tomiir (approved):

lukaisailovic (approved):

svenvoskamp (approved):

PR #2196, opened 2024-04-26
Chore/sa sign tests

tomiir:
# Breaking Changes

N/A

# Changes

- Adds W3m info section with `address` and `chainId`
- Adds `getAddress`, `getChainId` and `expectValidSignature` to `ModalValidator`
- Adds test to check for 6942 signature validity on smart-accounts

## Notes
Requires https://github.com/WalletConnect/secure-web3modal/pull/123

### DIFF
--- a/apps/laboratory/src/components/Ethers/EthersModalInfo.tsx
+++ b/apps/laboratory/src/components/Ethers/EthersModalInfo.tsx
@@ -1,0 +1,10 @@
+import * as React from 'react'
+import { useWeb3ModalAccount } from '@web3modal/ethers/react'
+
+import { Web3ModalInfo } from '../Web3ModalInfo'
+
+export function EthersModalInfo() {
+  const { isConnected, address, chainId } = useWeb3ModalAccount()
+
+  return isConnected ? <Web3ModalInfo address={address} chainId={chainId} /> : null
+}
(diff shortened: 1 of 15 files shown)


apps/laboratory/src/components/Ethers/EthersModalInfo.tsx
```diff
@@ -0,0 +1,10 @@
+import * as React from 'react'
+import { useWeb3ModalAccount } from '@web3modal/ethers/react'
+
+import { Web3ModalInfo } from '../Web3ModalInfo'
+
+export function EthersModalInfo() {
+  const { isConnected, address, chainId } = useWeb3ModalAccount()
```
enesozturk, line 7: This will probably cause some hydration issues. Since we don't fully support SSR, the `isConnected` value will be different between the server and the client.

Instead, we could:
-  Force it to use client-side rendering: https://nextjs.org/docs/messages/react-hydration-error#possible-ways-to-fix-it
- Or render the `<Web3ModalInfo/>` component even though we don't have connection yet but there wouldn't be any info inside.

tomiir: added a ready check 🖖

PR #2199, opened 2024-04-30
refactor: swap business logic improvements

enesozturk:
unit testing, loading ui,

# Breaking Changes

N/A

# Changes

- feat:
- fix:
- chore:

# Associated Issues

closes #...


packages/core/src/utils/SwapApiUtil.ts
```diff
@@ -100,20 +101,22 @@ export const SwapApiUtil = {
...
+            : `${token.chainId}:${ConstantsUtil.NATIVE_TOKEN_ADDRESS}`,
+          decimals: parseInt(token.quantity.decimals, 10),
+          logoUri: token.iconUrl,
+          eip2612: false,
+          quantity: token.quantity,
+          price: token.price,
+          value: token.value
+        } as SwapTokenWithBalance
+      }) || []
+    )
```
tomiir, line 120: why not using a spread?
```
      balances?.map(token => {
        return {
          ...token,
          address: token?.address
            ? token.address
            : `${token.chainId}:${ConstantsUtil.NATIVE_TOKEN_ADDRESS}`,
          decimals: parseInt(token.quantity.decimals, 10),
          logoUri: token.iconUrl,
          eip2612: false,
        } as SwapTokenWithBalance
      }) || []
    )
```

packages/core/tests/controllers/SwapController.test.ts
```diff
@@ -1,78 +1,72 @@
...
+  SwapController.state.sourceTokenPriceInUSD = _sourceToken.price
+  SwapController.setToToken(_toToken)
+  SwapController.state.toTokenPriceInUSD = _toToken.price
+  SwapController.state.toTokenAmount = SwapCalculationUtil.getToTokenAmount({
+    sourceToken: _sourceToken,
+    sourceTokenAmount: '1',
+    sourceTokenPrice: _sourceToken.price,
+    toToken: _toToken,
+    toTokenPrice: _toToken.price
+  })
```
tomiir, line 39: no way to actually set this using regular `SwapController` flows?

packages/core/tests/controllers/SwapController.test.ts
```diff
@@ -1,78 +1,72 @@
...
-    expect(SwapController.state.sourceToken?.address).toEqual(sourceToken.address)
-  })
-
-  it('should calculate gas price in Ether and USD as expected', () => {
-    const gasPriceInEther = SwapController.calculateGasPriceInEther(gasLimit, gasFee)
-    const gasPriceInUSD = SwapController.calculateGasPriceInUSD(gasLimit, gasFee)
-
-    expect(gasPriceInEther).toEqual(0.068395033074)
-    expect(gasPriceInUSD).toEqual(0.06395499714651795)
+    expect(SwapController.state.sourceToken?.address).toEqual(networkToken.address)
```
tomiir, line 54: shouldn't you test the usage of the method here? 
feels weird to rely on the beforeAllHook for this 🤔

packages/scaffold/src/views/w3m-unsupported-chain-view/index.ts
```diff
@@ -65,6 +66,23 @@ export class W3mUnsupportedChainView extends LitElement {
   }
 
   // -- Private ------------------------------------------- //
+  private descriptionTemplate() {
+    if (this.swapUnsupportedChain) {
+      return html`
+        <wui-text variant="small-400" color="fg-200" align="center">
+          The swap feature doesn’t support your current network. Switch to an available option
+          following to continue.
```
tomiir, line 74: I'd remove the following

packages/scaffold/src/views/w3m-unsupported-chain-view/index.ts
```diff
@@ -65,6 +66,23 @@ export class W3mUnsupportedChainView extends LitElement {
...
+        </wui-text>
+      `
+    }
+
+    return html`
+      <wui-text variant="small-400" color="fg-200" align="center">
+        This app doesn’t support your current network. Switch to an available option following to
```
tomiir, line 81: same

packages/core/tests/controllers/SwapController.test.ts
enesozturk: Are you asking for setting the `toTokenAmount ` value or all other things? These are the values when the SwapController get initialized. They are set with different endpoint responses. The only thing extra I'm mocking that the user selected the `toToken` and calculated the `toTokenAmount`. It looks a bit weird that we are using `SwapCalculationUtil` in the SwapController test but if I wrap it with a function in `SwapController` that would be unnecessary.

packages/core/tests/controllers/SwapController.test.ts
enesozturk: This makes sure that when user opens the Swap page, the first token should be selected. The `beforeAll` sounds not completely reliable but we have to be sure that the first token is selected when the Swap is ready.

Maybe we should call `initializeSwapState` in every test?

packages/core/tests/controllers/SwapController.test.ts
tomiir: I guess the proper way would be to mock the API response directly. 
There's a couple of examples in the `ApiController` tests

tomiir (approved):

svenvoskamp (approved):

PR #2205, opened 2024-05-03
[Draft] feat: support web3js

Muhammad-Altabba:
This is currently just a draft.

This MR is to support using web3.js with web3modal.

# Breaking Changes

N/A

# Changes

- feat: support web3js
- chore: read `EMAIL_DOMAIN` from env variable (and default to `'web3modal.msdc.co'` if not provided).

# Associated Issues

N/A


glitch-txs:
Hi thanks for the effort, we won't be adding another package but rather simplify support into a single one for ethers and web3js

Muhammad-Altabba:
> Hi thanks for the effort, we won't be adding another package but rather simplify support into a single one for ethers and web3js

Thanks @glitch-txs,
We hope this PR gives some ideas on how web3.js could be used with web3modal. Let us know please if we need to provide support for the code changes that would simplify the support of ethers and web3.js into the anticipated single package.

And for now, I published a temporary package as a demo at: https://www.npmjs.com/package/web3modal-web3js

Best regards,

glitch-txs:
Thanks @Muhammad-Altabba! We are planning to migrate from ethers support to a more agnostic approach by exposing the wallet provider on its own so devs would be able to work/choose any ethereum library they wish to use

vittominacori:
> We are working on a general structure refactor for web3modal and as @glitch-txs mentioned we won't need to implement web3.js client explicitly and it'll be library agnostic.

@enesozturk where we could track that new updates? Will this refactor be integrated in Appkit?

enesozturk:
> > We are working on a general structure refactor for web3modal and as @glitch-txs mentioned we won't need to implement web3.js client explicitly and it'll be library agnostic.
> 
> 
> 
> @enesozturk where we could track those new updates? Will this refactor be integrated in Appkit?

@vittominacori we started working on the big refactor and will be opening the PRs in the upcoming weeks. Currently cannot share a link/doc or something, keep following for the updates please

PR #2210, opened 2024-05-06
refactor: render tooltip in root

enesozturk:
# Breaking Changes

Refactor the tooltip component to render it in the root of the w3m instead of in the component of it's used. This is an approach that allow us to have more capabilities when rendering tooltips. In some situations, we might need to render a tooltip with variety children's. 

### wui-tooltip-select

The existing wui-tooltip-select component, originally styled as a wallet feature button, will now support a broader range of components based on the specific use case.

### Why render on the root?

1) Overflow Handling: There are instances where the tooltip content is lengthy, causing it to overflow beyond the boundaries of the w3m component. With the previous implementation, the overflow was hidden by the w3m. Rendering at the root level resolves this issue.

2) Positioning and Control: Certain tooltips necessitate rendering outside the w3m component for proper positioning and control. This change allows for precise placement and management of such tooltips.

# Changes

- Refactor `wui-tooltip-select` to `wui-wallet-feature-button`, since it's styling for only for wallet feature buttons on the account page.
- Add `TooltipController` to manage tooltip state
- Add `w3m-tooltip-trigger` to handle a variety children as a tooltip wrapper
- Add `w3m-tooltip` component to render this on the root.


svenvoskamp (changes requested):
Good refactor! ⚡️ 

I'm not to keen on the name `wui-wallet-feature-button`. I think we need a more general name, so we could reuse it in the future for other things. Also please fix danger issues.

enesozturk:
@svenvoskamp Totally - I wasn't quite sure about its naming but at the same time, I'm not sure if it's a reusable component in different places since we don't have any similar components. Any suggestion for a name?

Taking a look danger issues right now!

enesozturk:
@svenvoskamp made a few changes danger.js checks for false errors.

packages/ui/src/utils/ThemeUtil.ts
```diff
@@ -45,6 +45,7 @@ function createRootStyles(themeVariables?: ThemeVariables) {
     core: css`
       @import url('https://fonts.googleapis.com/css2?family=Inter:wght@400;500;600;700&display=swap');
       :root {
+        --w3m-modal-width: 360px;
```
tomiir, line 48: did we check with design that this is ok? just wondering if designs expect a fixed width

svenvoskamp:
> @svenvoskamp Totally - I wasn't quite sure about its naming but at the same time, I'm not sure if it's a reusable component in different places since we don't have any similar components. Any suggestion for a name?
> 
> Taking a look danger issues right now!

Yes we don't use it anywhere else, but I like to keep the component naming a bit more general. Maybe something like `wui-icon-button`

svenvoskamp (approved):

packages/scaffold/src/partials/w3m-tooltip/index.ts
```diff
@@ -0,0 +1,68 @@
...
+
+    const topValue = this.triggerRect.top
+    const leftValue = this.triggerRect.left
+
+    this.style.cssText = `
+    --w3m-tooltip-top: ${topValue}px;
+    --w3m-tooltip-left: ${leftValue}px;
+    --w3m-tooltip-parent-width: ${this.triggerRect.width / 2}px;
+    --w3m-tooltip-display: ${this.open ? 'flex' : 'none'};
+    --w3m-tooltip-opacity: ${this.open ? 1 : 0};
```
tomiir, line 54: Bit of a nitpick:

I think we usually prefix these with `local`. 
I think w3m would be a bit confusing to know if it's a generic style vs a local one when modifying the css

packages/ui/src/utils/ThemeUtil.ts
enesozturk: @tomiir Yeah, we use 360px as modal width. The thing I did is moving it to variable instead using in hard code value in CSs. 

Using variable is good when you have to use modal width in different places. I.e: I'm calculating the `max-width` of tooltip depending on the modal width:

```css
    max-width: calc(var(--w3m-modal-width) - var(--wui-spacing-xl));
```

packages/scaffold/src/partials/w3m-tooltip/index.ts
enesozturk: This is a good point. I've thought the same but the `w3m-tooltip` is a global component in w3m. Like snackbar and rendered on the root. So there should be only one tooltip rendered, that's why I wanted to use `--w3m-tooltip-*` naming.

tomiir (approved):

PR #2213, opened 2024-05-06
fix: sign modal not appearing when going through siwe flow

tomiir:
# Breaking Changes

N/A

# Changes
- fix: sign modal not popping up when using SIWE + email


arein (commented):
curious: How will this be covered by an automated test? (I know PR is draft)

apps/laboratory/tests/siwe-email.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  await modalWalletPage.sign()
+  await modalWalletPage.approveSign()
+  await modalWalletValidator.expectAcceptedSign()
+
+  targetChain = 'Ethereum'
+  await modalWalletPage.openAccount()
+  await modalWalletPage.openSettings()
+  await modalWalletPage.switchNetwork(targetChain)
+  await modalWalletPage.page.waitForTimeout(1000)
```
enesozturk, line 45: Do we have to wait here? Doesn't it waiting util chain name to be on the screen by itself?

tomiir: Siwe Popup is async, it reacts to state changes on the account so it has a small delay. 
Will test it out as I've changed some things that might make it work regardless 🖖

apps/laboratory/tests/shared/fixtures/w3m-email-fixture.ts
```diff
@@ -25,3 +28,29 @@ export const testMEmail = timingFixture.extend<ModalFixture>({
...
+
+    const mailsacApiKey = process.env['MAILSAC_API_KEY']
+    if (!mailsacApiKey) {
+      throw new Error('MAILSAC_API_KEY is not set')
+    }
+    const email = new Email(mailsacApiKey)
+    const tempEmail = email.getEmailAddressToUse(testInfo.parallelIndex)
+
+    await modalPage.emailFlow(tempEmail, context, mailsacApiKey)
+    await modalPage.page.waitForTimeout(1000)
```
enesozturk, line 45: Same here, instead of putting timeouts we should wait for some elements to be visible. There should be a certain reason when we use timeouts and would better be commented IMO.

enesozturk (approved):
LGTM! Just left some comments

PR #2214, opened 2024-05-06
chore: detailed canary timings

chris13524:
Adds more metrics for specific things we want to time. In this case:
- Time for the QR code to display once requested by the user
- Time for the session proposal to be received by the wallet once URI is pasted and connect clicked
- Time for a signature request be received by the wallet once initiated via the app


apps/laboratory/tests/shared/fixtures/timing-fixture.ts
```diff
@@ -0,0 +1,11 @@
+import { test as base } from '@playwright/test'
+
+export type TimingRecords = { item: string; timeMs: number }[]
```
arein, line 3: elegant!

arein (approved):

PR #2219, opened 2024-05-07
refactor: add in-app browser check for mobile

enesozturk:
# Breaking Changes

Adde in-app browser checks to be sure that if the app is opened on any wallet's in-app browser, it should be detected as mobile as well


tomiir (approved):

packages/core/src/utils/CoreHelperUtil.ts
```diff
@@ -3,11 +3,30 @@ import { ConstantsUtil } from './ConstantsUtil.js'
...
+      )
+    }
+
+    return false
+  },
+
+  isMobile() {
+    if (typeof window !== 'undefined') {
+      return (
+        this.isInAppBrowser() ||
```
enesozturk, line 27: Adding this somehow breaks the e2e tests. Prob it's going into some unexpected conditions. Will check later

PR #2223, opened 2024-05-07
:deprecate - injected connector

svenvoskamp:
# Changes

- deprecate injected connector from wagmi / ethers5 / ethers as we use multi injected provider discovery (EIP 6963)
- show deprecate warning on `enableInjected` flag
- add `installed` tag to extension wallets


packages/ethers/src/client.ts
```diff
@@ -244,12 +232,6 @@ export class Web3Modal extends Web3ModalScaffold {
           return Boolean(window.ethereum)
         }
 
-        if (ethersConfig.injected) {
-          if (!window?.ethereum) {
-            return false
-          }
-        }
-
```
tomiir, line 252: do we still need this? asking cause this comes from `ethersConfig`

packages/ethers/src/utils/defaultConfig.ts
```diff
@@ -4,7 +4,6 @@ import { CoinbaseWalletSDK } from '@coinbase/wallet-sdk'
 
 export interface ConfigOptions {
   enableEIP6963?: boolean
-  enableInjected?: boolean
```
tomiir, line 7: I don't think we should add breaking changes by changing the interface. 
I'd say we leave the interface as-is but just not use the flag. 
We should probably emit a warning deprecation message instead.

apps/laboratory/src/pages/library/wagmi.tsx
```diff
@@ -25,7 +25,6 @@ const modal = createWeb3Modal({
   metadata: ConstantsUtil.Metadata,
   termsConditionsUrl: 'https://walletconnect.com/terms',
   privacyPolicyUrl: 'https://walletconnect.com/privacy',
-  customWallets: ConstantsUtil.CustomWallets,
```
Cali93, line 28: Are we removing this because we will always display installed wallets ?

packages/ethers/src/client.ts
```diff
@@ -565,7 +544,7 @@ export class Web3Modal extends Web3ModalScaffold {
     localStorage.removeItem(EthersConstantsUtil.WALLET_ID)
     EthersStoreUtil.reset()
 
-    if (providerType === 'injected' || providerType === 'eip6963') {
+    if (providerType === 'eip6963') {
```
Cali93, line 547: Just double checking that the provider type isn't persisted in `localStorage` to avoid people getting in a broken state.

Cali93 (commented):
Are current tests covering this ? If not, can we add them ?

apps/laboratory/src/pages/library/wagmi.tsx
svenvoskamp: Will add them back, removed them for testing purposes

packages/ethers/src/client.ts
svenvoskamp: If a users latest connection was with the injected connector, it will lose the connection. They need to reconnect (with EIP6963). They won't get into a broken state where they can't disconnect.

packages/ethers/src/client.ts
svenvoskamp: We don't need this anymore

packages/ethers5/src/utils/defaultConfig.ts
```diff
@@ -4,47 +4,23 @@ import { CoinbaseWalletSDK } from '@coinbase/wallet-sdk'
...
 export function defaultConfig(options: ConfigOptions) {
-  const {
-    enableEIP6963 = true,
-    enableInjected = true,
-    enableCoinbase = true,
-    metadata,
-    rpcUrl,
-    defaultChainId
-  } = options
+  const { enableEIP6963 = true, enableCoinbase = true, metadata, rpcUrl, defaultChainId } = options
```
tomiir, line 18: I'd add a `console.warn` here if user is enabling the `enableInjected` flag

tomiir (approved):

PR #2229, opened 2024-05-09
refactor/swap improvements

enesozturk:
- **refactor: update swap input styles and price state for totoken**
- **refactor: business logic and ui improvements**

# Breaking Changes

N/A

# Changes

- feat:
- fix:
- chore:

# Associated Issues

closes #...


packages/core/src/controllers/SwapController.ts
```diff
@@ -245,9 +259,14 @@ export const SwapController = {
   },
 
   switchTokens() {
+    if (state.initializing || !state.initialized) {
```
tomiir, line 262: do we need this check in other methods ?

packages/core/src/controllers/SwapController.ts
```diff
@@ -675,6 +698,26 @@ export const SwapController = {
...
+    const insufficientNetworkTokenForGas = SwapCalculationUtil.isInsufficientNetworkTokenForGas(
+      state.networkBalanceInUSD,
+      state.gasPriceInUSD
+    )
+
+    if (insufficientNetworkTokenForGas || isInsufficientSourceTokenForSwap) {
+      return true
+    }
+
+    return false
```
tomiir, line 721: `return insufficientNetworkTokenForGas || isInsufficientSourceTokenForSwap`

packages/core/src/controllers/SwapController.ts
enesozturk: Nope, this this for disabling functionality when users click the switch button on the center because that button is still visible even thought the swap is initializing (when have skeletons view)

tomiir (approved):

PR #2236, opened 2024-05-10
Fix/send bugs

svenvoskamp:
# Changes

fix: https://linear.app/walletconnect/issue/APKT-338/sending-small-amounts-leads-to-weird-ui-issue
fix: resolved issue when sending native tokens was not possible on wagmi (wagmi released a fix for us) 
fix: resolved issue that metadata didn't show when using ethers + siwe


packages/scaffold/src/views/w3m-account-settings-view/index.ts
```diff
@@ -246,9 +246,9 @@ export class W3mAccountSettingsView extends LitElement {
     await ConnectionController.reconnectExternal(emailConnector)
 
     this.text =
-      this.preferredAccountType === W3mFrameRpcConstants.ACCOUNT_TYPES.SMART_ACCOUNT
-        ? 'Switch to your smart account'
-        : 'Switch to your EOA '
+      accountTypeTarget === W3mFrameRpcConstants.ACCOUNT_TYPES.SMART_ACCOUNT
```
enesozturk, line 249: Oh how we didn't notice before 🤔

packages/core/src/controllers/SendController.ts
```diff
@@ -91,7 +91,7 @@ export const SendController = {
 
   sendToken() {
     if (this.state.token?.address && this.state.sendTokenAmount && this.state.receiverAddress) {
-      SendController.sendERC20Token({
+      this.sendERC20Token({
```
enesozturk, line 94: If we are not calling the `sendToken` from a different context (like passing it to another object's callback etc.) this shouldn't make any difference, curious what was the issue here 🤔

enesozturk (approved):

packages/scaffold/src/views/w3m-account-settings-view/index.ts
svenvoskamp: Seemed to work great locally, but in lab it didn't. This should solve it.

packages/core/src/controllers/SendController.ts
svenvoskamp: There was no issue here, but I think this is more clean.

PR #2242, opened 2024-05-10
chore: Test Cases table and missing SA tests

lukaisailovic:
# Changes

-  adds basic test cases table


tomiir:
@lukaisailovic want to merge this and we update it later or will u update with new SIWE + SA test support?

enesozturk:
@lukaisailovic @tomiir I think we could merge this?

lukaisailovic:
@tomiir Can you please update with current testinng state?

enesozturk:
@tomiir @lukaisailovic I think we could close this. IMO it doesn't make sense to keep track of the test cases manually - this will tend to be expired very soon. Engineers shouldn't be responsible of maintaining this list. They should be responsible of adding tests for anything they developed. Since we started doing this. This table doesn't seem useful to me. Wdyt?

PR #2250, opened 2024-05-12
feat: phantom deeplink for mobile app browsers

phoenixVS:
# Breaking Changes

N/A

# Changes

- feat:
- fix:
- chore:

# Associated Issues

closes #...


packages/solana/src/client.ts
```diff
@@ -227,6 +229,17 @@ export class Web3Modal extends Web3ModalScaffold {
       }
     })
 
+    EventsController.subscribe(state => {
+      if (state.data.event === 'SELECT_WALLET' && state.data.properties?.name === 'Phantom') {
+        const isMobile = CoreHelperUtil.isMobile()
+        if (isMobile && !window.phantom) {
```
enesozturk, line 235: In server-side frameworks like Next.js, reaching to `window` object will possibly throw error since it's not available on the server side. So we should make an extra check for the window object before using it to make sure this condition is running only on the client side. @phoenixVS

packages/solana/src/client.ts
```diff
@@ -227,6 +229,17 @@ export class Web3Modal extends Web3ModalScaffold {
...
+          const href = window.location.href
+          const ref = `${href.startsWith('https') ? 'https' : 'http'}://${href.split('/')[2]}`
```
enesozturk, line 237: For better readability, could we separate the parts and stay away from one-liner template literals with many conditions inside? This will make it more readable. 

```suggestion
      const href = window.location.href;
      const protocol = href.startsWith('https') ? 'https' : 'http';
      const host = href.split('/')[2];
      const ref = `${protocol}://${host}`;

      window.location.href = `https://phantom.app/ul/browse/${href}?ref=${ref}`;
```

glitch-txs:
We should store this on Cloud IMO, in case we need it for more wallets, and automate it

enesozturk:
> We should store this on Cloud IMO, in case we need it for more wallets, and automate it

@glitch-txs Do we have common way to handle deeplinks across wallets? Phantom have this approach of deeplinks:

```js
https://phantom.app/ul/browse/${href}?ref=${ref}
```

What about the others? What could we store in our Cloud/DB?

packages/solana/src/client.ts
```diff
@@ -229,6 +231,19 @@ export class Web3Modal extends Web3ModalScaffold {
       }
     })
 
+    EventsController.subscribe(state => {
+      if (state.data.event === 'SELECT_WALLET' && state.data.properties?.name === 'Phantom') {
+        const isMobile = CoreHelperUtil.isMobile()
+        if (isMobile && typeof window !== 'undefined' && !window.phantom) {
```
enesozturk, line 237: @phoenixVS Could we use `CoreHelperUtil.isClient()` instead of `typeof window !== 'undefined'` sorry for late notice I forgot about it.

enesozturk (approved):

PR #2266, opened 2024-05-14
:feat social login - 1 - UI 

svenvoskamp:
# Changes

Breaking this PR (https://github.com/WalletConnect/web3modal/pull/2057) into smaller parts. This is the part of the UI this includes the following changes: 

* Adding `wui-list-button`
* Adding `wui-list-social`
* Adding several svg's 
* Added stories for new composites 
* Small tweaks to other components / composites to match design


packages/ui/src/components/wui-icon/index.ts
```diff
@@ -131,14 +133,15 @@ const svgOptions: Record<IconType, TemplateResult<2>> = {
...
+  twitter: xSvg,
   twitterIcon: twitterIconSvg,
   verify: verifySvg,
   verifyFilled: verifyFilledSvg,
   wallet: walletSvg,
   walletConnect: walletConnectSvg,
   walletPlaceholder: walletPlaceholderSvg,
-  warningCircle: warningCircleSvg
+  warningCircle: warningCircleSvg,
+  x: xSvg
```
tomiir, line 144: do we need both twitter and X?

packages/ui/src/composites/wui-input-text/styles.ts
```diff
@@ -102,6 +101,17 @@ export default css`
     left: var(--wui-spacing-l);
   }
 
+  .wui-size-mdl {
+    padding: 17.25px var(--wui-spacing-m) 17.25px var(--wui-spacing-m);
```
enesozturk, line 105: If these kind of values come from the design file, I think we should ask why we don't have regular spacing values. Because sometimes, the designer might miss some values and those could be unintentional. 

We should be following the practice of 2x (or similar) spacing system

packages/ui/src/composites/wui-input-text/styles.ts
```diff
@@ -102,6 +101,17 @@ export default css`
...
+  }
+  wui-icon + .wui-size-mdl,
+  wui-loading-spinner + .wui-size-mdl {
+    padding: 17.25px var(--wui-spacing-3xl) 17.25px 40px;
```
enesozturk, line 109: If possible, let's use variable for `40px` as well. (Prob we don't have it but not sure)

tomiir (approved):
minor comments, looks good overall!

packages/ui/src/composites/wui-list-button/index.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  // -- State & Properties -------------------------------- //
+  @property() public text = ''
+
+  @property({ type: Boolean }) public disabled = false
+
+  // -- Render -------------------------------------------- //
+  public override render() {
+    return html`
+      <button ?disabled=${this.disabled} ontouchstart>
```
enesozturk, line 20: Why do we have `ontouchstart`? I saw this on Tooltip component as well? What it helps for?

packages/ui/src/composites/wui-list-button/styles.ts
```diff
@@ -0,0 +1,18 @@
+import { css } from 'lit'
+
+export default css`
+  button {
+    column-gap: var(--wui-spacing-s);
+    padding: 16.5px var(--wui-spacing-l) 16.5px var(--wui-spacing-xs);
```
enesozturk, line 6: Same for `16.5px`

packages/ui/src/composites/wui-list-button/styles.ts
```diff
@@ -0,0 +1,18 @@
...
+    width: 100%;
+    background-color: var(--wui-gray-glass-002);
+    border-radius: var(--wui-border-radius-xs);
+    color: var(--wui-color-fg-100);
+    justify-content: center;
+    align-items: center;
+  }
+  button:disabled {
```
enesozturk, line 14: Let's separate styles with new line

```suggestion

  button:disabled {
```

packages/ui/src/composites/wui-list-social/index.ts
```diff
@@ -0,0 +1,55 @@
...
+  public override render() {
+    return html`
+      <button ?disabled=${this.disabled} ontouchstart>
+        <wui-logo logo=${this.logo}></wui-logo>
+        <wui-text
+          data-align=${this.align}
+          variant="paragraph-500"
+          color="inherit"
+          align=${this.align}
+          >${UiHelperUtil.capitalizeFirstLetter(this.name)}</wui-text
```
enesozturk, line 34: Can we use a CSS native solution here with `text-transform` property?

```CSS
text-transform: capitalize
```

packages/ui/src/composites/wui-list-social/styles.ts
```diff
@@ -0,0 +1,33 @@
...
+  button {
+    column-gap: var(--wui-spacing-s);
+    padding: 7px var(--wui-spacing-l) 7px var(--wui-spacing-xs);
+    width: 100%;
+    justify-content: flex-start;
+    background-color: var(--wui-gray-glass-002);
+    border-radius: var(--wui-border-radius-xs);
+    color: var(--wui-color-fg-100);
+  }
+  wui-text::first-letter {
```
enesozturk, line 13: Maybe same for these styles to separate them with new line?

packages/ui/src/components/wui-icon/index.ts
svenvoskamp: Magic still uses `twitter` as naming, so I think it is important to have both for now. We can adjust this later

packages/ui/src/utils/UiHelperUtil.ts
```diff
@@ -147,5 +147,8 @@ export const UiHelperUtil = {
       maximumFractionDigits: decimals,
       minimumFractionDigits: decimals
     })
+  },
+  capitalizeFirstLetter(string: string) {
```
enesozturk, line 151: Let's go with CSS native approach instead of string manipulation with JS?

packages/ui/src/composites/wui-list-button/index.ts
svenvoskamp: To enable touch animations on `mobile` devices

packages/ui/src/composites/wui-input-text/styles.ts
svenvoskamp: I agree, I think this happens in other places in our composites/components as well. The buttons and `wui-input-email` have a height of `56px`, changing these values will change this. So I think we should keep it as it is and revisit when we will split the ui package as part of the bigger refactor.

Cali93 (approved):
LGTM 💪 
I'm more interested on the magic integration part. Is [this PR](https://github.com/WalletConnect/secure-web3modal/pull/108) the right one ?

svenvoskamp:
> LGTM 💪 I'm more interested on the magic integration part. Is [this PR](https://github.com/WalletConnect/secure-web3modal/pull/108) the right one ?

Will follow up with a PR for our wallet package, the PR that you mention is indeed the magic integration part on the secure site

PR #2271, opened 2024-05-15
feat: adding erc7677 support on laboratory

KannuSingh:


arein (commented):
can you add some automated tests?

apps/laboratory/src/utils/EIP5792Utils.ts
```diff
@@ -19,7 +26,11 @@ export function getAtomicBatchSupportedChainInfo(
     const walletCapabilities = walletCapabilitiesString && parseJSON(walletCapabilitiesString)
     const accountCapabilities = walletCapabilities[address]
     const chainIds = accountCapabilities
-      ? Object.keys(accountCapabilities).map(chainIdAsHex => Number(chainIdAsHex))
+      ? Object.keys(accountCapabilities)
+          .filter(
+            chainIdAsHex => accountCapabilities[chainIdAsHex]?.[capability]?.supported === true
```
tomiir, line 31: do we need to check for `true` explicitly? 
is the expected type other than `boolean?`

apps/laboratory/src/utils/EIP5792Utils.ts
```diff
@@ -19,7 +26,11 @@ export function getAtomicBatchSupportedChainInfo(
     const walletCapabilities = walletCapabilitiesString && parseJSON(walletCapabilitiesString)
```
tomiir, line 26: I'd move the capabilities calculation to the upper context and keep the same interface for viem and ethers. wyt?
would that remove the need to duplicate the function? (even if not, I think it's worth)

apps/laboratory/src/components/Ethers/EthersSendCallsWithPaymasterServiceTest.tsx
```diff
@@ -0,0 +1,152 @@
...
+    return (
+      <Text fontSize="md" color="yellow">
+        Wallet does not support this feature
+      </Text>
+    )
+  }
+  if (paymasterServiceSupportedChains.length === 0) {
+    return (
+      <Text fontSize="md" color="yellow">
+        Account does not support paymaster service feature
```
tomiir, line 121: why the account? isn't this related to the supported chains by the wallet? Or account = wallet ? 
If that's the case I'd unify the terminology on the messages

tomiir (commented):
Minor comments related to code style mainly. let me know what u think! 
else looks good 💪

apps/laboratory/src/components/Ethers/EthersSendCallsWithPaymasterServiceTest.tsx
KannuSingh: capability is based on account address and chain, it can be possible to have other account address supporting this capability for same chain.

apps/laboratory/src/utils/EIP5792Utils.ts
KannuSingh: expected type is boolean, but because it getting serialized and some might set it as a string "true", but from ERC it specifically say to ignore if it not boolean.
>The only valid JSON-RPC values for this capability’s only member, supported, are true or false; if a returned value for supported is typed as a string or number, it SHOULD be considered malformed.

apps/laboratory/src/utils/EIP5792Utils.ts
KannuSingh: As per ERC the capability type is defined as
>type GetCapabilitiesResult = Record<`0x${string}`, <Record<string, any>>; // Hex chain id

but viem is returing the chainId as number i.e 
>type GetCapabilitiesResult = Record<number, <Record<string, any>>;

That is why, I tried ethers implementation as per ERC and didn't create the same interface for viem and ethers, I can process the data  and then create the common interface though.

tomiir: would make sense but don't think it's mandatory. would make the code nicer tho. Will drop an approve, if you agree  the abstraction makes sense let's do it else just go ahead and merge 🖖

tomiir (approved):

tomiir (approved):
🔥

enesozturk (approved):

PR #2273, opened 2024-05-15
refactor(swap): price impact calculation, beta tag

enesozturk:
# Changes

- Fix price impact calculation
- Add quote request for `dstAmount`
- Add `Beta` tag to header for Swap screens


packages/ethers/src/client.ts
```diff
@@ -306,6 +306,7 @@ export class Web3Modal extends Web3ModalScaffold {
 
       signMessage: async (message: string) => {
         const provider = EthersStoreUtil.state.provider
+        let hexedMessage = message
```
enesozturk, line 309: Should be coming from rebasing, remove

packages/scaffold/src/partials/w3m-swap-details/index.ts
```diff
@@ -1,40 +1,66 @@
...
 
-  @property() public sourceTokenSymbol?: string
+  @state() public sourceToken = SwapController.state.sourceToken
+
+  @state() public sourceTokenPrice = SwapController.state.sourceTokenPrice
+
+  @state() public toToken = SwapController.state.toToken
 
-  @property() public sourceTokenPrice?: number
+  @state() public toTokenAmount = SwapController.state.toTokenAmount
```
enesozturk, line 28: Since this component is a partial, it could reach SwapController and we don't have to pass properties. This makes the using it much easier

packages/siwe/src/client.ts
```diff
@@ -13,7 +13,8 @@ import {
   ConnectionController,
   RouterUtil,
   RouterController,
-  StorageUtil
+  StorageUtil,
```
enesozturk, line 16: Should be coming from rebasing, remove

packages/core/src/controllers/SwapController.ts
```diff
@@ -231,7 +232,7 @@ export const SwapController = {
   },
 
   setToTokenAmount(amount: string) {
-    state.toTokenAmount = amount
+    state.toTokenAmount = amount ? NumberUtil.formatNumberToLocalString(amount, 6) : ''
```
tomiir, line 235: what is `6` ?  can we not hardcode it?

packages/core/src/controllers/SwapController.ts
```diff
@@ -464,14 +466,30 @@ export const SwapController = {
...
+      gasPrice: state.gasFee,
+      amount: amountDecimal.toString()
     })
 
+    if (!quoteResponse?.quotes?.[0]) {
+      return
+    }
+
+    const toTokenAmount = NumberUtil.bigNumber(quoteResponse?.quotes?.[0].toAmount)
+      .dividedBy(10 ** toToken.decimals)
```
tomiir, line 488: is there an issue with dividing and getting a decimal here ? (since we are dealing with bignumbers)

packages/core/src/utils/SwapApiUtil.ts
```diff
@@ -22,6 +26,36 @@ export type TokenInfo = {
...
+    accountAddress,
+    gasPrice
+  }: {
+    fromTokenAddress: string
+    toTokenAddress: string
+    amount: string
+    accountAddress: string
+    gasPrice: string
+  }) {
+    const apiUrl = new URL(`${ONEINCH_API_BASE_URL}/swap/v6.0/137/quote`)
```
tomiir, line 42: why do we need to instantiate a new api her?

packages/scaffold/src/partials/w3m-swap-input/index.ts
```diff
@@ -75,31 +77,51 @@ export class W3mSwapInput extends LitElement {
...
+    const pasteKey = event.key === 'v' || event.key === 'V'
+    const cutKey = event.key === 'x' || event.key === 'X'
+
     const isComma = event.key === ','
     const isDot = event.key === '.'
     const isNumericKey = event.key >= '0' && event.key <= '9'
     const currentValue = this.value
 
-    if (!isNumericKey && !allowedKeys.includes(event.key) && !isDot && !isComma) {
+    // If command/ctrl key is not pressed, doesn't allow for a, c, v
```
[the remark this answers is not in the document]
tomiir: is it worth abstracting this?

packages/scaffold/src/views/w3m-swap-preview-view/index.ts
```diff
@@ -95,6 +103,12 @@ export class W3mSwapPreviewView extends LitElement {
   }
 
   // -- Private ------------------------------------------- //
+  private refreshTransaction() {
+    this.interval = setInterval(() => {
+      SwapController.getTransaction()
+    }, 10_000)
```
tomiir, line 109: what's the ttl on the api side?

packages/core/src/utils/SwapApiUtil.ts
enesozturk: This shouldn't be here anymore

enesozturk: Saw forgot to push lol

packages/scaffold/src/partials/w3m-swap-input/index.ts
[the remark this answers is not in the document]
enesozturk: Would worth considering actually - one concern is that this function have Swap related params like `target`. Somehow I need to refactor the parent component and this one as well. Will take a look

packages/core/src/controllers/SwapController.ts
enesozturk: Nope - this is an expected calculation that we are multiplying or dividing the amount values with the decimal value of the related token, we need to do this to make BigInt to Float

packages/scaffold/src/views/w3m-swap-preview-view/index.ts
enesozturk: Would be nice to validate with Max for best number to do this. This is also a UX question. The user should get the up-to-date swap values (token prices, quote, gas price etc.) periodically, I'm not sure if longer duration would create bad UX

packages/scaffold/src/partials/w3m-swap-input/index.ts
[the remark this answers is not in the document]
enesozturk: Totally - Created InputUtil in `common` so later we might need different kind of input manipulator functions

tomiir (approved):

PR #2274, opened 2024-05-15
:feat social login - 3 - AuthConnector 

svenvoskamp:
# Changes

Breaking this PR (https://github.com/WalletConnect/web3modal/pull/2057) into smaller parts. This is the part is to extend the `emailConnector` so it also support social login

* Rename `emailConnector` to `authConnector`
* Rename `emailProvider` to  `authProvider`
* Add `auth` config to `wagmi` and `ethers` config
* Change `email` naming` to `auth` naming in clients 
* Rename `EMAIL` to `AUTH` in `connectorType`
* Rename `EMAIL_CONNECTOR_ID` to `AUTH_CONNECTOR_ID`


tomiir (approved):
looks good overall! 
seems weird to split the feature flags for auth and enableEmail but think there's no other way around it if we want to preserve the interface

svenvoskamp:
> looks good overall!
> seems weird to split the feature flags for auth and enableEmail but think there's no other way around it if we want to preserve the interface

Agree, but to avoid breaking changes this is the way to go. We will add `email` to the flag when we do a major version change.

PR #2277, opened 2024-05-15
refactor/theme variables and v4 ui updates

enesozturk:
# Changes

This PR introduces changes to our design system theme variables. 

### Theme variable namings

We are using inconsistent naming convention for the colors that some of the starts with `--wui-color-*` and some of the starts with `--wui-*`. At some certain situations. This is making things a bit hard. 

Example case: at some components we might need set the local color with properties and we are following this approach which is easy and short way to make conditional stylings. Once we have different naming conventions for colors, we need to make extra conditions while setting the `--local-color` and this makes things harder.
```js
public override render() {
    this.style.cssText = `
      --local-color: ${`var(--wui-color-${this.color});`}
      ...
    `
    ...
  }
```

### Components updates

We have updates to many components at Figma, but buttons are one of the main components in our UI and they have now different colors, active & focus states and different variants. These components has been refactored completely:

- `wui-button`
- `wui-chip-button`

### Updating page containers

In V4, we removed bottom border of the header, and that's why the content of the page is closer to header and there is no gap between. This PR also includes to padding refactors of the page container `wui-flex` elements

### Theme `elementStyle` 

In `ThemeUtils` file, we have `elementStyle` styling that we use this in almost every component. This styles including many stylings for `a`, and `button`s, and their different states. Even though this seems good. While refactoring the `wui-button` realized that overriding button stylings in both places creates overhead. 

When we create a web component which include `<button/>` inside, we should decide it's stylings inside that new web component. That button shouldn't use the common styles from `elementStyle` object.

## Example playground run

https://github.com/WalletConnect/web3modal/assets/19428358/ab8162a1-6479-4a51-bdb3-036758179796


apps/gallery/stories/theme/Colors.mdx
```diff
@@ -89,12 +98,12 @@ import { Meta, ColorPalette, Title, ColorItem } from '@storybook/blocks'
...
-    title="Blue"
+    title="Accent"
     subtitle="80 | 90 | 100"
     colors={{
-      'wui-color-blue-80': '#2978CC',
-      'wui-color-blue-90': '#2D7DD2',
-      'wui-color-blue-100': '#3396FF'
+      'wui-accent-080': 'hsla(230, 59%, 55%, 1)',
+      'wui-accent-090': 'hsla(231, 76%, 61%, 1)',
+      'wui-accent-100': 'hsla(230, 100%, 67%, 1)'
```
tomiir, line 106: why use hsla here? cannot find a ref to these values in the colors figma (maybe I'm blind?)

packages/scaffold/src/partials/w3m-input-token/index.ts
```diff
@@ -48,13 +48,13 @@ export class W3mInputToken extends LitElement {
         text=${this.token.symbol}
         imageSrc=${this.token.iconUrl}
         @click=${this.handleSelectButtonClick.bind(this)}
-        >Select token</wui-token-button
```
tomiir, line 51: Is this ok?

packages/ui/src/composites/wui-button/styles.ts
```diff
@@ -7,60 +7,141 @@ export default css`
...
+    box-shadow: inset 0 0 0 1px var(--wui-color-success-glass-010);
   }
 
-  button[data-size='md'][data-icon-right='true'][data-icon-left='false'] {
-    padding: 8.2px var(--wui-spacing-s) 9px var(--wui-spacing-l);
+  button[data-variant='neutral'] {
+    background: transparent;
+    color: var(--wui-color-fg-100);
+    border: none;
+    box-shadow: inset 0 0 0 1px var(--wui-color-gray-glass-005);
```
tomiir, line 75: aren't  `box-shadow`s abstracted on a design-level? 
Would be nice to not repeat this (and other shadows) all over the place

enesozturk: Yeap, the `w3m-input-token` doesn't getting children

packages/scaffold/src/partials/w3m-input-token/index.ts
enesozturk: Yeap, the w3m-input-token doesn't getting children

packages/ui/src/composites/wui-button/styles.ts
enesozturk: `box-shadow` is used for border functionality since we don't have inset borders and different variants have different shadow (border) colors. Moving them to a single place would be nice tbh. I'd say let's do that improvement in another PR

tomiir (approved):

PR #2283, opened 2024-05-17
Add Custom Wallet Option on Labs

glitch-txs:
# Breaking Changes

N/A

# Changes

- feat: Add Custom Wallet Option on Labs

image

- fix:
- chore:


arein:
awesome!

apps/laboratory/src/layout/CustomWallet.tsx
```diff
@@ -0,0 +1,103 @@
...
+    name: '',
+    image_url: '',
+    mobile_link: '',
+    desktop_link: '',
+    webapp_link: ''
+  })
+
+  function handleCustomWallet() {
+    localStorage.setItem(CUSTOM_WALLET, JSON.stringify(customWallet))
+    location.reload()
```
enesozturk, line 34: Why do we need to reload? Is it something we might need to improve on w3m?

enesozturk (approved):

apps/laboratory/src/layout/CustomWallet.tsx
glitch-txs: to re-instantiate Web3Modal

tomiir (approved):

PR #2227, opened 2024-05-08
added test for sendCalls and getCallsStatus

KannuSingh:
Added test component in laboratory for ethers and wagmi for EIP5792 methods sendCalls and getCallsStatus

PR #2285, opened 2024-05-20
fix: revert "added test for sendCalls and getCallsStatus (#2227)"

enesozturk:
This reverts commit 169c1a92b61742ab6f84b0f906e75e69d01d7644.

# Breaking Changes

We have an issue with `@wagmi/core` package that when we use it on lab as dependency, it's breaking our current auth connectors and email login is not available. We need to investigate what is causing it and we need to make sure #2227 is not breaking it.

# Associated Issues

- [Slack thread](https://walletconnect.slack.com/archives/C03RVH94K5K/p1716193735413749)
- [Slack thread](https://walletconnect.slack.com/archives/C03RVH94K5K/p1716013956301549)


tomiir:
This is the solution, no need to revert:
https://github.com/WalletConnect/web3modal/pull/2290

Will check that preview works and close this if it's solved 🖖

enesozturk:
Been fixed in https://github.com/WalletConnect/web3modal/pull/2290

tomiir (approved):

PR #2289, opened 2024-05-20
fix: disconnect logic for EIP6963 & Injected provider types for @web3modal/ethers

### LINKED ISSUE #2288: [bug] Modal doesn't properly disconnect when disconnecting - @web3modal/ethers

https://lab.web3modal.com/library/ethers/

### Summary

The `@web3modal/ethers` library is not handling the disconnect logic correctly. When selecting MetaMask from the modal, the MetaMask extension does not pop up, and it auto-connects the wallet instead. 

### Steps to Reproduce
1. **Open the WalletConnect Modal**:
   - Launch the modal to connect to a wallet.

2. **Select MetaMask**:
   - Select MetaMask from the modal options.
3. **Select a Wallet**:
   - Select a Wallet from the Metamask Pop up .

4. **Disconnect and Try to Connect Again**:
   - Attempt to disconnect the MetaMask wallet.
   - Open the modal again to select MetaMask.
   - The wallet should connect automatically.

### Expected Behavior
Each time the WalletConnect modal is opened, it should prompt the MetaMask extension to pop up and allow the user to select the wallet they want to connect to. This should happen every time the modal is launched, ensuring the user can switch between different wallets as needed.

### Actual Behavior
When selecting MetaMask, the extension does not pop up, and the wallet auto-connects. This prevents the user from selecting a different wallet each time, which is necessary for certain use cases.

@web3modal/ethers

hmzakhalid:
# Changes
- fix: disconnect logic for EIP6963 & Injected provider types in packages/ethers

# Associated Issues

closes #2288


hmzakhalid:
Hey @tomiir, Can you take a look at this Please?

tomiir:
hey @hmzakhalid ! ty for the contribution. 
I'm bumping this up with the rest of the team as not sure we want to add more clicks to the connection flow. 
I'm thinking maybe a proper way to handle this would be with a config flag on the ethers configuration that would dictate if we should or should not revoke the permissions. Agree there are cases where this would be a better ux, but would like input from @glitch-txs @enesozturk  as well. 

Code looks good overall, would try-catch the requests to the wallet in case they are not supported to prevent unexpected behaviors but otherwise looks good!

packages/ethers/src/client.ts
```diff
@@ -316,6 +313,39 @@ export class Web3Modal extends Web3ModalScaffold {
...
+          }[]
+          const ethAccountsPermission = permissions.find(
+            permission => permission.parentCapability === 'eth_accounts'
+          )
+
+          if (ethAccountsPermission) {
+            await provider.request({
+              method: 'wallet_revokePermissions',
+              params: [{ eth_accounts: {} }]
+            })
```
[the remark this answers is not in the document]
hmzakhalid: added the try-catch block. but also not sure if we need the, I assumed all wallets would support these methods.

hmzakhalid:
> hey @hmzakhalid ! ty for the contribution. I'm bumping this up with the rest of the team as not sure we want to add more clicks to the connection flow. I'm thinking maybe a proper way to handle this would be with a config flag on the ethers configuration that would dictate if we should or should not revoke the permissions. Agree there are cases where this would be a better ux, but would like input from @glitch-txs @enesozturk as well.
> 
> Code looks good overall, would try-catch the requests to the wallet in case they are not supported to prevent unexpected behaviors but otherwise looks good!

Thank you for the feedback and for bringing this up with the team. I'd like to point out that this should be the default behavior when the disconnect button is clicked. 
- Users expect that clicking a disconnect button will fully disconnect their wallet from the site, including revoking any permissions. This is a standard behavior that users are familiar with, and it aligns with their expectations. 
- The `wagmi` connector, for instance, ensures that the wallet is fully disconnected from the site. 
- If the wallet is not fully disconnected, it can lead to confusion and frustration, as users might not understand why they are still connected or why the wallet does not prompt them to reconnect properly. (As it happened in my case).

Would love feedback from the rest of the team.

tomiir:
Been re-reading this. I'm on board. Verified and it matches wagmi's behavior. 
Would love to get confirmation from team and we could move ahead with it 💪

hmzakhalid:
Hey @svenvoskamp , Could you please take a look at this PR as well?

svenvoskamp (approved):
LGTM, could you please fix lint issues?

tomiir (approved):
🚀

packages/ethers/src/client.ts
```diff
@@ -319,6 +316,47 @@ export class Web3Modal extends Web3ModalScaffold {
           // eslint-disable-next-line no-negated-condition
         } else if (providerType === ConstantsUtil.AUTH_CONNECTOR_ID) {
           await this.authProvider?.disconnect()
+        } else if (providerType === ConstantsUtil.EIP6963_CONNECTOR_ID && provider) {
+          try {
+            const permissions = (await provider.request({ method: 'wallet_getPermissions' })) as {
```
enesozturk, line 321: Don't we have the types of the response here? Why linter showing that assertion is unnecessary? If we don't know the return type, maybe we should override the lint for these lines? @tomiir @svenvoskamp

hmzakhalid: Nope, the return type is a generic Promise<any> so ts can't infer the exact type of the response.
```js
request(req: W3mFrameTypes.RPCRequest): Promise<any>;
```

Also, i dont see the linting issue. its TS complaining in the line below `'permissions' is of type 'unknown'.`

hmzakhalid: Nvm, I just ran the linter and saw the issue

hmzakhalid: Fixed it, will wait for others to reply before resolving the conversation.

hmzakhalid:
Shouldn't this disconnect function also be updated? i suppose this is used in the `useDisconnect` hook
https://github.com/hmzakhalid/web3modal/blob/V4/packages/ethers/src/client.ts#L616-L637

@enesozturk @svenvoskamp @tomiir

enesozturk:
> Shouldn't this disconnect function also be updated? i suppose this is used in the `useDisconnect` hook https://github.com/hmzakhalid/web3modal/blob/V4/packages/ethers/src/client.ts#L616-L637
> 
> @enesozturk @svenvoskamp @tomiir

@hmzakhalid That's a good point - yes. The public `disconnect()` function is for hooks and the JS client and should do the same thing as `connectionControllerClient.disconnect` does.

hmzakhalid:
> > Shouldn't this disconnect function also be updated? i suppose this is used in the `useDisconnect` hook https://github.com/hmzakhalid/web3modal/blob/V4/packages/ethers/src/client.ts#L616-L637
> > @enesozturk @svenvoskamp @tomiir
> 
> @hmzakhalid That's a good point - yes. The public `disconnect()` function is for hooks and the JS client and should do the same thing as `connectionControllerClient.disconnect` does.

Thought so, pushed the updated code.

tomiir:
Final request: given we are repeating the same thing 3 times, can we abstract it to a method/util?
Basically something like `disconnectProvider({ provider, providerType }) => void` 🙏 
Else looks good 🖖 🚀

glitch-txs (approved):

hmzakhalid:
Bumping this Up

hmzakhalid:
Bumping this up @glitch-txs @tomiir

PR #2291, opened 2024-05-20
fix: renovate config

chris13524:
Fixes several issues with the config, based on my reviewing of the docs:
- `includePaths` must be a pattern matching package manager files (e.g. package.json), not a prefix
- `matchPackageNames` does not support wildcards

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -9,22 +9,22 @@
   ],
   "packageRules": [
     {
-      "matchPackageNames": ["*"],
+      "matchPackagePatterns": ["*"],
       "enabled": false
     },
     {
       "matchPackageNames": [
         "@walletconnect/utils",
-        "@web3modal/*",
         "wagmi",
         "@wagmi/connectors",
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -9,22 +9,22 @@
   ],
   "packageRules": [
     {
-      "matchPackageNames": ["*"],
+      "matchPackagePatterns": ["*"],
```
tomiir, line 12: do we need it if it's wildcard? 
can't tell from docs if it's required or not 
`Use this field if you want to have one or more package names patterns in your package rule.` makes it seem optional

tomiir (approved):

renovate.json
chris13524: Tested it on a fork and I got this: 
image

chris13524:
Confirmed this config works on a fork:

image

bkrem (approved):

PR #2297, opened 2024-05-21
feat: coinbase smart accounts

tomiir:
# Breaking Changes

N/A

# Changes

- feat: upgrade to `wagmi/connectors@5` to include Coinbase Smart Accounts in the SDK. Add specific coinbase wallet connector ui component. Update `@coinbase/sdk` to V4 on ethers and ethers5
- fix:
- chore:


examples/vue-wagmi/package.json
```diff
@@ -7,8 +7,8 @@
     "build:examples": "vite build"
   },
   "dependencies": {
-    "@wagmi/connectors": "4.3.9",
-    "@wagmi/core": "2.9.7",
+    "@wagmi/connectors": "5.0.3",
```
enesozturk, line 10: Seems like a major version upgrade. Do we sure this won't break anything or did we make related changes? Do we have a migration source for it?

packages/wagmi/src/utils/defaultWagmiCoreConfig.ts
```diff
@@ -60,9 +60,10 @@ export function defaultWagmiConfig({
   if (enableCoinbase !== false) {
     connectors.push(
       coinbaseWallet({
+        version: '4',
         appName: metadata?.name ?? 'Unknown',
         appLogoUrl: metadata?.icons[0] ?? 'Unknown',
-        enableMobileWalletLink: true
+        preference: 'all'
```
enesozturk, line 66: Since this `preference` doesn't tell what exactly it is, could we add a JSDoc for it and reference to their docs?
```suggestion
    /**
     * Determines which wallet options to display in Coinbase Wallet SDK.
     * @property preference
     *   - `all`: Show both smart wallet and EOA options.
     *   - `smartWalletOnly`: Show only smart wallet options.
     *   - `eoaOnly`: Show only EOA options.
     * @see https://www.smartwallet.dev/sdk/v3-to-v4-changes#parameters
     */
    preference: "all";

```

enesozturk (approved):
Left some comments

packages/scaffold-utils/src/EthersTypesUtil.ts
```diff
@@ -28,7 +29,7 @@ export interface RequestArguments {
 
 export interface Provider {
   request: <T>(args: RequestArguments) => Promise<T>
-  on: <T>(event: string, listener: (data: T) => void) => void
+  on: <T>(event: any, listener: (data: T) => void) => void
```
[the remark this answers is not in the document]
tomiir: little hack I need to fix somehow, since coinbase events are hard-typed then this expects actual rpc method names instead of `string`.

will def fix before merging 🖖

examples/vue-wagmi/package.json
tomiir: sorted most breaking changes 
https://www.smartwallet.dev/sdk/v3-to-v4-changes

PR #2298, opened 2024-05-22
fixed: solflare network mismatch

BeycanDeveloper:
# Breaking Changes

The necessary parameter was sent to the Solflare adapter that needs the network value with a mapping suitable for the network structure in Web3Modal and the network structure in Solana Wallet Adapter packages. Thus, the network mismatch problem was solved.

# Changes

- feat: Solana packages
- fix: Solflare network mismatch
- chore: No

# Associated Issues

closes #...

### DIFF
--- a/packages/solana/src/client.ts
+++ b/packages/solana/src/client.ts
@@ -196,7 +196,7 @@ export class Web3Modal extends Web3ModalScaffold {
     }
     this.syncNetwork(chainImages)
 
-    this.walletAdapters = createWalletAdapters()
+    this.walletAdapters = createWalletAdapters(options.defaultChain)
     this.WalletConnectConnector = new WalletConnectConnector({
       relayerRegion: 'wss://relay.walletconnect.com',
       metadata,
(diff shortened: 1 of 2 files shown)


glitch-txs:
thanks! how can I reproduce this issue? could you share steps to repro? Thanks!

BeycanDeveloper:
Choose Devnet in Solflare and try sign message.

https://web3modal-laboratory-v3.vercel.app/library/solana/

image

BeycanDeveloper:
@glitch-txs I sent a message to you on Twitter. Can you answer? Also, Web3Modal packages cannot load in Windows. I tried in WSL still have some problems. Why?

glitch-txs:
@BeycanDeveloper you don't need WSL, will check on twitter 👍

phoenixVS:
Thank you, good point! I've reproduced it, but if we rely on default chain it is erroneous for test and dev net
So I propose this approach, also I added listener for chain switching
https://github.com/WalletConnect/web3modal/pull/2304

BeycanDeveloper:
@phoenixVS Yes, you are right, this is a more correct approach :D. I forgot about the situation when the network changed. It must be because my own project is running over the initial network.

tomiir:
@phoenixVS was this fixed by https://github.com/WalletConnect/web3modal/pull/2304 ?
if so, can we close this pr?

PR #2300, opened 2024-05-22
chore: enable email and onramp by default

tomiir:
# Breaking Changes

N/A

# Changes

- feat:
- fix:
- chore: enable onramp and email login by default


glitch-txs:
why this breaking change?

enesozturk (approved):

glitch-txs:
Should we default socials here as well?

PR #2305, opened 2024-05-23
fix: makes `getMessageParams` siwe client method optional

ganchoradkov:
# Breaking Changes

N/A

# Changes

- feat:
- fix:
- chore:

# Associated Issues

closes #...


packages/ethers/src/client.ts
```diff
@@ -190,12 +190,14 @@ export class Web3Modal extends Web3ModalScaffold {
           onUri(uri)
         })
 
-        if (siweConfig?.options?.enabled) {
+        const params = await siweConfig?.getMessageParams?.()
+        // Must perform these checks to satify optional types
+        if (siweConfig?.options?.enabled && params && Object.keys(params || {}).length > 0) {
```
tomiir, line 195: are the params required for the `authenticate` call?

ganchoradkov: yes they are required. Idea is, if the message params are empty, it fallsback to legacy connect + sign

tomiir (approved):
LGTM

PR #2319, opened 2024-05-24
fix: update-email test

ignaciosantise:
# Changes
- fix: check if new email is visible at the end of the test

**Before:**
In the last step, the test checks if the account view is visible

**Now:**
-Check if the account view is visible again
-Check if the email in the account view is the new email

### DIFF
--- a/apps/laboratory/tests/shared/pages/ModalPage.ts
+++ b/apps/laboratory/tests/shared/pages/ModalPage.ts
@@ -263,9 +263,9 @@ export class ModalPage {
     await this.page.waitForTimeout(300)
   }
 
-  async updateEmail(mailsacApiKey: string, index: number) {
+  async updateEmail(mailsacApiKey: string) {
     const email = new Email(mailsacApiKey)
-    const newEmailAddress = email.getEmailAddressToUse(index)
+    const newEmailAddress = email.getEmailAddressToUse(1)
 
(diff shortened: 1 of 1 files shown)


apps/laboratory/tests/shared/pages/ModalPage.ts
```diff
@@ -263,9 +263,9 @@ export class ModalPage {
     await this.page.waitForTimeout(300)
   }
 
-  async updateEmail(mailsacApiKey: string, index: number) {
+  async updateEmail(mailsacApiKey: string) {
     const email = new Email(mailsacApiKey)
-    const newEmailAddress = email.getEmailAddressToUse(index)
+    const newEmailAddress = email.getEmailAddressToUse(1)
```
enesozturk, line 268: @ignaciosantise I'd really appreciate it if you could tell more about the situation - why previously it had issues, how this is improving etc - in the PR description. Since tests have gotten overwhelming last few months and are very fragile , we might miss some details 🙏🏽

ignaciosantise: Actually, this change is not needed. It was needed because Luka changed the `getEmailAddressToUse` function, but now apparently it changed again 😆 I'll revert this

enesozturk (approved):

PR #2332, opened 2024-05-27
:feat social provider storage logic 

svenvoskamp:
# Changes

* Add storage logic for `@w3m/connected_social 
* Add social icon when logged in with social provider + disable change email flow 
* Remove separator when `enableWalletConnect` is false


packages/scaffold/src/partials/w3m-account-default-widget/index.ts
```diff
@@ -221,20 +221,27 @@ export class W3mAccountDefaultWidget extends LitElement {
...
+        iconSize=${socialProvider ? 'xl' : 'sm'}
         data-testid="w3m-account-email-update"
-        ?chevron=${true}
-        @click=${() => this.onGoToUpdateEmail(email)}
+        ?chevron=${!socialProvider}
+        @click=${() => {
+          if (!socialProvider) {
+            this.onGoToUpdateEmail(email)
+          }
+        }}
```
tomiir, line 244: can we add the condition inside  `onGoToUpdateEmail`  instead?

packages/scaffold/src/views/w3m-connecting-social-view/index.ts
```diff
@@ -107,10 +118,17 @@ export class W3mConnectingSocialView extends LitElement {
...
             this.connecting = true
             const uri = event.data.resultUri as string
 
             await this.authConnector.provider.connectSocial(uri)
+
+            if (this.socialProvider) {
+              StorageUtil.setConnectedSocialProvider(this.socialProvider)
+            }
             await ConnectionController.connectExternal(this.authConnector)
           }
```
tomiir, line 133: do we need to clear the AccountController's `socialWindow` field here?

packages/scaffold/src/partials/w3m-account-auth-button/index.ts
```diff
@@ -0,0 +1,56 @@
...
+
+@customElement('w3m-account-auth-button')
+export class W3mAccountAuthButton extends LitElement {
+  // -- Render -------------------------------------------- //
+  public override render() {
+    const type = StorageUtil.getConnectedConnector()
+    const authConnector = ConnectorController.getAuthConnector()
+
+    if (!authConnector || type !== 'AUTH') {
+      return null
```
enesozturk, line 19: Do we need to set the `display: none`? When the `w3m-account-auth-button` rendered in a parent and if there is a `gap` between elements. Not setting `display: none` causing weird empty spaces between elements

tomiir (approved):

packages/scaffold/src/views/w3m-connecting-social-view/index.ts
```diff
@@ -24,10 +26,14 @@ export class W3mConnectingSocialView extends LitElement {
   // -- State & Properties -------------------------------- //
   @state() private socialProvider = AccountController.state.socialProvider
 
+  @state() private socialWindow = AccountController.state.socialWindow
+
   @state() protected error = false
```
enesozturk, line 31: Where these `error` and `connecting` are got updated? Cannot see they are updated with a listener or something. 

I'm concerned if we should call `this.getMessage()` in every render by calling it on `render()` funciton. 

If we are listening those values, we'd better to update this.message in the event listener or in life cycle methods IMO.

packages/wallet/src/W3mFrameStorage.ts
```diff
@@ -16,9 +16,13 @@ export const W3mFrameStorage = {
     return null
   },
 
-  delete(key: string) {
+  delete(key: string, social?: boolean) {
     if (W3mFrameHelpers.isClient) {
-      localStorage.removeItem(`${W3mFrameConstants.STORAGE_KEY}${key}`)
+      if (social) {
+        localStorage.removeItem(key)
```
enesozturk, line 22: Confused about this one, why there is no prefix like `W3mFrameConstants.STORAGE_KEY` for `social`?

enesozturk (approved):
LGTM ✅ @svenvoskamp left some comments

packages/wallet/src/W3mFrameStorage.ts
svenvoskamp: Because we need to add that item outside of the wallet package. But we need to remove it when we clear the cache.

tomiir: Is this not set by us and thus it lacks the prefix?

svenvoskamp: We set it oursvelves, but outside of the wallet package

PR #2341, opened 2024-05-29
refactor/siwe enable signout options

enesozturk:
- **chore: save state**
- **feat: handle signout on network change option for siwe**

# Breaking Changes

N/A

# Changes

- feat:
- fix:
- chore:

# Associated Issues

closes #...


packages/scaffold/src/modal/w3m-modal/index.ts
```diff
@@ -167,19 +168,38 @@ export class W3mModal extends LitElement {
...
+      if (session && previousNetworkId && newNetworkId && previousNetworkId !== newNetworkId) {
+        if (SIWEController.state._client?.options.signOutOnNetworkChange) {
+          await SIWEController.signOut()
+          this.onSiweNavigation()
+        }
+
+        return
       }
+
+      this.onSiweNavigation()
```
enesozturk, line 202: If not one of them, it's initialized from undefined to valid address so redirect to SIWE

svenvoskamp (approved):

apps/laboratory/src/utils/SiweUtils.ts
```diff
@@ -4,6 +4,8 @@ import { createSIWEConfig, formatMessage } from '@web3modal/siwe'
 import { WagmiConstantsUtil } from '../utils/WagmiConstants'
 
 export const siweConfig = createSIWEConfig({
+  signOutOnAccountChange: true,
```
enesozturk, line 7: These are already `true` by default but added just in case to not miss these props in case we change them in the future.

We'd better to add e2e tests that disable these. So test runner should run the same tests without expecting SIWE sign when change network. I'm not sure how we could achieve this since these are the lab's settings.

PR #2352, opened 2024-06-03
add universal link flag

glitch-txs:
**TODO**
- [x] Filter wallets if they don't support Universal Link
- [ ] Fix iOS issue

## How to test the preview in Telegram:

1. Look for the bot called @appkit_labs_bot in Telegram
2. Send any message or press start.
3. Labs will be open, choose the "wagmi default" option
4. Press the connect button, it should redirect to the wallet (any wallet) and be able to connect

# Changes

- feat: Add `enableUniversalLinks` flag to use Universal Linking on mobile, this will improve [Telegram Mini Apps](https://docs.ton.org/develop/dapps/telegram-apps/) support

Blocked by https://github.com/WalletConnect/cloud-app/pull/1195

# Associated Issues

closes #...


packages/scaffold/src/partials/w3m-connecting-wc-mobile/index.ts
```diff
@@ -36,8 +41,15 @@ export class W3mConnectingWcMobile extends W3mConnectingWidget {
     if (this.wallet?.mobile_link && this.uri) {
       try {
         this.error = false
-        const { mobile_link, name } = this.wallet
-        const { redirect, href } = CoreHelperUtil.formatNativeUrl(mobile_link, this.uri)
+        const { mobile_link: deep_link, link_mode: universal_link, name } = this.wallet
+        let link = deep_link
+        if (OptionsController.state.enableUniversalLinks) {
+          if (!universal_link) {
+            throw new Error('Universal Link is undefined.')
```
ignaciosantise, line 48: should we throw an error here? or just use the deeplink as fallback

glitch-txs: The flag is meant to use only Universal Link, in this case because Telegram doesn't support Deep Linking, so it would be the same for this particular case. Generally it is prefered Deeplinking and if it's undefined it will fallback to Universal Link in the API.

I added this error message because it's possible that a wallet supports DeppLinking but not Universal Link in which case with an error message it's easier to debug and we'll know what the issue is right away in case there are any reports.

So for general use cases this flag shouldn't be used only edge cases like the Telegram Mini Apps one.

ignaciosantise: mm okey. This solution is kind of a patch, the really good solution would be that if `enableUniversalLinks`, we should only show wallets compatible with universal links

enesozturk (approved):

packages/scaffold/src/partials/w3m-connecting-wc-mobile/index.ts
glitch-txs: Yeah sounds good! As for now we only need this as a Canary so I guess we could add this

glitch-txs: should be filtered now

packages/core/src/utils/TypeUtil.ts
```diff
@@ -103,6 +103,7 @@ export interface WcWallet {
   image_url?: string
   order?: number
   mobile_link?: string | null
+  link_mode?: string | null
```
ignaciosantise, line 106: this might bring confusion as we are developing another feature called link-mode 👀

glitch-txs: this comes from Cloud https://github.com/WalletConnect/cloud-app/pull/1195

apps/laboratory/src/pages/_app.tsx
```diff
@@ -7,12 +8,22 @@ import type { Session } from 'next-auth'
...
+
 export default function App({
   Component,
   pageProps
 }: AppProps<{
   session: Session
 }>) {
+  useEffect(() => {
+    window.Telegram?.WebApp.ready()
+  }, [])
```
tomiir, line 25: is this reqquired? what happens if it fails?

packages/scaffold-ui/src/partials/w3m-connecting-wc-mobile/index.ts
```diff
@@ -36,11 +41,20 @@ export class W3mConnectingWcMobile extends W3mConnectingWidget {
...
         this.error = false
-        const { mobile_link, name } = this.wallet
-        const { redirect, href } = CoreHelperUtil.formatNativeUrl(mobile_link, this.uri)
+        const { mobile_link: deep_link, link_mode: universal_link, name } = this.wallet
+        let link = deep_link
+        let target: '_self' | '_blank' = '_self'
+        if (OptionsController.state.enableUniversalLinks) {
+          if (!universal_link) {
+            throw new Error('Universal Link is undefined.')
+          }
```
tomiir, line 50: emabling universal links makes  it so it's required to have one? Or should it use one if it has it and fall back to previous behavior if not?

apps/laboratory/src/pages/_app.tsx
glitch-txs: This is used by Telegram to initialize the website inside a Mini App. Outside telegram it doesn't have any effect. We can remove this from labs if needed, it's just here for testing purposes.

packages/scaffold-ui/src/partials/w3m-connecting-wc-mobile/index.ts
tomiir: Aren't you calling `filterWAlletsWithUniversalLinks` ? You could make it so it's types as returning string[] and that would sort it out I think 🤔

glitch-txs: that function is only call if the flag is set to true, so it's not possible to determinate the right type there

tomiir (approved):

tomiir:
@glitch-txs should we move to draft until the issue with iOS is resolved ?

PR #2353, opened 2024-06-03
:fix social bug bash

svenvoskamp:
# Changes

* fix: discord returns a `0` on the username, do a check if this exists and remove it if thats the case
* fix: ellipse string when email is longer than 30 chars


packages/scaffold/src/partials/w3m-account-auth-button/index.ts
```diff
@@ -49,6 +50,18 @@ export class W3mAccountAuthButton extends LitElement {
...
+  private getAuthName(email: string) {
+    if (this.socialUsername) {
+      if (this.socialProvider === 'discord' && this.socialUsername.endsWith('0')) {
+        return this.socialUsername.slice(0, -1)
+      }
+
+      return this.socialUsername
+    }
+
+    return email.length > 30 ? `${email.slice(0, -3)}...` : email
```
tomiir, line 63: Wouldnt this leave the full email without the .com?

tomiir (approved):

packages/scaffold/src/partials/w3m-account-auth-button/index.ts
svenvoskamp: If the email is really long it will.. we can also split it in the middle? Don't really know which one I prefer

PR #2357, opened 2024-06-04
:feat add `email` to `auth` parameter & rename `emailConnector` for `authConnector`

svenvoskamp:
# Breaking changes

* rename `emailConnector` for `authConnector` 
* add `email` to `auth` parameter 
* `showWallets` and `email` defaulted to be true

# Changes

* Set `socialProvider` first before making fetch call, so users always see which social provider they are connected to. 
* Small styling fix when 2 auth providers are enabled


packages/wagmi/src/connectors/AuthConnectorExport.ts
```diff
@@ -0,0 +1,23 @@
...
+  showWallets?: boolean
+  email?: boolean
+}
+
+export function authConnector(parameters: AuthParameters) {
+  return authConnectorWagmi({
+    ...parameters,
+    email: true,
+    showWallets: true
+  })
```
enesozturk, line 22:
```suggestion
  return authConnectorWagmi({
    email: true,
    showWallets: true,
    ...parameters,
  })
```

enesozturk (approved):
LGTM 🚀

PR #2358, opened 2024-06-04
refactor: swap ux and business logic improvements

enesozturk:
# Changes

- Refactor swap loading logics to provide better UX to users:
   - After approval TX or swap TX, show


packages/ui/src/assets/svg/image.ts
svenvoskamp, line 1: I think this needs formatting?

svenvoskamp (approved):
LGTM ✅

PR #2360, opened 2024-06-04
chore: use walletconnect.org

tomiir:
# Breaking Changes

N/A

# Changes

- feat:
- fix:
- chore: remove `walletconnect.com` from our rpc urls and replace with `walletconnect.org`


packages/core/src/utils/CoreHelperUtil.ts
```diff
@@ -174,33 +175,16 @@ export const CoreHelperUtil = {
...
-    } catch {
-      return false
-    }
-  },
-
   getApiUrl() {
-    return CoreHelperUtil.isRestrictedRegion()
-      ? 'https://api.web3modal.org'
-      : 'https://api.web3modal.com'
+    return 'https://api.web3modal.org'
```
chris13524, line 179: Should this be a constant too?

chris13524 (approved):

packages/core/src/utils/CoreHelperUtil.ts
enesozturk: +

packages/core/src/utils/CoreHelperUtil.ts
```diff
@@ -174,33 +175,16 @@ export const CoreHelperUtil = {
...
-      ? 'https://rpc.walletconnect.org'
-      : 'https://rpc.walletconnect.com'
+    return CommonConstants.BLOCKCHAIN_API_RPC_URL
   },
 
   getAnalyticsUrl() {
-    return CoreHelperUtil.isRestrictedRegion()
-      ? 'https://pulse.walletconnect.org'
-      : 'https://pulse.walletconnect.com'
+    return 'https://pulse.walletconnect.org'
```
enesozturk, line 187: We could make this constant too IMO

packages/solana/src/utils/scaffold/SolanaHelpersUtils.ts
```diff
@@ -7,7 +8,7 @@ import type { Chain, Provider } from './SolanaTypesUtil.js'
 
 export const SolHelpersUtil = {
   detectRpcUrl(chain: Chain, projectId: string) {
-    if (chain.rpcUrl.includes('rpc.walletconnect.com/')) {
+    if (chain.rpcUrl.includes(ConstantsUtil.BLOCKCHAIN_API_RPC_URL)) {
```
enesozturk, line 11: @tomiir The previous version doesn't include the protocol in the condition;

```js
    if (chain.rpcUrl.includes('rpc.walletconnect.com/')) {
```

While this one have full url as param of `includes`. Maybe this might affect the result? Worth double checking

ignaciosantise:
We should also remove the fallback logic in `@walletconnect/core`

packages/solana/src/client.ts
```diff
@@ -208,7 +208,7 @@ export class Web3Modal extends Web3ModalScaffold {
 
     this.walletAdapters = createWalletAdapters(chain?.chainId)
     this.WalletConnectConnector = new WalletConnectConnector({
-      relayerRegion: 'wss://relay.walletconnect.com',
+      relayerRegion: 'wss://relay.walletconnect.org',
```
tomiir, line 211: @chris13524 is this correct as well? Found it while triple checking the code.

PR #2364, opened 2024-06-05
refactor: disable disconnected call for siwe if not connected

enesozturk:
# Changes

When we not connected, we are still calling the client's disconnect function and if the SIWE enabled, this causing unnecessary async function and this causing the modal to be closed with a promise delay.

## Also considered

We could do optimistic updates and directly close the modal and make the `signOut` calls on the background. While this might sound good, might cause some issues for certain situations


chris13524 (commented):
Test case?

tomiir (approved):
Looks good but a test would be nice 🙏

svenvoskamp (approved):

enesozturk:
> Looks good but a test would be nice 🙏

This is something that could be tested with integration tests, not E2E because it's a decision when to trigger SIWE actions rather than any affect to how they are working. And this change is just effecting the modal when it's in `disconnected` state.

If we decide to test these kind of details, it'll bring too much workload to our E2E tests which their right place is integration tests. Maybe we could consider having them with Jest rather than Playwright

PR #2366, opened 2024-06-05
chore: make wallets unique by id before updating state

tomiir:
# Breaking Changes

N/A

# Changes

- feat: add `uniqueBy` core helper util
- fix:
- chore: make wallets unique by id

### DIFF
--- a/packages/core/src/utils/CoreHelperUtil.ts
+++ b/packages/core/src/utils/CoreHelperUtil.ts
@@ -283,5 +283,9 @@ export const CoreHelperUtil = {
     }
 
     return false
+  },
+
+  uniqueBy<T>(arr: T[], key: keyof T) {
+    return Array.from(new Map(arr.map(i => [i[key], i])).values())
   }
 }
--- a/packages/core/src/controllers/ApiController.ts
+++ b/packages/core/src/controllers/ApiController.ts
@@ -188,7 +188,7 @@ export const ApiController = {
       ...(images as string[]).map(id => ApiController._fetchWalletImage(id)),
       CoreHelperUtil.wait(300)
     ])
-    state.wallets = [...state.wallets, ...data]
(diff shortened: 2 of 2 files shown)


enesozturk (approved):

packages/core/src/utils/CoreHelperUtil.ts
```diff
@@ -267,5 +267,9 @@ export const CoreHelperUtil = {
     }
 
     return false
+  },
+
+  uniqueBy<T>(arr: T[], key: keyof T) {
+    return Array.from(new Map(arr.map(i => [i[key], i])).values())
```
enesozturk, line 273: @tomiir 

Seems we have many iterations here so I wanted to check O(n) complexity. 

While the current solution have the O(n) complexity, I do think we could make it more performant:
- `arr.map(I => [I[key], I])`: O(n)
- `new Map(arr.map(...))`: O(n)
- `Array.from(map.values())`: O(n)

O(n) + O(n) + O(n) = O(3n), which is equals to O(n).

If we go with the following approach below, it's still O(n) but since the only iteration is `.filter`, the general performance is much faster. 

```ts
function uniqueBySet<T>(arr: T[], key: keyof T): T[] {
  const seen = new Set();
  return arr.filter(item => {
    const keyValue = item[key];
    if (seen.has(keyValue)) {
      return false;
    }
    seen.add(keyValue);
    return true;
  });
}
```

Compared both functions with `10_000` items here is the results:

- uniqueByMap took `2.899999976158142` ms
- uniqueBySet took `0.699999988079071` ms

with `100_000` items:
- uniqueByMap took `13.800000011920929` ms
- uniqueBySet took `2.2999999821186066` ms

tomiir: updated

PR #2367, opened 2024-06-05
fix: use defaultRPC Url if present in wagmi

tomiir:
# Breaking Changes

N/A

# Changes

- feat:
- fix: use default transport url if present
- chore:

### DIFF
--- a/packages/wagmi/src/utils/helpers.ts
+++ b/packages/wagmi/src/utils/helpers.ts
@@ -45,12 +45,16 @@ export function getEmailCaipNetworks() {
   }
 }
 
-export function getTransport({ chainId, projectId }: { chainId: number; projectId: string }) {
-  const RPC_URL = CoreHelperUtil.getBlockchainApiUrl()
+export function getTransport({ chain, projectId }: { chain: Chain; projectId: string }) {
+  const defaultRpcUrl = chain.rpcUrls.default?.http[0]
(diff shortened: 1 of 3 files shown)


packages/wagmi/src/utils/helpers.ts
```diff
@@ -45,12 +45,16 @@ export function getEmailCaipNetworks() {
...
+  }
 
-  if (!PresetsUtil.WalletConnectRpcChainIds.includes(chainId)) {
+  const RPC_URL = CoreHelperUtil.getBlockchainApiUrl()
+  if (!PresetsUtil.WalletConnectRpcChainIds.includes(chain.id)) {
     return http()
   }
 
-  return http(`${RPC_URL}/v1/?chainId=${ConstantsUtil.EIP155}:${chainId}&projectId=${projectId}`)
+  return http(`${RPC_URL}/v1/?chainId=${ConstantsUtil.EIP155}:${chain.id}&projectId=${projectId}`)
```
enesozturk, line 59: Even though Typescript help us to make sure we are passing `Chain` typed data and it has `id` prop, using property access in template literals increases my anxiety, lol. 

Not sure if we should check `chain.id` before using it, but just wanted to share to discuss maybe. I think we are still okay to go. (I hope)

enesozturk (approved):

PR #2377, opened 2024-06-07
chore: add assetlinks for kotlin sample dapp and wallet

lukaisailovic:
# Changes
- chore: added assetlinks file to laboratory with Kotlin & RN sample apps


tomiir (approved):
Just a double check that the `.debug`  packages are ok too?

PR #2390, opened 2024-06-11
Fix/add featured widget

svenvoskamp:
# Changes

- fix: add featured widget

closes https://github.com/WalletConnect/web3modal/issues/2383


packages/scaffold/src/views/w3m-connect-view/index.ts
```diff
@@ -49,8 +49,9 @@ export class W3mConnectView extends LitElement {
             <w3m-connect-recent-widget></w3m-connect-recent-widget>
             <w3m-connect-announced-widget></w3m-connect-announced-widget>
             <w3m-connect-injected-widget></w3m-connect-injected-widget>
-            <w3m-connect-coinbase-widget></w3m-connect-coinbase-widget>
+            <w3m-connect-featured-widget></w3m-connect-featured-widget>
```
enesozturk, line 52: @svenvoskamp What about the `w3m-connect-coinbase-widget` one? Does the featured widgets including coinbase as well? If yes, should we remove `w3m-connect-coinbase-widget`?

svenvoskamp: @tomiir @enesozturk I think coinbase is part of `w3m-connect-external-widget`? Please correct me if I'm wrong. 

I think we can remove the `w3m-connect-coinbase-widget` and replace it for the external one

tomiir: I think the only external one we have is coinbase. We can remove either of them I think, but CB widget has some logic to hide the duplicated connectors

svenvoskamp: Okay, then I will remove the external one

enesozturk (approved):

PR #2391, opened 2024-06-11
fix: telegram icon

tomiir:
# Breaking Changes

N/A

# Changes

- feat:
- fix: telegram icon showing discord
- chore:


enesozturk (approved):

svenvoskamp (approved):
Do we need this icon somewhere?  I initially added this as I thought it would be part of social login, but don't we don't use it, so we can maybe remove it?

PR #2392, opened 2024-06-11
refactor: coinbase connector popup improvements

enesozturk:
# Changes

### `onConnect` trigger logic

When users select a connector, we are redirecting them to `w3m-connecting-external-view`  and trigger the `onConnect` method in the first render and watch changes. While this is working fine 99%, there are some issues with the ones which opens a popup.

Coinbase's new SDK using popups and modern browsers disable popups if they are triggered programmatically rather than user action. If a user not trigger the popup, they are being disabled. This results to not have popup opened on some browsers/devices. I.e Safari/iOS. 

iOS Safari app have `Block pop-ups` setting enabled by default under `Settings > Safari > General`. Most users don't aware of this and the ones who select Coinbase won't be realizing there should be popup opened.

### Fix

To fix this, we need to trigger the connection once the users select the Coinbase app from the wallet list, and at the same time, redirect them to loading page (`w3m-connecting-external-view`) and watch changes.

This PR introduces changes to connector methods to handle this specific case for Coinbase.


packages/scaffold/src/partials/w3m-connect-external-widget/index.ts
```diff
@@ -53,7 +63,45 @@ export class W3mConnectExternalWidget extends LitElement {
   }
 
   // -- Private Methods ----------------------------------- //
+  /**
+   * Replicate of `onConnectProxy` method from `w3m-connecting-external-view` which is only used for Coinbase connector.
+   */
+  private async onCoinbaseConnector(connector: Connector) {
+    try {
+      ConnectionController.setWcError(true)
```
tomiir, line 71: shouldn't this be false?

tomiir (approved):

packages/scaffold/src/partials/w3m-connect-external-widget/index.ts
svenvoskamp, line 1: Shouldn't this be part of the `w3m-connect-coinbase-widget`? In this PR https://github.com/WalletConnect/web3modal/pull/2390 we decided to remove external widget. Wdyt?

PR #2393, opened 2024-06-12
feat: 5.0 missing events 

tomiir:
# Breaking Changes

N/A

# Changes
Adds missing events for missing flows
- Send flow
- Swaps
- Names

Adds `isSmartAccount` flag to transaction history and SIWE events

- feat: add missing wallet features events
- fix: 
- chore:


packages/core/src/controllers/SendController.ts
```diff
@@ -103,6 +118,18 @@ export const SendController = {
       this.state.gasPrice &&
       this.state.token?.quantity.decimals
     ) {
+      EventsController.sendEvent({
```
enesozturk, line 121: Why there is two `SEND_INITIATED` events here?

svenvoskamp: One is for native token and the other one for `erc-20`

svenvoskamp (approved):

PR #2406, opened 2024-06-12
chore(renovate): only auto rebase renovate PRs when conflicted

bkrem:
## Context

- Renovatebot will by default auto rebase all open renovate PRs when the base branch is updated.
- This is causing Vercel build cascades that cause long queue times unnecessarily for more important builds.
- We already use this rule here: https://github.com/WalletConnect/web-examples/blob/main/renovate.json#L10

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -7,6 +7,7 @@
     ":semanticCommits",
     ":semanticCommitScope(deps)"
   ],
+  "rebaseWhen": "conflicted",
   "packageRules": [
     {
       "matchPackagePatterns": ["*"],


chris13524 (commented):
I don't think this works because PRs must incorporate all changes before they can be merged, not just when conflicted

bkrem:
> I don't think this works because PRs must incorporate all changes before they can be merged, not just when conflicted

Point is that you manually trigger a rebase via the checkbox in the renovate PR when necessary:
Screenshot 2024-06-12 at 16 34 32

This isn't a must have but it was pretty painful just now when we were trying to deploy the website changes and renovatebot kept clogging up the vercel queue with rebases of open PRs

chris13524 (approved):

tomiir (approved):

PR #2408, opened 2024-06-12
chore: change configs to use main and V5

tomiir:
# Breaking Changes

N/A

# Changes

- feat:
- fix:
- chore: make CI trigger on V5 and main


.changeset/config.json
```diff
@@ -5,7 +5,7 @@
   "fixed": [],
   "linked": [],
   "access": "public",
-  "baseBranch": "V4",
+  "baseBranch": "V5",
```
chris13524, line 8: What does this do?

tomiir: https://github.com/changesets/changesets/blob/main/docs/config-file-options.md#basebranch-git-branch-name

Will move to main

tomiir: In fact, this should only be used on releases so it's correct to point to V5

PR #2412, opened 2024-06-13
Fix/social safari

svenvoskamp:
# Changes

- fix: to make sure popup are not blocked on various browsers, directly open the popup and set the location href after the fetch call for all browsers 

Fixes https://linear.app/walletconnect/issue/APKT-462/bug-cant-login-using-arc-browser-safari


packages/scaffold/src/partials/w3m-social-login-widget/index.ts
```diff
@@ -162,32 +160,21 @@ export class W3mSocialLoginWidget extends LitElement {
...
-    }
+    this.popupWindow = CoreHelperUtil.returnOpenHref(
+      '',
+      'popupWindow',
+      'width=600,height=800,scrollbars=yes'
+    )
 
     try {
       if (authConnector && socialProvider) {
         const { uri } = await authConnector.provider.getSocialRedirectUri({
```
enesozturk, line 171: What if there is an error in this request? Would the popup be still open? If yes I think we should close it

enesozturk:
@svenvoskamp Another issue I'm not sure if it could be done easily; after opening the social popup and closing it, it still stay at loading screen and doesn't show "try again"/"declined" screens. Would it be feasible to achieve?

enesozturk (approved):

svenvoskamp:
> @svenvoskamp Another issue I'm not sure if it could be done easily; after opening the social popup and closing it, it still stay at loading screen and doesn't show "try again"/"declined" screens. Would it be feasible to achieve?

I've added an interval that checks if the popup has been closed, then check if connecting is false & route is on `connectingSocial`. If this is the case, we can perfom a  `goBack()` on the routerController

packages/scaffold/src/views/w3m-connecting-social-view/index.ts
```diff
@@ -163,6 +163,14 @@ export class W3mConnectingSocialView extends LitElement {
   }
 
   private connectSocial() {
+    const interval = setInterval(() => {
+      if (this.socialWindow?.closed) {
+        if (!this.connecting && RouterController.state.view === 'ConnectingSocial') {
+          RouterController.goBack()
+        }
+        clearInterval(interval)
```
enesozturk, line 171: I'm confused how this is working in terms of closures. The callback function is not binded to the parent class and I wouldn't expect the `interval` value is available inside the `setInterval`'s callback function. Weird, let's double check this in case it'd be problematic in different environments. Smells to me but cannot be sure tbh

PR #2414, opened 2024-06-13
:feat exclude wallet RDNS

svenvoskamp:
# Changes

feat: Added a new `excludeWalletRDNS` param. This array contains RDNS entries for wallet domain names. Previously, developers could not exclude wallets injected from EIP6963. Now, they can add the RDNS names of wallets they wish to filter out.

fixes https://linear.app/walletconnect/issue/APKT-473/customers-cannot-override-6963-to-exclude-non-compatible-wallets


tomiir (approved):

Cali93 (commented):
Can we add tests to cover this ? 🙏

svenvoskamp:
Closing as this is not the ideal solution

PR #2417, opened 2024-06-13
fix: close modal if no need to reauth

tomiir:
# Summary
Fix issue where an integration with `signOutOnNetworkChange: false` would get stuck on the loading modal after switching in the wallet. 

# Changes

- feat:
- fix: close modal if there's no need to re-auth SIWE
- chore:

# Associated Issues

closes #APKT-596

### DIFF
--- a/apps/laboratory/tests/siwe.spec.ts
+++ b/apps/laboratory/tests/siwe.spec.ts
@@ -10,28 +10,28 @@ testMWSiwe.afterEach(async ({ modalValidator, walletValidator, browserName }) =>
   await walletValidator.expectDisconnected()
 })
 
-// testMWSiwe(
-//   'it should sign in with ethereum',
-//   async ({ modalPage, walletPage, modalValidator, walletValidator }) => {
-//     const uri = await modalPage.getConnectUri()
-//     await walletPage.connectWithUri(uri)
-//     await walletPage.handleSessionProposal(DEFAULT_SESSION_PARAMS)
-//     await modalValidator.expectAuthenticated()
-//     await modalValidator.expectConnected()
-//     await walletValidator.expectConnected()
-//     await modalPage.disconnect()
-//   }
-// )
+testMWSiwe(
+  'it should sign in with ethereum',
+  async ({ modalPage, walletPage, modalValidator, walletValidator }) => {
+    const uri = await modalPage.getConnectUri()
+    await walletPage.connectWithUri(uri)
+    await walletPage.handleSessionProposal(DEFAULT_SESSION_PARAMS)
+    await modalValidator.expectAuthenticated()
+    await modalValidator.expectConnected()
+    await walletValidator.expectConnected()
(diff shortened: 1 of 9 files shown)


packages/scaffold/src/views/w3m-network-switch-view/index.ts
```diff
@@ -41,6 +39,7 @@ export class W3mNetworkSwitchView extends LitElement {
 
   // -- Render -------------------------------------------- //
   public override render() {
+    console.log('SWITCH NETWORK VIEW')
```
enesozturk, line 42: Log

packages/scaffold/src/modal/w3m-modal/index.ts
```diff
@@ -189,16 +187,6 @@ export class W3mModal extends LitElement {
         return
       }
 
-      // If the network has changed and signOnNetworkChange is enabled, sign out
```
enesozturk, line 192: What if the user change the network from the wallet when the w3m is closed state? I guess this is the only place that will trigger SIWE 🤔

apps/laboratory/tests/siwe.spec.ts
```diff
@@ -32,3 +32,21 @@ testMWSiwe(
     await modalValidator.expectUnauthenticated()
   }
 )
+
+testMWSiwe(
```
Cali93, line 36: Awesome to see a test covering this 💪

packages/scaffold/src/views/w3m-network-switch-view/index.ts
```diff
@@ -134,7 +134,12 @@ export class W3mNetworkSwitchView extends LitElement {
       this.error = false
       if (this.network) {
         await NetworkController.switchActiveNetwork(this.network)
-        if (!OptionsController.state.isSiweEnabled) {
+        if (OptionsController.state.isSiweEnabled) {
+          const { SIWEController } = await import('@web3modal/siwe')
+          if (SIWEController.state._client?.options?.signOutOnNetworkChange) {
+            await SIWEController.signOut()
+          }
+        } else {
```
Cali93, line 142: Should this logic live in `switchActiveNetwork` to avoid code duplication ?

Cali93 (approved):
LGTM, just a minor suggestion to maybe abstract this logic to avoid duplication.

packages/scaffold/src/views/w3m-network-switch-view/index.ts
tomiir: That causes a circular dependency cause it'd require `siwe` in `core` which is imported in `siwe` 😵 
But agree not cool to duplicate code

tomiir: abstracted to a utility 🙏

enesozturk (approved):

svenvoskamp (approved):

PR #2418, opened 2024-06-14
:feat exclude eip6963 wallets 

svenvoskamp:
# Changes

- get rdns from excluded wallets and check if they match in the specific UI partials 
- Add unit test


packages/core/src/controllers/ApiController.ts
```diff
@@ -193,6 +195,27 @@ export const ApiController = {
...
+        entries: String(ids.length),
+        chains: NetworkController.state.caipNetwork?.id,
+        include: ids?.join(',')
+      }
+    })
+
+    if (data) {
+      data.forEach(wallet => {
+        if (wallet?.rdns) {
+          state.excludedRDNS.push(wallet.rdns)
```
tomiir, line 213: is this ok?

PR #2423, opened 2024-06-14
fix: reorder chains to have current chain Id as main message

tomiir:
# Breaking Changes

N/A

# Changes

- feat:
- fix: issue where connection would default to Ethereum when connecting to react-wallet
- chore:

# Notes
We should implement wallet_addEthereumChain on react-wallet 

# Associated Issues

closes https://linear.app/walletconnect/issue/W3W-113/when-doing-1-click-auth-siwe-on-base-i-am-switched-back-to-ethereum


ganchoradkov (approved):
💯

PR #2450, opened 2024-06-20
fix: gallery build issues

tomiir:
# Breaking Changes

N/A

# Changes

- feat:
- fix: gallery build issues caused by `common` as devDependency on `ui` and not properly set at turbo build level
- chore:

### DIFF
--- a/turbo.json
+++ b/turbo.json
@@ -23,7 +23,8 @@
       "cache": false
     },
     "build:core": {
-      "outputs": ["dist/**", "^build:common", "^build:wallet"],
+      "outputs": ["dist/**"],
+      "dependsOn": ["^build:common", "^build:wallet"],
       "cache": false
     },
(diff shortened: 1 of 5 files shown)


glitch-txs (approved):

turbo.json
```diff
@@ -13,16 +13,17 @@
...
     },
     "build:ui": {
       "outputs": ["dist/**"],
+      "dependsOn": ["^build:common"],
       "cache": false
     },
-    "build:wallet": {
-      "outputs": ["dist/**"],
+    "build:core": {
+      "outputs": ["dist/**", "^build:common", "^build:wallet"],
```
enesozturk, line 26: Do we sure that outputs should include build scripts?

tomiir: this is misplaced 🤔

PR #2458, opened 2024-06-23
add Solana chain ID workaround

glitch-txs:
# fix: 

- There's currently an issue on Solana integrations, we have an old and a new chain ID for Mainnet and Devnet. Some wallets and some dapps are using the old chain ID and vice versa.
This PR adds a patch to support both cases and throw a warning in the dapp is the wallet or the dapp are using the old chain.

Tested with WalletConnect Web Sample Wallet and Trust Wallet on Android. (Connection + Sign Message)

- When the session was deleted from the wallet side while the app was closed, on page reload it would show up as connected.


packages/solana/src/utils/chainPath/constants.ts
```diff
@@ -0,0 +1,6 @@
+export const SolanaChainIDs = {
+	Mainnet: 'solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp',
```
enesozturk, line 2: What about the testnet? Don't we have to include it as well? Or maybe not any changes to it? @glitch-txs

glitch-txs: testnet ID was added last so no need

packages/solana/src/utils/chainPath/index.ts
```diff
@@ -0,0 +1,55 @@
...
+
+export function getDefaultChainFromSession(
+  session: SessionTypes.Struct,
+  selectedChain: ChainIDType
+) {
+  const chains = session.namespaces['solana']?.accounts.map(
+    (account: string) => `solana:${account.split(':')[1]}`
+  )
+
+  if (selectedChain === Chains.Mainnet) {
```
enesozturk, line 39: @glitch-txs Shouldn't these be like the following:
```suggestion
  if (selectedChain === Chains.Mainnet_Deprecated) {
```

enesozturk: Or okay this logic working another way

enesozturk (approved):

PR #2463, opened 2024-06-25
add Universal Provider option

glitch-txs:
# Changes

This is a flag for internal use on our WalletAdapter client
The purpose is to filter out wallets that do not support WC protocol when using the universal provider

- feat: 
- [x] Add a "low" level flag for universal provider -> `isUniversalProvider`.
- [x]  Remove Browser option from wallets when `isUniversalProvider` is set to `true`.
- [x] Filter out injected-only wallets from "All Wallets" when `isUniversalProvider` is set to `true`.

closes https://github.com/WalletConnect/walletconnect-solana/issues/25


packages/core/src/controllers/ApiController.ts
```diff
@@ -65,6 +65,14 @@ export const ApiController = {
     }
   },
 
+  _filterOutExtensions(wallets: WcWallet[]) {
+    if (OptionsController.state.isUniversalProvider) {
+      return wallets.filter(w => Boolean(w.mobile_link))
```
enesozturk, line 70: Why we only enable mobile wallets if universal provider enabled?

glitch-txs: updated to Mobile + Desktop + Webapps

enesozturk (approved):

packages/scaffold-ui/src/views/w3m-connecting-wc-view/index.ts
```diff
@@ -133,7 +133,7 @@ export class W3mConnectingWcView extends LitElement {
     const { mobile_link, desktop_link, webapp_link, injected, rdns } = this.wallet
     const injectedIds = injected?.map(({ injected_id }) => injected_id).filter(Boolean) as string[]
     const browserIds = rdns ? [rdns] : injectedIds ?? []
-    const isBrowser = browserIds.length
+    const isBrowser = OptionsController.state.isUniversalProvider ? false : browserIds.length
```
tomiir, line 136:
```suggestion
    const isBrowser = !OptionsController.state.isUniversalProvider || browserIds.length
```

packages/scaffold-ui/src/views/w3m-connecting-wc-view/index.ts
```diff
@@ -133,7 +133,7 @@ export class W3mConnectingWcView extends LitElement {
     const { mobile_link, desktop_link, webapp_link, injected, rdns } = this.wallet
     const injectedIds = injected?.map(({ injected_id }) => injected_id).filter(Boolean) as string[]
     const browserIds = rdns ? [rdns] : injectedIds ?? []
-    const isBrowser = browserIds.length
+    const isBrowser = !OptionsController.state.isUniversalProvider || browserIds.length
```
glitch-txs, line 136:
```suggestion
    const isBrowser = OptionsController.state.isUniversalProvider ? false : browserIds.length
```

PR #2467, opened 2024-06-26
chore: Verify API tests

chris13524:
# Changes

- chore: Verify API tests

# Other

>Danger: Ignore apps/laboratory/src/pages/library/verify-valid.tsx uses localhost: which is likely a mistake


apps/laboratory/tests/shared/pages/ModalPage.ts
```diff
@@ -35,6 +49,32 @@ export class ModalPage {
...
+        response.headers.forEach((value, key) => {
+          headers[key] = value
+        })
+        const body = Buffer.from(await response.arrayBuffer())
+        await route.fulfill({
+          status: response.status,
+          headers,
+          body
+        })
+      })
```
tomiir, line 75: if not too much trouble I'd move this to some auxiliary method like `routeMaliciousRoute` 
Also, what is it exactly doing? Cause you are routing to the malicious url but replacing the url on the request and proxying the get request?

apps/laboratory/tests/shared/utils/project.ts
```diff
@@ -37,18 +37,18 @@ const braveOptions: UseOptions = {
...
+    testIgnore: /(?:social\.spec\.ts|verify\.spec\.ts).*$/u
   },
   'Desktop Brave/ethers': {
-    testIgnore: /(?:email\.spec\.ts|smart-account\.spec\.ts|social\.spec\.ts).*$/u,
+    testIgnore: /(?:email\.spec\.ts|smart-account\.spec\.ts|social\.spec\.ts|verify\.spec\.ts).*$/u,
     useOptions: braveOptions
   },
   'Desktop Firefox/ethers': {
-    testIgnore: /(?:social\.spec\.ts).*$/u
+    testIgnore: /(?:social\.spec\.ts|verify\.spec\.ts).*$/u
```
tomiir, line 51: why disable this?

apps/laboratory/tests/verify.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+testMVerifyEvil(
+  'connection and signature requests from scam verified domain should show as scam domain',
+  async ({ modalPage, modalValidator, context }) => {
+    const walletPage = new WalletPage(await context.newPage())
+    await walletPage.load()
+    const walletValidator = new WalletValidator(walletPage.page)
+
+    const uri = await modalPage.getConnectUri()
+    await walletPage.connectWithUri(uri)
+    await expect(walletPage.page.getByText('Website flagged')).toBeVisible()
```
tomiir, line 100: shouldn't this be on the validator?

apps/laboratory/tests/shared/utils/project.ts
chris13524: Not sure; I removed

apps/laboratory/tests/shared/pages/ModalPage.ts
chris13524: Refactored in latest version

apps/laboratory/tests/verify.spec.ts
chris13524: What do you mean?

lukaisailovic: assertions are generally on the validator (ModalValidator)

lukaisailovic:
Why did we remove `ModalValidator` from the Fixture and instead instantiate it multiple times?

chris13524:
> Why did we remove ModalValidator from the Fixture and instead instantiate it multiple times?

I did that in [this PR](https://github.com/WalletConnect/web3modal/pull/2590) as we were trying to improve test reliability and fixtures were a potential source of distrust.

apps/laboratory/src/pages/library/verify-valid.tsx
```diff
@@ -0,0 +1,56 @@
+import { createWeb3Modal } from '@web3modal/wagmi/react'
```
tomiir, line 1: How is this different from our regular pages?

apps/laboratory/tests/shared/fixtures/w3m-verify-domain-mismatch-fixture.ts
```diff
@@ -0,0 +1,12 @@
+import type { ModalFixture } from './w3m-fixture'
+import { ModalPage } from '../pages/ModalPage'
+import { timingFixture } from './timing-fixture'
+
+export const testMVerifyDomainMismatch = timingFixture.extend<ModalFixture>({
```
tomiir, line 5: Why use a fixture here?

apps/laboratory/tests/shared/pages/ModalPage.ts
```diff
@@ -8,13 +8,28 @@ import { DeviceRegistrationPage } from './DeviceRegistrationPage'
...
+  | 'verify-domain-mismatch'
+  | 'verify-evil'
+  | 'all'
 
 function getUrlByFlavor(baseUrl: string, library: string, flavor: ModalFlavor) {
   const urlsByFlavor: Partial<Record<ModalFlavor, string>> = {
     default: `${baseUrl}library/${library}/`,
-    external: `${baseUrl}library/external/`
+    external: `${baseUrl}library/external/`,
+    'verify-valid': `${baseUrl}library/verify-valid/`,
```
tomiir, line 30: Why the special cases?

apps/laboratory/src/pages/library/verify-valid.tsx
chris13524: The metadata URL is set to `http://localhost:3000` and it uses a special project ID with Verify API enabled on localhost

apps/laboratory/tests/shared/fixtures/w3m-verify-domain-mismatch-fixture.ts
chris13524: Because we are making `ModalPage` inside the fixture and I need to supply custom config to it, requiring a different fixture.

I could make a new `ModalPage` using the same fixture, but then wouldn't it be doing unnecessary loading of the wrong page?

apps/laboratory/tests/shared/pages/ModalPage.ts
chris13524: Because we need to load the special configurations that are on these pages

apps/laboratory/tests/verify.spec.ts
chris13524: I don't think we need to over-abstract this. IMO fine as-is

lukaisailovic (approved):

tomiir (approved):

PR #2474, opened 2024-06-27
feat: removed wallet adapters per wallet, added wallet standard support

phoenixVS:
# Breaking Changes

N/A

# Changes

- feat:
- fix:
- chore:

# Associated Issues

closes #...


packages/solana/src/connectors/walletAdapters.ts
```diff
@@ -1,49 +1,33 @@
...
-export function syncInjectedWallets(
-  w3mConnectors: Connector[],
-  adapters: Record<AdapterKey, BaseWalletAdapter>
-) {
-  supportedWallets.forEach(wallet => {
-    if (window[wallet as keyof Window]) {
+export function syncInjectedWallets(w3mConnectors: Connector[], adapters: BaseWalletAdapter[]) {
+  for (const adapter of adapters) {
+    const name = adapter.name.toLocaleLowerCase() as keyof Window
+    if (window[name]) {
```
glitch-txs, line 10: let's add them all, type should be 'EXTERNAL' instead of 'ANNOUNCED'; Adapters shouldn't be shown as _installed_.

apps/laboratory/src/pages/library/solana.tsx
```diff
@@ -5,6 +5,42 @@ import { solana, solanaDevnet, solanaTestnet } from '../../utils/ChainsUtil'
...
 import { ConstantsUtil } from '../../utils/ConstantsUtil'
 import { SolanaTests } from '../../components/Solana/SolanaTests'
+import {
+  AlphaWalletAdapter,
+  AvanaWalletAdapter,
+  BitpieWalletAdapter,
+  CloverWalletAdapter,
+  Coin98WalletAdapter,
+  CoinbaseWalletAdapter,
+  CoinhubWalletAdapter,
```
glitch-txs, line 15: Let's just add 3 or 4 adapters max to labs

packages/solana/src/connectors/walletAdapters.ts
phoenixVS: But wagmi / ethers shows them as "installed" if there is a provider injected in Window

packages/solana/src/connectors/walletAdapters.ts
```diff
@@ -1,49 +1,20 @@
...
```
glitch-txs, line 9: Let's remove this conditional, all the adapters should show up on the modal even when not installed.

apps/laboratory/src/pages/library/solana.tsx
```diff
@@ -22,7 +29,14 @@ const modal = createWeb3Modal({
   enableAnalytics: false,
   termsConditionsUrl: 'https://walletconnect.com/terms',
   privacyPolicyUrl: 'https://walletconnect.com/privacy',
-  customWallets: ConstantsUtil.CustomWallets
+  customWallets: ConstantsUtil.CustomWallets,
+  adapters: [
```
glitch-txs, line 33: Would it be possible to change the naming here from `adapters` to `wallets`?
We will introduce an adapter flag for chains and might collide with this one.

packages/solana/src/connectors/walletAdapters.ts
glitch-txs: we will do this with the Wallet Standard. Wagmi will show the connectors if they were added regardless as well.

glitch-txs (approved):
LGTM

PR #2476, opened 2024-06-28
fix:  `wallet_addEthereumChain` condition

ganchoradkov:
# Breaking Changes

N/A

# Changes
- fix: Updated the condition to send add chain request to only check for user rejection. It is now consistent with [wagmi's connector ](https://github.com/wevm/wagmi/blob/main/packages/connectors/src/walletConnect.ts#L327C9-L328C52)

# Associated Issues

closes #...


glitch-txs (approved):

packages/ethers/src/client.ts
```diff
@@ -1255,19 +1255,13 @@ export class Web3Modal extends Web3ModalScaffold {
...
-              switchError.code === EthersConstantsUtil.ERROR_CODE_DEFAULT ||
-              switchError?.data?.originalError?.code ===
-                EthersConstantsUtil.ERROR_CODE_UNRECOGNIZED_CHAIN_ID
-            ) {
-              await EthersHelpersUtil.addEthereumChain(
-                WalletConnectProvider as unknown as Provider,
-                chain
-              )
-            } else {
+            if (/(?<temp1>user rejected)/u.test(switchError?.message)) {
```
enesozturk, line 1258: @ganchoradkov Why this has difference with the wagmi's own implementation? As I see the wagmi's code have case-insensivity while this one doesn't have:

```js
const switchError1 = { message: "user rejected" };
const switchError2 = { message: "User rejected" };

// First regex (case-sensitive, named group)
const regex1 = /(?<temp1>user rejected)/u;

// Second regex (case-insensitive)
const regex2 = /(user rejected)/i;

// Test cases
console.log(regex1.test(switchError1.message)); // true
console.log(regex2.test(switchError1.message)); // true

console.log(regex1.test(switchError2.message)); // false
console.log(regex2.test(switchError2.message)); // true
```

Do you think we might need this?

enesozturk (changes requested):

packages/ethers/src/client.ts
ganchoradkov: appkit has extra ts rules that force `case-sensitive, named group`

ganchoradkov: or we can go with `//ts-ignore` usage, anyway it is sifficient for the use-case they are updated

ganchoradkov: fixed by lower casing the error msg before the check

PR #2481, opened 2024-06-29
fix wcPromise issue in old project setups

glitch-txs:
# Breaking Changes

N/A

# Changes

- feat:
- fix: Change WalletConnect connection promise logic to support old project environments.
- chore:

# Associated Issues

closes #...

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -34481,7 +34481,7 @@
       },
       "devDependencies": {
         "@types/bn.js": "5.1.5",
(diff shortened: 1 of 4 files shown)


tomiir (approved):
Looks good, just wondering why this promise pattern was used in the first place. Seems like it was configuring some things related to images before awaiting the promise. 
Are you sure this doesn't impact other flows?

enesozturk (approved):

PR #2482, opened 2024-06-30
Add Wallet Standard to AppKit + Solana

glitch-txs:
# Breaking Changes

N/A

# Changes

- feat: Add Wallet Standard support
- fix:
- chore:

# Associated Issues

closes https://github.com/WalletConnect/web3modal/issues/2469


apps/laboratory/src/pages/library/solana.tsx
```diff
@@ -30,13 +25,7 @@ const modal = createWeb3Modal({
...
   privacyPolicyUrl: 'https://walletconnect.com/privacy',
   customWallets: ConstantsUtil.CustomWallets,
-  wallets: [
-    new BackpackWalletAdapter(),
-    new HuobiWalletAdapter(),
-    new PhantomWalletAdapter(),
-    new SolflareWalletAdapter(),
-    new TrustWalletAdapter()
-  ]
+  wallets: [new BackpackWalletAdapter(), new HuobiWalletAdapter(), new SolflareWalletAdapter()]
```
enesozturk, line 28: Why Phantom and Trust has been removed from the array? @glitch-txs

packages/solana/src/client.ts
```diff
@@ -330,9 +318,13 @@ export class Web3Modal extends Web3ModalScaffold {
     return SolStoreUtil.state.connection
   }
 
-  public async checkActiveProviders() {
+  public async checkActiveProviders(standardAdapter?: StandardWalletAdapter[]) {
```
enesozturk, line 321: Maybe a plural name?
```suggestion
  public async checkActiveProviders(standardAdapters?: StandardWalletAdapter[]) {
```

packages/solana/src/utils/wallet-standard/watchStandard.ts
```diff
@@ -0,0 +1,35 @@
...
+      if (!standardAdapters || standardAdapters.length === 0) {
+        standardAdapters = [...wrapWalletsWithAdapters(wallets)]
+      } else {
+        standardAdapters = [...standardAdapters, ...wrapWalletsWithAdapters(wallets)]
+      }
+      callback(standardAdapters)
+    }),
+    on('unregister', (...wallets) => {
+      standardAdapters = standardAdapters.filter(standardAdapter =>
+        wallets.some(wallet => wallet === standardAdapter.wallet)
```
enesozturk, line 22: Is the `.wallet` field string here? It's not understandable name and sounds like it's an object. If it's an object, this equality condition wouldn't work.

apps/laboratory/src/pages/library/solana.tsx
glitch-txs: because they support the Wallet Standard, so to properly test the Wallet Standard.

packages/solana/src/utils/wallet-standard/watchStandard.ts
glitch-txs: > If it's an object, this equality condition wouldn't work.

Why not?

I took this from Solana Labs (Anza) source code btw.

packages/solana/src/client.ts
```diff
@@ -341,13 +333,30 @@ export class Web3Modal extends Web3ModalScaffold {
           this.setWalletConnectProvider(account?.split(':')[2])
         }
       } else {
-        const wallet = walletId?.split('_')[1]
-        const adapter = this.walletAdapters.find(a => a.name.toLocaleLowerCase() === wallet)
-        if (adapter && wallet && window[wallet as keyof Window]) {
+        const walletArray = walletId?.split('_') ?? ''
```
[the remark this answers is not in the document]
glitch-txs: yep, I was being lazy, but `''[0]` is same as `[][0]`

packages/solana/src/client.ts
```diff
@@ -426,44 +461,40 @@ export class Web3Modal extends Web3ModalScaffold {
...
-              this.connectionSettings
-            )
-          )
-          this.setAddress(
-            this.walletAdapters
-              .find(
-                adapter => adapter.name.toLocaleLowerCase() === providerType.toLocaleLowerCase()
-              )
-              ?.publicKey?.toString()
+    if (chain) {
```
enesozturk, line 467: @glitch-txs I think `this.chain` is not same with the `chain` that returned from `getChainFromCaip`. `this.chain` should be used for a reference to `solana` to indicate what chain is this client. See the following line:
```
private chain: AvailableChain = CommonConstantsUtil.CHAIN.SOLANA
```

packages/solana/src/utils/wallet-standard/adapter.ts
```diff
@@ -0,0 +1,548 @@
...
+      this.signIn = this.#signIn
+    } else {
+      delete this.signIn
+    }
+  }
+
+  #changed: StandardEventsListeners['change'] = properties => {
+    // If accounts have changed on the wallet, reflect this on the adapter.
+    if ('accounts' in properties) {
+      const account = this.#wallet.accounts[0]
```
enesozturk, line 246: Why we are checking the first account every time? Let's say there are 3 accounts and 1 is connected. Is the connected one always in the first index?

packages/solana/src/utils/wallet-standard/watchStandard.ts
enesozturk: > Why not?
Because it's JavaScript. In JS worlds two objects are never the same even though they have exact same fields. And the equality condition `===` would never work with objects. They only work with primitive values (strings, boolean, number) Generally if you need such a condition, some special deep equality checker methods would have been used. See https://www.syncfusion.com/blogs/post/deep-compare-javascript-objects

If Solana Labs using that way, prob wallet is not an object, otherwise they are doing wrong too.

packages/solana/src/client.ts
phoenixVS: Here it is used in SolStoreUtil, which does not rewrite this.chain, and stores data about chainId, explorer, etc.
while, as I understood this.chain tells 'EVM' or 'solana' in use. Probably we should rename it to "mode"?

phoenixVS (approved):

packages/solana/src/utils/wallet-standard/adapter.ts
phoenixVS: Yes, when account is switched, wallet.accounts has only one element - with info about active account

packages/solana/src/client.ts
enesozturk: @phoenixVS `mode` wouldn't be a good naming IMO. Solana and EVMs are chains while the Solana Mainnet / Solana DevNet / Polygon etc are networks. 

The `chain` should be actually used for higher scope values; `Ethereum`, `Solana`, `Polkadot`, `Bitcoin` etc.

PR #2489, opened 2024-07-01
fix(ux): truncate and center long emails

segunadebayo:
# Breaking Changes

Nope

# Changes

- Truncate long email
- Center text
- Add Storybook example for line clamp

**BEFORE**
image

**AFTER**
CleanShot 2024-07-01 at 15 43 19@2x

# Associated Issues

Closes https://linear.app/walletconnect/issue/APKT-253/[improvement]-long-emails-are-looking-broken-in-confirm-email-screen


packages/scaffold-ui/src/utils/w3m-email-otp-widget/index.ts
```diff
@@ -71,8 +71,17 @@ export class W3mEmailOtpWidget extends LitElement {
         gap="l"
       >
         <wui-flex flexDirection="column" alignItems="center">
-          <wui-text variant="paragraph-400" color="fg-100">Enter the code we sent to</wui-text>
-          <wui-text variant="paragraph-500" color="fg-100">${this.email}</wui-text>
+          <wui-text variant="paragraph-400" color="fg-100" align="center"
+            >Enter the code we sent to</wui-text
+          >
```
enesozturk, line 76: Smol pick, for better readability: 
```suggestion
          <wui-text variant="paragraph-400" color="fg-100" align="center">
            Enter the code we sent to
          </wui-text>
```

The formatter not able to do this since these are template literals, but at least I could sugges few VSCode extension might help you while building UIs with Lit:
- https://marketplace.visualstudio.com/items?itemName=bierner.lit-html
- https://marketplace.visualstudio.com/items?itemName=bashmish.es6-string-css

packages/scaffold-ui/src/utils/w3m-email-otp-widget/styles.ts
```diff
@@ -4,4 +4,8 @@ export default css`
   wui-loading-spinner {
     margin: 9px auto;
   }
+
+  wui-text.email-text {
+    max-width: 300px;
```
enesozturk, line 9: Why do we prefer max width here? Is it designer's decision? IMO we should avoid static width/heights and let the gaps (margin/padding) handle the responsiveness. AFAIK the modal width have 360px so maybe the parent element might have horizontal padding of `--wui-spacing-xl`?

tomiir: I believe this is because you need to set a specific width in order to have the ellipsis work

segunadebayo: Correct. You need to have a maximum inline width for ellipsis to kick in.

But you have a point, adding some padding might work, but I'm careful not to affect other aspects of the UI right now.

packages/scaffold-ui/src/utils/w3m-email-otp-widget/index.ts
segunadebayo: Good catch. Thank you

packages/scaffold-ui/src/utils/w3m-email-otp-widget/styles.ts
enesozturk: The line-clamp shouldn't depend on the width of the elements. It should work dynamically. You could have a look in Tailwind's examples (by playing with card's width):
- https://tailwindcss.com/docs/line-clamp#basic-usage

But if it's still a case, I'm okay with the decision. Just let avoid static sizing.

enesozturk (approved):

packages/scaffold-ui/src/utils/w3m-email-otp-widget/styles.ts
enesozturk: + we are rendering the modal full width of the screen with `(max-width: 430px)` media query. This makes it even though we have enough space, the email will be truncated, which doesn't sound good IMO. Could we double check for mobile? @segunadebayo 🙏

segunadebayo: Refactored the code to use container based styling instead.

CleanShot 2024-07-02 at 12 09 56@2x

tomiir (approved):
🚀

PR #2492, opened 2024-07-02
chore: fix RandomLink hrefs undefined

chris13524:
# Changes

- chore: fix RandomLink hrefs undefined

### DIFF
--- a/apps/laboratory/src/components/RandomLink.tsx
+++ b/apps/laboratory/src/components/RandomLink.tsx
@@ -6,11 +6,7 @@ import { useState, useEffect } from 'react'
 export function RandomLink({ hrefs, children }: { hrefs: string[]; children: React.ReactNode }) {
   const [href, setHref] = useState<string>()
   useEffect(() => {
-    const newHref = hrefs[Math.floor(Math.random() * hrefs.length)]
-    if (!newHref) {
-      throw new Error('No hrefs provided')
-    }
-    setHref(newHref)
+    setHref(hrefs[Math.floor(Math.random() * hrefs.length)])
   }, [hrefs])
 
   return href ? <Link href={href}>{children}</Link> : <></>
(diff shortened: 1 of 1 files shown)


apps/laboratory/src/components/RandomLink.tsx
```diff
@@ -7,7 +7,7 @@ export function RandomLink({ hrefs, children }: { hrefs: string[]; children: Rea
   const [href, setHref] = useState<string>()
   useEffect(() => {
     const newHref = hrefs[Math.floor(Math.random() * hrefs.length)]
-    if (!href) {
+    if (!newHref) {
```
KannuSingh, line 10: can we do this as its an array `hrefs.length === 0`

KannuSingh: do we wanted to check the hrefs passed to the component? right

chris13524: This wouldn't assert that `newHref` is not-`undefined` because array indexes return `| undefined`. Removed the check entirely since `href` can be `undefined`

chris13524: >do we wanted to check the hrefs passed to the component? right
The original goal was to avoid the type error assigning `string | undefined` to `Link[href]` which is `string`

KannuSingh: I think this will fix it
```
export function RandomLink({ hrefs, children }: { hrefs: string[]; children: React.ReactNode }) {
  const [href, setHref] = useState<string>()
  useEffect(() => {
    if (hrefs.length > 0) {
      setHref(hrefs[Math.floor(Math.random() * hrefs.length)])
    }
  }, [hrefs])

  return href ? <Link href={href}>{children}</Link> : <></>
}

```
we need to check `hrefs.length > 0` before setting  ```setHref(hrefs[Math.floor(Math.random() * hrefs.length)])```

tomiir (approved):

PR #2500, opened 2024-07-03
Feat ERC7715 grant_permissions support on lab

KannuSingh:
# Description
This PR is enabling support for ERC 7715 new JSON-RPC method `wallet_grantPermissions` for DApp to request permissions from a wallet in order to execute transactions on users’s behalf. This enables :
- Executing transactions for users without a wallet connection.

The PR would need addition of one manadory .env values  `NEXT_PUBLIC_PIMLICO_KEY` which is used to submit userOperation.

This PR also enable testing this on a local AA infra running over forked sepolia chain.
For more information on how to set up local testing, refer to [how to setup local-testing](https://docs.pimlico.io/permissionless/how-to/local-testing#extension-testing-against-forked-state).

When running the project **locally** and wanting to use the local forked network and AA setup, just add the `.env` variable values, and the project will automatically use the local setup:

```
NEXT_PUBLIC_LOCAL_BUNDLER_URL=
NEXT_PUBLIC_LOCAL_PAYMASTER_URL=
NEXT_PUBLIC_LOCAL_CLIENT_URL=
```

# Interface changes 

Page: /library/wagmi-permissions  

# Breaking Changes

# Changes

- feat: ERC7715 grant_permissions support on lab
- fix:
- chore:

# Associated Issues

Closes RES-6


apps/laboratory/src/hooks/usePermissions.ts
```diff
@@ -0,0 +1,114 @@
...
+    useUserOpBuilder()
+  const { signer, signerPrivateKey } = useLocalSigner()
+  async function buildAndSendTransactionsWithPermissions(
+    grantedPermissionsResponse: GrantPermissionsReturnType,
+    actions: Execution[]
+  ): Promise<`0x${string}`> {
+    if (!signerPrivateKey || !signer) {
+      throw new Error('No dapp signer key available.')
+    }
+    const apiKey = process.env['NEXT_PUBLIC_PIMLICO_KEY']
```
lukaisailovic, line 27: We don't check if this exists

apps/laboratory/src/components/Wagmi/WagmiRequestPermissionsTest.tsx
```diff
@@ -0,0 +1,156 @@
...
+      const targetPublicKey = signer?.publicKey
+      if (!targetPublicKey) {
+        throw new Error('Local signer not initialized')
+      }
+      const _walletClient = createWalletClient({
+        chain: sepolia,
+        transport: custom(ethereumProvider)
+      }).extend(walletActionsErc7715())
+
+      const grantPermissionsResponse = await _walletClient.grantPermissions({
```
lukaisailovic, line 54: Is this supposed to be `walletClient` or `publicClient`?

lukaisailovic (approved):

lukaisailovic:
PR is missing description

apps/laboratory/src/components/Wagmi/WagmiCreatePrivateKeySignerTest.tsx
```diff
@@ -0,0 +1,47 @@
...
+      toast({
+        title: 'Failure',
+        description: 'Failed to create new signer',
+        type: 'error'
+      })
+    }
+  }
+
+  useEffect(() => {
+    const storedLocalSignerPrivateKey = getItem(LOCAL_SIGNER_KEY)
```
enesozturk, line 32: We are calling `localStorage` which is not available on the server side and I'd expect this will throw an error in the Next.js environment. We might need to make this component client by adding `"use client"` in the first line of the file.

segunadebayo: We probably don't need the "use client" directive since `app/laboratory` uses the pages directory not the app dir in Next.js

This is also run within an effect, so I don't expect it to throw an error.

tomiir (commented):
@lukaisailovic do we want a test for this? It's not appkit specific behavior but if we want it to be tested it would require a playwright test

lukaisailovic:
> @lukaisailovic do we want a test for this? It's not appkit specific behavior but if we want it to be tested it would require a playwright test

@tomiir I don't think its necessary now, since its not a part of AppKit. This is the preparation for integration of Sessions in our Universal Wallet. At that point we will add tests.

apps/laboratory/src/hooks/usePermissions.ts
```diff
@@ -0,0 +1,191 @@
...
+
+    const { signerData, permissionsContext } = permissions
+
+    if (!signerData?.userOpBuilder || !signerData.submitToAddress || !permissionsContext) {
+      throw new Error(`Invalid permissions ${JSON.stringify(permissions, bigIntReplacer)}`)
+    }
+    const userOpHash = getUserOperationHash({
+      userOperation: {
+        ...userOp
+      },
```
enesozturk, line 83: If we don't pass any other fields here, we don't have to destruct the object and directly use the ref. (In case it's not required for cloning object purposes)
```suggestion
    const userOpHash = getUserOperationHash({
      userOperation: userOp,
```

apps/laboratory/src/components/Wagmi/WagmiCreatePrivateKeySignerTest.tsx
enesozturk: Good points. Just to double check if it'd be problematic. Seems like we are okay with it ✅

tomiir (approved):
My only question is:  do we need to clear the signer when the user disconnects? 
Else looks good!

PR #2521, opened 2024-07-08
chore: add clientId to BlockchainAPI requests

tomiir:
# Description
Adds `clientId` query parameter to blockchain API requests when connecting via the WC relay.

## Type of change

# Associated Issues
Closes APKT-672

# Checklist


packages/core/src/controllers/BlockchainApiController.ts
```diff
@@ -32,6 +32,7 @@ import type {
   BlockchainApiRegisterNameParams
 } from '../utils/TypeUtil.js'
 import { OptionsController } from './OptionsController.js'
+import { proxy } from 'valtio'
```
glitch-txs, line 35: all core imports should be from valtio/vanilla to avoid react as required dep

packages/core/src/utils/FetchUtil.ts
```diff
@@ -94,6 +97,9 @@ export class FetchUtil {
         }
       })
     }
+    if (this.clientId) {
```
enesozturk, line 100: @tomiir Is the `clientId` required param? Should we return any error if it's not set correctly?

enesozturk (approved):
Left a comment 🙌🏽

packages/core/src/utils/FetchUtil.ts
tomiir: No, it's only set on WC relay connections so it might not be set

arein (approved):

PR #2525, opened 2024-07-09
Feat/solana onramp

phoenixVS:
# Description

## Type of change

# Associated Issues

For GH issues: closes #...

# Showcase (Optional)

# Checklist


packages/core/src/controllers/OnRampController.ts
```diff
@@ -93,6 +100,15 @@ export const OnRampController = {
...
   setSelectedProvider(provider: OnRampProvider | null) {
+    if (provider && provider.name === 'meld') {
+      const pubKey =
+        process.env['NODE_ENV'] === 'production' ? MELD_PROD_PUBLIC_KEY : MELD_DEV_PUBLIC_KEY
+      provider.url += pubKey
+      const currency = ChainController.state.activeChain === 'solana' ? 'SOL' : 'USDC'
+      provider.url += `&destinationCurrencyCode=${currency}`
+      const address = AccountController.state.address
+      provider.url += `&walletAddress=${address}`
+    }
```
enesozturk, line 111: Consider using the `URL` class to construct the URL. It offers a more maintainable and readable approach for handling query parameters.

```suggestion
if (provider && provider.name === 'meld') {
  const pubKey = process.env['NODE_ENV'] === 'production' ? MELD_PROD_PUBLIC_KEY : MELD_DEV_PUBLIC_KEY;
  const currency = ChainController.state.activeChain === 'solana' ? 'SOL' : 'USDC';
  const address = AccountController.state.address;

  const url = new URL(provider.url);
  url.searchParams.append('publicKey', pubKey);
  url.searchParams.append('destinationCurrencyCode', currency);
  url.searchParams.append('walletAddress', address);

  provider.url = url.toString();
}
```

packages/core/src/utils/ConstantsUtil.ts
```diff
@@ -5,10 +5,23 @@ export const ONRAMP_PROVIDERS = [
...
+    label: 'Meld.io',
+    name: 'meld',
+    feeRange: '1-2%',
+    url: 'https://meldcrypto.com',
+    supportedChains: ['evm', 'solana']
   }
 ]
 
+export const MELD_DEV_PUBLIC_KEY =
+  '?publicKey=WXETMsajb7XcQBm7mcxAab:q3MtzJpiEMtXVNXsqYkAnAaBkgStybGVtZ'
```
enesozturk, line 21: Addition to using `URL` class comment, keeping query param with `?publicKey=` prefix is not a good approach, instead we should generate query params on top, this variable should only keep the key.

packages/ui/src/assets/visual/meld.ts
```diff
@@ -0,0 +1,31 @@
+import { svg } from 'lit'
+export const meldSvg = svg`<svg xmlns="http://www.w3.org/2000/svg" width="40" height="42" viewBox="0 0 40 42" fill="none">
```
enesozturk, line 2: We generally remove the `xmlns="http://www.w3.org` property from the SVGs to optimize the files. Could we remove it?

```suggestion
export const meldSvg = svg`<svg width="40" height="42" viewBox="0 0 40 42" fill="none">
```

enesozturk (changes requested):

packages/core/src/controllers/OnRampController.ts
enesozturk: + use `ConstantsUtil.Chain.SOLANA` for `"solana"` hard coded value

enesozturk (approved):

PR #2529, opened 2024-07-10
chore: SA Tests switch network before flow

tomiir:
# Description

- Simplify flows by switching to supported Network (Sepolia) first 
- Unify signature and signature validation tests

## Type of change

# Checklist


lukaisailovic (approved):

enesozturk (approved):

apps/laboratory/tests/smart-account.spec.ts
```diff
@@ -13,32 +13,59 @@ testModalSmartAccount.beforeEach(async ({ modalValidator }) => {
   await modalValidator.expectConnected()
 })
 
-testModalSmartAccount('it should sign with eoa', async ({ modalPage, modalValidator }) => {
-  await modalPage.sign()
-  await modalPage.approveSign()
-  await modalValidator.expectAcceptedSign()
+testModalSmartAccount('it should use a Smart Account', async ({ modalPage, modalValidator }) => {
+  const walletModalPage = modalPage as ModalWalletPage
```
enesozturk, line 17: @tomiir This tests says `it should use a Smart Account` but it's not switching to SA, is it expected?

tomiir: I changed the fixture so that it switches to Polygon first, which defaults to the SA cause the EOA has no funds

apps/laboratory/tests/smart-account.spec.ts
```diff
@@ -13,32 +13,59 @@ testModalSmartAccount.beforeEach(async ({ modalValidator }) => {
...
+  const signature = await walletModalPage.getSignature()
+  const address = await walletModalPage.getAddress()
+  const chainId = await walletModalPage.getChainId()
+
+  await walletModalValidator.expectValidSignature(signature, address, chainId)
 })
 
 testModalSmartAccount(
-  'it should switch to its smart account and sign',
+  'it should switch to its eoa and sign',
```
enesozturk, line 46: And this one says switch to it's eoa instead of SA, but in the test we are expecting it to be switched to SA

apps/laboratory/tests/smart-account.spec.ts
enesozturk: I see. I didn't know it's SA by default on Polygon. Btw, it's switching to Sepolia, rather than Polygon, it's SA by default also? https://github.com/WalletConnect/web3modal/pull/2529/files#diff-4fcba1bfcbc52350c9de92ef4be0a3fda9781493d37f43eb0a6c433294b411d5R26

tomiir: My bad I meant sepolia (same behavior)

apps/laboratory/tests/smart-account.spec.ts
tomiir: nono, the expect is for what the button should show, so if we expect the button to show "Switch to your EOA" means you are on your SA

PR #2534, opened 2024-07-11
:feat add login with farcaster

svenvoskamp:
# Description

Added support for login with farcaster

## Type of change

# Associated Issues

Closes [APKT-xxx](https://linear.app/walletconnect/issue/APKT-450/login-with-farcaster)

# Checklist


packages/core/src/controllers/RouterController.ts
```diff
@@ -24,6 +24,7 @@ export interface RouterControllerState {
     | 'ChooseAccountName'
     | 'Connect'
     | 'ConnectingExternal'
+    | 'ConnectingFarcaster'
```
enesozturk, line 27: Why this have separate page?

enesozturk: Oh QR, ok

enesozturk: QR is not loaded in preview: https://web3modal-laboratory-c59cqvi45-walletconnect1.vercel.app/library/ethers-all-internal/

enesozturk: Took at least 10sec but loaded

packages/scaffold-ui/src/partials/w3m-social-login-list/index.ts
```diff
@@ -37,12 +48,74 @@ export class W3mSocialLoginList extends LitElement {
...
+    if (socialProvider) {
+      AccountController.setSocialProvider(socialProvider, ChainController.state.activeChain)
+
+      EventsController.sendEvent({
+        type: 'track',
+        event: 'SOCIAL_LOGIN_STARTED',
+        properties: { provider: socialProvider }
+      })
+    }
+    if (socialProvider === 'farcaster') {
```
enesozturk, line 74: Do we have an enum for social options - we'd better to use enums/constants rather than hard coded strings

packages/scaffold-ui/src/partials/w3m-social-login-widget/index.ts
```diff
@@ -153,37 +153,57 @@ export class W3mSocialLoginWidget extends LitElement {
...
       })
-      RouterController.push('ConnectingSocial')
     }
-    const authConnector = ConnectorController.getAuthConnector()
-    this.popupWindow = CoreHelperUtil.returnOpenHref(
-      '',
-      'popupWindow',
-      'width=600,height=800,scrollbars=yes'
-    )
+    if (socialProvider === 'farcaster') {
```
enesozturk, line 163: Same here

packages/scaffold-ui/src/views/w3m-connecting-farcaster-view/index.ts
```diff
@@ -0,0 +1,227 @@
...
+
+  @state() protected loading = false
+
+  public authConnector = ConnectorController.getAuthConnector()
+
+  public constructor() {
+    super()
+    this.unsubscribe.push(
+      ...[
+        AccountController.subscribe(val => {
```
enesozturk, line 41: Can we use `subscribeKey` instead of `subscribe` to lower the re-renders. `subscribeKey` only triggered when the given key updated while `subscribe` triggered when any of the unused fields changed too.

packages/scaffold-ui/src/views/w3m-connecting-farcaster-view/index.ts
```diff
@@ -0,0 +1,227 @@
...
+        }
+      }}
+    >
+      Open farcaster</wui-button
+    >`
+  }
+
+  private onRenderProxy() {
+    if (!this.ready && this.uri) {
+      this.timeout = setTimeout(() => {
```
enesozturk, line 170: Cannot find where I did comment similar but same applies here: if we use `setTimeout` to resolve a problem, IMO we should add comments on top of it to describe why we use this solution. Bc later when we maintain the code, `setTimeout`s are hard to understand why they are used.

enesozturk (approved):

packages/core/src/controllers/RouterController.ts
svenvoskamp: This is Magic latency, don't think we can do anything on our side to make this better

packages/scaffold-ui/src/partials/w3m-social-login-widget/index.ts
```diff
@@ -153,37 +154,57 @@ export class W3mSocialLoginWidget extends LitElement {
...
-      'width=600,height=800,scrollbars=yes'
-    )
+    if (socialProvider === SocialProviderEnum.Farcaster) {
+      RouterController.push('ConnectingFarcaster')
+      const authConnector = ConnectorController.getAuthConnector()
+
+      if (authConnector) {
+        if (!AccountController.state.farcasterUrl) {
+          try {
+            const { url } = await authConnector.provider.getFarcasterUri()
```
tomiir, line 171: this is what we maybe can call earlier in the flow and have the uri available? just spitballing

packages/wallet/src/W3mFrameProvider.ts
```diff
@@ -735,8 +800,11 @@ export class W3mFrameProvider {
     W3mFrameStorage.set(W3mFrameConstants.SOCIAL_USERNAME, username)
   }
 
-  private setEmailLoginSuccess(email: string) {
-    W3mFrameStorage.set(W3mFrameConstants.EMAIL, email)
+  private setEmailLoginSuccess(email?: string | null) {
+    if (email) {
+      W3mFrameStorage.set(W3mFrameConstants.EMAIL, email)
+    }
+
```
tomiir, line 807: Do we want to keep this or add a new one?
I mean you're technically not logging in with email 🤔

Maybe it's a matter of renaming?

packages/scaffold-ui/src/partials/w3m-social-login-widget/index.ts
svenvoskamp: I tried this, but magic/farcaster are calling status calls after we request the farcaster URI.

tomiir (approved):

packages/core/src/controllers/RouterController.ts
svenvoskamp: Added farcaster logo. We can pre-load, but magic/farcaster fires status calls after farcaster URI request. Thats why I prefer to load when user requests it.

PR #2544, opened 2024-07-15
fix: sync accounts in wagmi and subscribe to account change

tomiir:
# Description
Wagmi was not properly syncing the accounts fetched on connection.
It wasn't responding to account switching either. 

- Serialized `accounts` on `syncAccount` method. 
- Added `subscribeShouldUpdateToAddress` to subscribe to account address changes.

## Type of change

- [] Chore (non-breaking change that addresses non-functional tasks, maintenance, or code quality improvements)
- [] New feature (non-breaking change which adds functionality)
- [] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# Checklist [WIP]

 ^ We need to expose multiple accounts on a single network from react-wallet to add automated tests

### DIFF
--- a/apps/laboratory/tests/shared/validators/ModalValidator.ts
+++ b/apps/laboratory/tests/shared/validators/ModalValidator.ts
@@ -109,7 +109,6 @@ export class ModalValidator {
   }
 
   async expectMultipleAccounts() {
-    await this.page.waitForTimeout(1500)
     const accounts = await this.page.getByTestId('switch-address-item').all()
 
     expect(accounts.length).toBeGreaterThan(1)
--- a/apps/laboratory/tests/shared/validators/ModalValidator.ts
+++ b/apps/laboratory/tests/shared/validators/ModalValidator.ts
@@ -109,6 +109,9 @@ export class ModalValidator {
   }
 
   async expectMultipleAccounts() {
+    await expect(this.page.getByText('Switch Address')).toBeVisible({
+      timeout: MAX_WAIT
+    })
     const accounts = await this.page.getByTestId('switch-address-item').all()
 
     expect(accounts.length).toBeGreaterThan(1)
--- a/apps/laboratory/tests/shared/validators/ModalValidator.ts
+++ b/apps/laboratory/tests/shared/validators/ModalValidator.ts
@@ -107,4 +107,11 @@ export class ModalValidator {
     const externalConnector = this.page.getByTestId(/^wallet-selector-external/u)
     await expect(externalConnector).toBeVisible()
   }
+
+  async expectMultipleAccounts() {
+    await this.page.waitForTimeout(1500)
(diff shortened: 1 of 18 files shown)


apps/laboratory/tests/shared/validators/ModalValidator.ts
```diff
@@ -107,4 +107,11 @@ export class ModalValidator {
     const externalConnector = this.page.getByTestId(/^wallet-selector-external/u)
     await expect(externalConnector).toBeVisible()
   }
+
+  async expectMultipleAccounts() {
+    await this.page.waitForTimeout(1500)
```
enesozturk, line 112: Why do we have timeout here? Is it for page transition or loading the accounts etc? Since we have just a few timeouts in the tests (and they are mostly required), we should write a comment on top of them to make sure in the feature it'll make the maintenance easier, wdyt?

packages/wagmi/src/client.ts
```diff
@@ -377,6 +379,26 @@ export class Web3Modal extends Web3ModalScaffold {
     })
 
     this.setEIP6963Enabled(w3mOptions.enableEIP6963 !== false)
+
+    this.subscribeShouldUpdateToAddress((newAddress?: string) => {
```
enesozturk, line 383: Is this something we need to implement for other clients (ethers/ethers5)?

packages/wagmi/src/client.ts
```diff
@@ -430,11 +455,16 @@ export class Web3Modal extends Web3ModalScaffold {
...
         this.syncProfile(address, chainId),
         this.syncBalance(address, chainId),
-        this.syncConnectedWalletInfo(connector),
         this.setApprovedCaipNetworksData()
       ])
+      if (connector) {
+        this.syncConnectedWalletInfo(connector)
+      }
+      if (addresses?.length) {
+        this.setAllAccounts(addresses.map(addr => ({ address: addr, type: 'eoa' })))
```
enesozturk, line 464: Do we sure this wouldn't break the SA flows by forcing the accounts switch to EOA? Like if there is any changes to accounts, we are updating them with EOA type, but what about the SA ones?

enesozturk (approved):

apps/laboratory/tests/shared/validators/ModalValidator.ts
tomiir: NICE CATCH. This was for testing only

packages/wagmi/src/client.ts
tomiir: was already implemented, I added this cause it was missing 🖖

packages/wagmi/src/client.ts
tomiir: Good point, will take a look as I just moved code around.

PR #2545, opened 2024-07-16
[TDW] move from npm to pnpm

segunadebayo:
# Description

This PR marks the start of improvements in our overall development experience. This will happen in two steps (and two pull requests):

- **Migration to PNPM (current):** This will reveal hoisted dependencies that weren't installed explicitly and could cause issues on the consumer side. It also helps rely less on turbo since pnpm has a task runner system (and it's fast)

- **Remove the need to run `watch` and read from source directly:** This will greatly improve TS performance and the need to "Restart TS Server" when working across packages. Of course, this will come in a different PR.

My end goal is to make our lives a bit easier. It's probably not much, but it's worth it.

For this PR, I've done the following:

- Moved workspace definition to `pnpm-workspace.yaml`
- Change all `@web3modal/*` package to point to `workspace:*`
- Renamed the commands across all packages to be shorter and compatible with the pnpm task runner
- (Debatable) Used pnpm task runner everywhere, it's almost as fast as turbo but without the caching (since we always use `cache: false` in turbo). Happy to bring back turbo once everything works as expected.

## Type of change

# Associated Issues

For GH issues: closes #...

# Showcase (Optional)

# Checklist


.npmrc
```diff
@@ -1 +1,2 @@
 script-shell=sh
+shamefully-hoist=true
```
segunadebayo, line 2: This was the only way I could get `packages/cdn` to build. Without it, it throws some vite errors related to the node-polyfill.

Setting shamefully hoist makes it easy to use code from a dependency that has not been explicitly added to `package.json`

I'm unhappy with that, but I'll fix the cdn build in a separate PR/task.

Dockerfile.canary
```diff
@@ -1,15 +1,19 @@
 FROM node:20-bookworm
+ENV PNPM_HOME="/pnpm"
+ENV PATH="$PNPM_HOME:$PATH"
+RUN corepack enable
```
segunadebayo, line 4: Got these from https://pnpm.io/docker

examples/vue-solana/src/App.vue
```diff
@@ -1,7 +1,7 @@
 <script lang="ts" setup>
 import {
   createWeb3Modal,
-  defaultConfig,
+  defaultSolanaConfig,
```
segunadebayo, line 4: There was no export of `defaultConfig` from the solane/vue package. The correct import is `defaultSolanaConfig`

Mostly mirrored the set up from `examples/react-solana`

package.json
```diff
@@ -63,8 +44,10 @@
     "eslint-plugin-prettier": "5.1.3",
     "eslint-plugin-require-extensions": "0.1.3",
     "prettier": "3.1.1",
+    "test-exclude": "7.0.1",
```
segunadebayo, line 47: For some reason, `vitest` required this. I'm guessing it's related to the dependency hoisting again.

packages/solana/package.json
```diff
@@ -19,6 +19,11 @@
       "import": "./dist/esm/exports/react.js",
       "default": "./dist/esm/exports/react.js"
     },
+    "./vue": {
+      "types": "./dist/types/exports/vue.d.ts",
+      "import": "./dist/esm/exports/vue.js",
+      "default": "./dist/esm/exports/vue.js"
+    },
```
segunadebayo, line 26: the `/vue` subpath was not exported. I wonder how this ever worked. The build failed without this.

Dockerfile.canary
```diff
@@ -1,15 +1,19 @@
...
+
 RUN apt update && apt install -y awscli
 
 WORKDIR /src
 
 COPY ../ ./
-RUN npm ci
-RUN npm run build
+RUN pnpm install
```
tomiir, line 11: should this be a clean install too?

.npmrc
tomiir: @enesozturk @svenvoskamp any clues as to why this fails with the vite errors?

package.json
```diff
@@ -4,54 +4,35 @@
...
+    "build:gallery": "pnpm build; pnpm --filter=./apps/gallery build",
+    "build:laboratory": "pnpm build; pnpm --filter=./apps/laboratory build",
+    "build:demo": "pnpm build; pnpm --filter=./apps/demo build",
+    "build:examples": "pnpm build; pnpm --filter=./examples -r build",
+    "test": "pnpm -r test --if-present",
     "coverage:merge": "node ./scripts/coverage-merge.js",
     "coverage:summary": "node ./scripts/coverage-summary.js",
-    "typecheck": "turbo run typecheck",
-    "lint": "turbo run lint",
+    "typecheck": "pnpm --if-present -r typecheck",
```
tomiir, line 22: why `--if-present` ?

packages/scaffold-utils/package.json
```diff
@@ -33,15 +33,17 @@
...
+    "clean": "rm -rf dist",
+    "build": "tsc --build",
     "watch": "tsc --watch",
     "typecheck": "tsc --noEmit",
     "lint": "eslint . --ext .js,.jsx,.ts,.tsx"
   },
   "dependencies": {
-    "@web3modal/core": "5.0.7",
-    "@web3modal/polyfills": "5.0.7",
+    "@coinbase/wallet-sdk": "4.0.3",
```
tomiir, line 43: why add this dependency?

packages/solana/package.json
tomiir: I believe this was recently merged by @phoenixVS

package.json
segunadebayo: because not every package in the workspace has that command

packages/scaffold-utils/package.json
segunadebayo: https://github.com/WalletConnect/web3modal/blob/4a242fa2da1fef410e8df7c489204b0f3e4ad68c/packages/scaffold-utils/src/EthersTypesUtil.ts#L2

packages/solana/package.json
segunadebayo: Great. I'll rebase accordingly

.npmrc
segunadebayo: To be specific, here's the error

CleanShot 2024-07-18 at 17 41 46@2x

lukaisailovic:
Changes look good.
Should we add a check in dangerfile if someone accidentally commit npm/yarn lock?

segunadebayo:
@lukaisailovic I believe this is already covered here
CleanShot 2024-07-19 at 15 54 21@2x

Dockerfile.canary
enesozturk: Wdym clean install? 🤔

apps/gallery/package.json
```diff
@@ -4,14 +4,15 @@
...
-    "dev:gallery": "storybook dev -p 6006",
-    "build:gallery": "storybook build -o out",
+    "dev": "storybook dev -p 6006",
+    "build": "storybook build -o out",
     "lint": "eslint . --ext .js,.jsx,.ts,.tsx"
   },
   "dependencies": {
-    "@web3modal/common": "5.0.7",
-    "@web3modal/ui": "5.0.7",
+    "@web3modal/common": "workspace:*",
```
enesozturk, line 12: Finally. Love this feature.

package.json
enesozturk: Shouldn't the turborepo already do this conditional run? If there is no script in a package, it's just not running as I know, prob something specific to pnpm?

package.json
```diff
@@ -4,57 +4,39 @@
...
+    "build:laboratory": "pnpm build:packages; pnpm --filter=./apps/laboratory build",
+    "build:demo": "pnpm build:packages; pnpm --filter=./apps/demo build",
+    "build:examples": "pnpm build:packages; pnpm --filter=./examples -r build",
+    "test": "pnpm -r test --if-present",
     "coverage:merge": "node ./scripts/coverage-merge.js",
     "coverage:summary": "node ./scripts/coverage-summary.js",
-    "typecheck": "turbo run typecheck",
-    "lint": "turbo run lint",
+    "typecheck": "pnpm --if-present -r typecheck",
+    "lint": "pnpm -r lint",
```
enesozturk, line 23: [-r parameter](https://pnpm.io/cli/remove#--recursive--r) not really clear for me. Do we have to use it? It looks like we need to be careful about, could you explain why we used it?

segunadebayo: `--recursive` flag is used to run a script in every workspace package.

package.json
segunadebayo: This is specific to `pnpm`

svenvoskamp (approved):
LGTM

tomiir (approved):

Dockerfile.canary
tomiir: `npm ci` does a clean install in CI envs. Wondering if we need to do the same (-i think it's `pnpm --fronzen-lockfile`)

package.json
enesozturk: Ah yes, the way they described is quite quite different just why:
> When used inside a [workspace](https://pnpm.io/workspaces), removes a dependency (or dependencies) from every workspace package.

PR #2570, opened 2024-07-19
feat: add unit tests for clients

enesozturk:
# Description

This PR introduces unit testing which checks the initial state of the controllers after initialization of ethers, wagmi, and solana packages.

We are a bit limited to check life cycle updates for these clients. These unit tests only cover the things updates of the first snapshot after initialization (after constructor being called). Plus, the client files are not quite flexible to be fully covered by unit tests bc triggering some functions require user/wallet actions and these are not possible on unit tests or not tooling is not easy to setup.

Here are the covered things:
- [x] Check if the the flags are being set correctly
- [x] Check if the networks are being set correctly
- [x] Check if the connectors are being set correctly

## Type of change

# Associated Issues

For Linear issues: Closes APKT-777

# Showcase (Optional)

# Checklist


packages/ethers/tests/client.test.ts
```diff
@@ -0,0 +1,103 @@
...
+    allWallets: 'SHOW',
+    includeWalletIds: ['8a0ee50d1f22f6651afcae7eb4253e52a3310b90af5daef78a8c4929a9bb99d4'],
+    excludeWalletIds: ['c03dfee351b6fcc421b4494ea33b9d4b92a984f87aa76d1663bb28705e95034a'],
+    featuredWalletIds: ['1ae92b26df02f0abca6304df07debccd18262fdf5fe82daa81593582dac9a369'],
+    allowUnsupportedChain: true
+  })
+})
+
+// -- Tests --------------------------------------------------------------------
+describe('createWeb3Modal by wagmi', () => {
```
zoruka, line 62:
```suggestion
describe('createWeb3Modal by ethers', () => {
```

svenvoskamp (approved):

.vscode/settings.json
```diff
@@ -6,7 +6,6 @@
   "eslint.validate": ["javascript", "javascriptreact", "typescript", "typescriptreact"],
   "explorer.sortOrder": "type",
   "extensions.ignoreRecommendations": true,
-  "files.autoSave": "onWindowChange",
```
svenvoskamp, line 9: IMO we can remove this one

enesozturk: You reverting the changes? @svenvoskamp

packages/wagmi/tests/client.test.ts
```diff
@@ -0,0 +1,100 @@
...
+    expect(OptionsController.state.disableAppend).toEqual(undefined)
+    expect(NetworkController.state.allowUnsupportedChain).toBe(true)
+  })
+
+  it('should set connectors as expected', () => {
+    const authConnector = ConnectorController.state.connectors.find(c => c.type === 'AUTH')
+    expect(ConnectorController.state.connectors.length).toBeGreaterThan(0)
+    expect(authConnector).toBeDefined()
+    expect(authConnector?.socials).toEqual(['google', 'x', 'discord', 'apple', 'github'])
+  })
```
tomiir, line 87: should we do some combinations of connectors to check if they are properly set? 
Also, would check for CB connector

enesozturk: We should def do that, I've tried this in the TDW but was more complex then I thought. What I want to do is that refactoring whole these tests when we switch to new architecture. We'll need to remove these files and move to adapters, and we could refactor the tests in that time as well.

packages/ethers/package.json
```diff
@@ -40,6 +40,7 @@
     "build": "tsc --build",
     "watch": "tsc --watch",
     "typecheck": "tsc --noEmit",
+    "test": "vitest run --dir tests --coverage.enabled=true --coverage.reporter=json --coverage.reporter=json-summary --coverage.reportOnFailure=true",
```
tomiir, line 43: will need to add vitest and vitest-coverage to devDeps

packages/solana/package.json
```diff
@@ -48,6 +48,7 @@
     "build": "tsc --build",
     "watch": "tsc --watch",
     "typecheck": "tsc --noEmit",
+    "test": "vitest run --dir tests --coverage.enabled=true --coverage.reporter=json --coverage.reporter=json-summary --coverage.reportOnFailure=true",
```
tomiir, line 51: same

packages/wagmi/package.json
```diff
@@ -48,6 +48,7 @@
     "build": "tsc --build",
     "watch": "tsc --watch",
     "typecheck": "tsc --noEmit",
+    "test": "vitest run --dir tests --coverage.enabled=true --coverage.reporter=json --coverage.reporter=json-summary --coverage.reportOnFailure=true",
```
tomiir, line 51: same

enesozturk:
We are in transition of the switching to multi-chain architecture and within that, we want to refactor whole clients (aka adapters in multi-chain structure). Thus, let's close this and create tests under adapters with the refactor. I feel like the income of this effort might not worth it to push right now. Will create a refactor ticket which includes the unite testing for the adapters as well.

PR #2572, opened 2024-07-19
refactor/wallet card item with image optimization

enesozturk:
# Description

Introduces refactor to our wallet cards to optimize initially fetched images to reduce unnecessary downloaded images when "All Wallets" page opened. This will provide more than 50% save from downloaded image sources :chefskiss:

Here is the results when compare the previous version:
- Previously: `~234kb`
- Now: `~109kb`

## Type of change

# Associated Issues

For Linear issues: Closes APKT-545
For GH issues: closes #...

# Showcase (Optional)

# Checklist


packages/core/tests/utils/AssetUtil.test.ts
```diff
@@ -0,0 +1,99 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest'
```
enesozturk, line 1: Not related to this optimization work. Adding missing unit test for `AssetUtil`

packages/core/src/utils/AssetUtil.ts
```diff
@@ -1,7 +1,25 @@
...
+
+    return undefined
+  },
+
+  getWalletImageById(imageId?: string) {
+    if (imageId) {
+      return AssetController.state.walletImages[imageId]
+    }
+
+    return undefined
```
tomiir, line 20:
```suggestion

    return AssetController.state?.walletImages?.[imageId]
```

packages/core/tests/utils/AssetUtil.test.ts
```diff
@@ -0,0 +1,99 @@
...
+    networkImages: {
+      ethereum: 'ethereum-logo-blob-url'
+    }
+  }
+})
+
+// -- Tests --------------------------------------------------------------------
+describe('AssetUtil', () => {
+  it('should get the connector image from connector object', () => {
+    // @ts-expect-error it's a partial connector object
```
tomiir, line 61: I'd apply the disable globally instead of on every test

packages/scaffold-ui/src/partials/w3m-all-wallets-list-item/index.ts
```diff
@@ -0,0 +1,101 @@
...
+
+  // -- Lifecycle ----------------------------------------- //
+  constructor() {
+    super()
+    this.observer = new IntersectionObserver(
+      entries => {
+        entries.forEach(entry => {
+          if (entry.isIntersecting) {
+            this.visible = true
+            this.fetchImageSrc()
```
tomiir, line 32: Is there any way we could request this data concurrently? 
I mean we have like 4 wallets per line, would be 4 requests, could we request those 4 concurrently? wyt?

packages/core/tests/utils/AssetUtil.test.ts
enesozturk: Good idea, I don't have strong opinion but I wouldn't want to discard all benefits of the TS, instead stoping type checking (with `@ts-nocheck` on top of the file), only use disable TS for a single thing sounds a better IMO

packages/core/src/utils/AssetUtil.ts
enesozturk: The `image_id` is optional field in the wallet object, there is always a case that it could be undefined, thus it's necessary to check if the imageId is valid to make a property access

packages/scaffold-ui/src/partials/w3m-all-wallets-list-item/index.ts
svenvoskamp: I think we need to move the intersectionObserve to the parent component `w3m-all-wallets-list`. We can add the id of the wallet as a data-attribute, so we can store this data and fetch 4 at the same time?

svenvoskamp (approved):
LGTM, reducing the fetch requests would be a nice improvement

packages/core/tests/utils/AssetUtil.test.ts
tomiir: given it's a partial you could maybe cast it to `Partial<Connector>` ?
Else agree with your point but spamming disables is kind of the same thing 🤣

PR #2573, opened 2024-07-19
feat: enableSwaps option

enesozturk:
# Description

Introduces `enableSwaps` flag which allow users to enable/disable swap feature. Also we are adding swap feature to the default account view so even with regular wallets, users will be able to use swaps.

By default, swaps are disabled on regular flow, enabled on email flow.

## Type of change

# Associated Issues

For Linear issues: Closes APKT-752, Closes, APKT-618
For GH issues: closes #...

# Showcase (Optional)

# Checklist


packages/core/src/controllers/SwapController.ts
```diff
@@ -655,16 +660,20 @@ export const SwapController = {
...
+    if (isAuthConnector) {
+      RouterController.pushTransactionStack({
+        view: null,
+        goBack: true,
+        onSuccess() {
+          SnackController.showLoading('Approving transaction...')
+        }
+      })
+    } else {
+      SnackController.showLoading('Approve spending increase')
```
tomiir, line 675: Do we need any loading indicator on the screen as well?

packages/core/src/controllers/SwapController.ts
```diff
@@ -702,14 +714,18 @@ export const SwapController = {
...
+      RouterController.pushTransactionStack({
+        view: 'Account',
+        goBack: false,
+        onSuccess() {
+          SnackController.showLoading(snackbarPendingMessage)
+          SwapController.resetState()
+        }
+      })
+    } else {
+      SnackController.showLoading(snackbarPendingMessage)
```
tomiir, line 727: same

packages/core/src/controllers/SwapController.ts
```diff
@@ -739,6 +755,9 @@ export const SwapController = {
         }
       })
       SwapController.resetState()
+      if (!isAuthConnector) {
+        RouterController.replace('Account')
+      }
```
tomiir, line 760: What happens if it's the auth connector?

enesozturk: If it's auth connector, we are handing the success and error states on top with `RouterController.pushTransactionStack`, there is no need to call something here for auth connectors bc the way they are working is different.

packages/core/src/controllers/SwapController.ts
enesozturk: Yes, this is just a feedback to the users that let them know they need to approve the TX, toast will be visible for few seconds.

packages/core/src/controllers/SwapController.ts
enesozturk: Yes, this is just a feedback to the users that let them know they need to approve the TX, toast will be visible for few seconds.

tomiir (approved):

PR #2615, opened 2024-07-30
Wagmi: Erc7715 permissions with Appkit embedded wallet

KannuSingh:
# Description

Added wagmi-permissions-email implementation to support erc7715 permissions flow with walletconnect co-signer service using Appkit Universal wallet.
Dapp will use passkey as a signer and walletConnect co-signer will use ECDSA key as a signer on userOperation submitted.
This should be tested with Appkit Universal wallet, once the PR for universal wallet is merged

## Type of change

# Associated Issues

For GH issues: closes #...

# Showcase (Optional)

# Checklist


KannuSingh:
The reason it have separate `wagmi-permissions`and `wagmi-permissions-email` is bc we don't have [PR](https://github.com/WalletConnect/walletconnect-monorepo/pull/4576) merged on ethereum-provider rn and for sample wallet connector on wagmiConfig is configured manually to include support for erc7715 method. 
I was unable to do so for wagmiConfig for email(Appkit embedded wallet), if it can be done we can then merge both features

lukaisailovic:
@KannuSingh What does that mean? Can we test/use both wallets?

packages/wagmi/src/client.ts
```diff
@@ -754,6 +754,14 @@ export class Web3Modal extends Web3ModalScaffold {
...
             break
           }
+          case W3mFrameConstants.RPC_RESPONSE_TYPE_OBJECT: {
+            if (super.isTransactionStackEmpty()) {
+              super.close()
+            } else {
+              super.popTransactionStack()
+            }
+            break
+          }
```
tomiir, line 764: @enesozturk @svenvoskamp any issue with this? will this affect swap / send or other flows?

apps/laboratory/src/utils/CommonUtils.ts
```diff
@@ -30,3 +30,64 @@ export function encodeSecp256k1PublicKeyToDID(publicKey: string) {
   // Construct the did:key
   return `did:key:zQ3s${encodedPublicKey}`
 }
+
+export function encodePublicKeyToDID(publicKey: string, keyType: string): string {
```
tomiir, line 34: nit: this could be a `CryptoUtil` or `EncodingUtil`

apps/laboratory/src/utils/CommonUtils.ts
```diff
@@ -30,3 +30,64 @@ export function encodeSecp256k1PublicKeyToDID(publicKey: string) {
...
+  const didPrefix = keyTypeToDIDPrefix[keyType]
+
+  // Construct the did:key
+  return `${didPrefix}${encodedPublicKey}`
+}
+
+export function decodeUncompressedPublicKey(uncompressedPublicKey: string): `0x${string}` {
+  const uncompressedPublicKeyBuffer = Buffer.from(uncompressedPublicKey, 'base64')
+
+  if (uncompressedPublicKeyBuffer.length !== 65) {
```
tomiir, line 70: I'd make 65 a constant like PUBLIC_KEY_LENGTH

apps/laboratory/src/utils/CommonUtils.ts
```diff
@@ -30,3 +30,64 @@ export function encodeSecp256k1PublicKeyToDID(publicKey: string) {
...
+    throw new Error('Invalid uncompressed public key length')
+  }
+
+  const header = uncompressedPublicKeyBuffer[0]
+  if (header !== 0x04) {
```
tomiir, line 75: I'd set `0x04` as a constant like `PUBLIC_KEY_HEADER_HEAD` or similar
in general it's best not to have literals lying around in the code (even more for crypto stuff)

apps/laboratory/src/hooks/useWalletConnectCosigner.ts
```diff
@@ -0,0 +1,238 @@
...
+  async function coSignUserOperation(
+    address: string,
+    projectId: string,
+    coSignData: CoSignRequest
+  ): Promise<CoSignResponse> {
+    const url = `${baseUrl}/${encodeURIComponent(address)}/sign`
+
+    try {
+      const response = await axios.post<CoSignResponse>(url, coSignData, {
+        params: { projectId },
```
tomiir, line 209: seems most code is repeated in these functions. I'd abstract to some common api util or similar and call it with different params from here

tomiir: Applies for all of the functions in this file

apps/laboratory/src/hooks/useWalletConnectCosigner.ts
```diff
@@ -0,0 +1,238 @@
...
+  constructor(
+    public status: number,
+    message: string
+  ) {
+    super(message)
+    this.name = 'CoSignerApiError'
+  }
+}
+
+export function useWalletConnectCosigner() {
```
tomiir, line 73: address and projectId are repeated all along, would it make sense to make them params and state vars instead of requiring them in all methiods?

apps/laboratory/src/hooks/useWagmiActiveCapabilities.ts
```diff
@@ -65,9 +74,12 @@ export function useWagmiAvailableCapabilities({
...
     if (provider instanceof W3mFrameProvider) {
-      return ['wallet_sendCalls', 'wallet_getCapabilities', 'wallet_getCallsStatus'].includes(
-        method
-      )
+      return [
+        'wallet_sendCalls',
+        'wallet_getCapabilities',
+        'wallet_getCallsStatus',
+        'wallet_grantPermissions'
+      ].includes(method)
```
tomiir, line 82: can we move these to a constant?

apps/laboratory/src/context/WagmiPermissionsContext.tsx
```diff
@@ -0,0 +1,105 @@
...
+  setGrantedPermissions: React.Dispatch<
+    React.SetStateAction<GrantPermissionsReturnType | undefined>
+  >
+  setWCCosignerData: React.Dispatch<React.SetStateAction<AddPermissionResponse | undefined>>
+  setPasskey: (value: PasskeyStorageType) => void
+  clearGrantedPermissions: () => void
+}
+function noop() {
+  console.warn('WagmiPermissionsContext used outside of provider')
+}
```
tomiir, line 32: is this needed?

apps/laboratory/src/context/WagmiPermissionsContext.tsx
```diff
@@ -0,0 +1,105 @@
...
+        setWCCosignerData
+      }}
+    >
+      {children}
+    </WagmiPermissionsContext.Provider>
+  )
+}
+export function useWagmiPermissions() {
+  const context = useContext(WagmiPermissionsContext)
+  if (context === undefined) {
```
tomiir, line 100: could this be if !context? or do you need to explicitly check against undefined?

apps/laboratory/src/hooks/useWalletConnectCosigner.ts
KannuSingh: I think it better to accept these params on methods for now, will refactor it if structure changes later(expected).

apps/laboratory/src/hooks/useWalletConnectCosigner.ts
KannuSingh: I think its already abstracted, will not affect much if tried to abstract it more.

lukaisailovic: @KannuSingh I think @tomiir was referencing the request code specifically. If you check your methods that interact with co-signer, you will notice that about 90% of the code is the same. The only difference is the request. This can be handled with a new method to interact with co-signer and usage of generics to provide type

lukaisailovic: This should be extracted to a util, named something like `sendCoSignerRequest<RequestType>` or similar

apps/laboratory/src/hooks/useWalletConnectCosigner.ts
lukaisailovic: > I think it better to accept these params on methods for now
Why? Its fine sharing an opinion but we need to explain the reasoning behind it.

apps/laboratory/src/hooks/useWalletConnectCosigner.ts
KannuSingh: extracted the logic and create the similar method

packages/wagmi/src/client.ts
KannuSingh: @chris13524 this might be causing the issue "After switching networks it closes the modal as opposed to showing the switch button again?"

KannuSingh: @tomiir I think this is causing the few test to fails, I tried removing this particular change and then run test (eventually all the test passed), but the issue of modal not getting close was introduced.
Please let me know if there is any other way to handle this change.

apps/laboratory/src/components/Wagmi/WagmiPurchaseDonutSyncPermissionsTest.tsx
```diff
@@ -68,6 +67,7 @@ export function WagmiPurchaseDonutWithPermissionsTest() {
         await fetchDonutsOwned()
       }
     } catch (error) {
+      // Console.log(error)
```
tomiir, line 70: Remove comment 🙏

apps/laboratory/src/hooks/useERC7715PermissionsAsync.ts
```diff
@@ -43,6 +48,16 @@ export function usePermissions() {
       actions
     })
 
+    /*
```
tomiir, line 51: Is this needed?

apps/laboratory/src/hooks/useERC7715PermissionsSync.ts
```diff
@@ -0,0 +1,181 @@
...
+
+    const callData = await getCallDataWithContext(publicClient, {
+      userOpBuilderAddress: signerData.userOpBuilder,
+      sender: signerData.submitToAddress,
+      permissionsContext: permissionsContext as `0x${string}`,
+      actions
+    })
+
+    /*
+     * Comment
```
tomiir, line 51: Same

packages/wagmi/src/client.ts
```diff
@@ -742,21 +742,8 @@ export class Web3Modal extends Web3ModalScaffold {
...
-              super.close()
-            } else {
-              super.popTransactionStack()
-            }
-            break
-          }
-          default:
-            break
-        }
+      provider.onRpcSuccess(() => {
```
tomiir, line 745: This needs to be done on all clients

tomiir (approved):

PR #2621, opened 2024-07-30
:hotfix fix svg error when requesting farcaster qr code

svenvoskamp:
# Description

Add a `setTimeout` to `onRenderProxy`. This setTimeout is needed to avoid the beginning of the animation from not starting to resize immediately and some weird svg errors

## Type of change

# Associated Issues

For Linear issues: Closes [APKT-xxx](https://linear.app/walletconnect/issue/APKT-818/farcaster-show-a-blank-qr-view-on-the-second-click)

# Checklist


tomiir (approved):

tomiir:
@svenvoskamp @enesozturk  this is a workaround in many animations right?

WYT about making it some generic util so instead of having some timeouts on some components we just have a call to 
`resyncAnimation` or similar. 

Any way we can fix this at animation level? no, right?

enesozturk:
> @svenvoskamp @enesozturk this is a workaround in many animations right?
> 
> WYT about making it some generic util so instead of having some timeouts on some components we just have a call to `resyncAnimation` or similar.
> 
> Any way we can fix this at animation level? no, right?

I was not aware this issue on the JS side tbh. Cannot say I understood the problem 100%, is it something related to call stack? @svenvoskamp 

We could consider moving to fully CSS animations, as I remember the dynamic heights is a bit painful so we just use JS animations for the modal heights. But there might still be a chance we could do that.

svenvoskamp:
> > @svenvoskamp @enesozturk this is a workaround in many animations right?
> > WYT about making it some generic util so instead of having some timeouts on some components we just have a call to `resyncAnimation` or similar.
> > Any way we can fix this at animation level? no, right?
> 
> I was not aware this issue on the JS side tbh. Cannot say I understood the problem 100%, is it something related to call stack? @svenvoskamp
> 
> We could consider moving to fully CSS animations, as I remember the dynamic heights is a bit painful so we just use JS animations for the modal heights. But there might still be a chance we could do that.

Yes it is the call stack. Otherwise it will have the wrong values and end up full white. This happens for qr code as well. CSS animations would be best the best solution, but quite hard to achieve I think.

enesozturk:
> but quite hard to achieve I think

Challange accepted

enesozturk (approved):

PR #2632, opened 2024-07-31
chore: change chainId type to accept string and its dependencies

zoruka:
# Description

This is the first step of a two steps process to change W3Frame chainId type from number to string. First we must update from `number` to `string | number`, push a new version and update its dependencies then change it to `string`.

## Type of change

# Associated Issues

For Linear issues: APKT-826

# Showcase (Optional)

# Checklist


packages/ethers/src/client.ts
```diff
@@ -912,7 +912,17 @@ export class Web3Modal extends Web3ModalScaffold {
...
-        EthersStoreUtil.setChainId(chainId)
+
+        /**
+         * Warning: This is a temporary fix to handle both string and number chainId
+         * `chainId` will be only string after the dependencies are updated
+         */
+        if (typeof chainId === 'string') {
+          EthersStoreUtil.setChainId(NetworkUtil.caipNetworkIdToNumber(chainId as CaipNetworkId))
+        } else {
+          EthersStoreUtil.setChainId(chainId)
```
tomiir, line 923: this is repeated in wagmi and here => I'd abstreact it to `@web3modal/common/NetworkUtil`

tomiir (approved):

packages/wallet/src/W3mFrameProvider.ts
```diff
@@ -422,7 +422,7 @@ export class W3mFrameProvider {
     })
   }
 
-  public async switchNetwork(chainId: number) {
+  public async switchNetwork(chainId: number | string) {
```
enesozturk, line 425: Could we have `ChainId` type somewhere in common place? I guess we already have that type but should be for something else. Let's double check if we could use in these cases, to not repeate type definitions

packages/wallet/src/W3mFrameTypes.ts
```diff
@@ -125,7 +126,7 @@ export namespace W3mFrameTypes {
 
   export interface Network {
     rpcUrl: string
-    chainId: number
+    chainId: number | CaipNetworkId
```
enesozturk, line 129: Why it could be `1` (number) and `eip155:1` (caip string)? In which cases I mean, isn't the reason we do it `number | string` that Solana chains have IDs as string? I'm confused a bit for this

enesozturk (approved):

PR #2633, opened 2024-07-31
chore: fix playwright tests

chris13524:
- Tests must not have arbitrary delays. This also [resulted](https://walletconnect.slack.com/archives/C03RVH94K5K/p1722432260728989?thread_ts=1722368999.184019&cid=C03RVH94K5K) in mis-calculation of the canary latency.
- The canary test environment must not be modified


apps/laboratory/tests/siwe-email.spec.ts
```diff
@@ -18,6 +18,7 @@ const emailSiweTest = test.extend<{ library: string }>({
 emailSiweTest.describe.configure({ mode: 'serial' })
 
 emailSiweTest.beforeAll(async ({ browser, library }, testInfo) => {
+  emailSiweTest.setTimeout(120000)
```
tomiir, line 21: is this not needed on email.spec as well?

chris13524: It's there

tomiir:
@chris13524 how are you dealing with the arbitrary delays here? You mean setting consistent timeouts?

chris13524:
> how are you dealing with the arbitrary delays here?

I just removed them. If there are consequences then we can add necessary compensation elsewhere but I don't see the need for this yet in my test runs.

tomiir (approved):

PR #2644, opened 2024-08-01
fix: add getApprovedCaipNetworks implementation

tomiir:
# Description
Solana client did not have a `getApprovedCaipNetworks` implementation and always defaulted to allowing all connections. 
Added  `getApprovedCaipNetworks` implementation for WC connections and for StandardWalletAdapter providers which have the available chain info.

## Type of change

# Associated Issues

For Linear issues: Closes APKT-769

# Checklist

### DIFF
--- a/packages/solana/src/client.ts
+++ b/packages/solana/src/client.ts
@@ -14,7 +14,7 @@ import { ConstantsUtil as CommonConstantsUtil } from '@web3modal/common'
 import { SolConstantsUtil, SolHelpersUtil, SolStoreUtil } from './utils/scaffold/index.js'
 import { WalletConnectConnector } from './connectors/walletConnectConnector.js'
 
-import type { BaseWalletAdapter, StandardWalletAdapter } from '@solana/wallet-adapter-base'
+import type { BaseWalletAdapter } from '@solana/wallet-adapter-base'
 import type { PublicKey, Commitment, ConnectionConfig } from '@solana/web3.js'
 import type UniversalProvider from '@walletconnect/universal-provider'
 import type {
@@ -32,6 +32,7 @@ import type { Chain as AvailableChain } from '@web3modal/common'
 
 import type { ProviderType, Chain, Provider, SolStoreUtilState } from './utils/scaffold/index.js'
 import { watchStandard } from './utils/wallet-standard/watchStandard.js'
+import { StandardWalletAdapter } from './utils/wallet-standard/adapter.js'
 
 export interface Web3ModalClientOptions extends Omit<LibraryOptions, 'defaultChain' | 'tokens'> {
   solanaConfig: ProviderType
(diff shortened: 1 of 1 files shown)


packages/solana/src/client.ts
```diff
@@ -110,13 +111,52 @@ export class Web3Modal extends Web3ModalScaffold {
...
+            if (!provider) {
+              throw new Error(
+                'networkControllerClient:getApprovedCaipNetworks - connector is undefined'
+              )
+            }
+            const ns = provider?.session?.namespaces
+            const nsMethods = ns?.['solana']?.methods
+            const nsChains = ns?.['solana']?.chains
+            const result = {
+              supportsAllNetworks: nsMethods?.includes('wallet_addSolanaChain') ?? false,
```
glitch-txs, line 127: This method doesn't exist in Solana, so this is always going to be false.

glitch-txs (changes requested):

packages/solana/src/client.ts
```diff
@@ -110,13 +111,51 @@ export class Web3Modal extends Web3ModalScaffold {
...
+              approvedCaipNetworkIds: nsChains as CaipNetworkId[] | undefined
+            }
 
-          resolve(result)
+            resolve(result)
+          } else {
+            const provider = SolStoreUtil.state.provider
+            if (provider && provider instanceof StandardWalletAdapter) {
+              const solanaNetworkNameToIdMap: Record<string, string> = {
+                mainnet: '5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp',
```
enesozturk, line 135: Can be used from `SolanaChainIDs` in `packages/solana/utils/chainPath/constants.ts`?

apps/laboratory/tests/wallet.spec.ts
```diff
@@ -39,8 +39,25 @@ sampleWalletTest.afterAll(async () => {
...
+sampleWalletTest.only(
+  'it should show testnet and devnet disabled on solana',
+  async ({ library }) => {
+    if (library !== 'solana') {
+      return
+    }
+
+    await modalPage.openModal()
+    await modalPage.openNetworks()
+    await modalValidator.expectNetworksDisabled('Solana Testnet')
```
enesozturk, line 51: Since we don't change something in the lab examples, why it's expected to be disabled?

packages/solana/src/client.ts
```diff
@@ -110,13 +111,51 @@ export class Web3Modal extends Web3ModalScaffold {
...
+            const provider = await this.WalletConnectConnector.getProvider()
+            if (!provider) {
+              throw new Error(
+                'networkControllerClient:getApprovedCaipNetworks - connector is undefined'
+              )
+            }
+            const ns = provider?.session?.namespaces
+            const nsChains = ns?.['solana']?.chains
+            const result = {
+              supportsAllNetworks: false,
```
enesozturk, line 126: Simple if the wallet is not connected with WC; we are setting `supportsAllNetworks` as `true`, and setting it as `true`  if connected with WC no matter what right? Addition to the question in testing, why the Testnet getting not supported? Confused a bit 🙏🏽

glitch-txs (approved):
Approving so it's not blocked. Need to test the extensions 🙏🏻

packages/solana/src/client.ts
tomiir: If connected via WC => depends on the namespaces of the session
If wallet adapter => depends on the wallet's `chains` 
else => there's nothing to compare, we allow all networks (default state on other clients as well)

apps/laboratory/tests/wallet.spec.ts
tomiir: Because since the wallet does not support it, it's not on namespace and should not be enabled

PR #2652, opened 2024-08-05
chore: Relay QR generation flow improvements

tomiir:
# Description

- WC Relay QR code generation has flaky behaviors.
- There is a lot of overhead in the connection setup: repeated event handlers, multiple connection requests that are stacked on top of each other, complex retry logic with two different timing conditions and poor error handling.
- This PR refactors the QR flow to bring some improvements:
  => Only subscribe to `display_uri` event once per client
  => Only re-initialize the connection flow after 15 minutes, given the sdk internally handles retries for that time
  => Always regenerate the URI on an error.

WIP: 
- Handling internal errors / network errors => Somehow we get no messages nor session updates from event handlers

## Type of change

# Associated Issues

For Linear issues: Closes APKT-144

# Checklist

- [] Code in this PR is covered by automated tests (Unit tests, E2E tests) 

### DIFF
--- a/packages/ethers/src/client.ts
+++ b/packages/ethers/src/client.ts
@@ -194,15 +194,11 @@ export class Web3Modal extends Web3ModalScaffold {
     }
 
     const connectionControllerClient: ConnectionControllerClient = {
-      connectWalletConnect: async onUri => {
+      connectWalletConnect: async () => {
         const WalletConnectProvider = await this.getWalletConnectProvider()
         if (!WalletConnectProvider) {
           throw new Error('connectionControllerClient:getWalletConnectUri - provider is undefined')
         }
-
-        WalletConnectProvider.on('display_uri', (uri: string) => {
-          onUri(uri)
-        })
 
         // When connecting through walletconnect, we need to set the clientId in the store
         const clientId = await WalletConnectProvider.signer?.client?.core?.crypto?.getClientId()
@@ -698,6 +694,10 @@ export class Web3Modal extends Web3ModalScaffold {
     }
 
     this.walletConnectProvider = await EthereumProvider.init(walletConnectProviderOptions)
+
+    this.walletConnectProvider.on('display_uri', (uri: string) => {
+      this.setQRCodeURI(uri)
+    })
 
     await this.checkActiveWalletConnectProvider()
   }
--- a/packages/scaffold-ui/src/views/w3m-connecting-wc-view/index.ts
+++ b/packages/scaffold-ui/src/views/w3m-connecting-wc-view/index.ts
@@ -3,7 +3,6 @@ import {
   AssetUtil,
   ConnectionController,
   ConnectorController,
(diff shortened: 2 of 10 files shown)


packages/ethers/src/client.ts
```diff
@@ -699,6 +695,10 @@ export class Web3Modal extends Web3ModalScaffold {
 
     this.walletConnectProvider = await EthereumProvider.init(walletConnectProviderOptions)
 
+    this.walletConnectProvider.on('display_uri', (uri: string) => {
```
glitch-txs, line 698: minor but you could also

```ts
this.walletConnectProvider.on('display_uri', this.setQRCodeURI)
```

packages/scaffold/src/client.ts
```diff
@@ -327,6 +327,10 @@ export class Web3ModalScaffold {
     BlockchainApiController.setClientId(clientId)
   }
 
+  protected setQRCodeURI: (typeof ConnectionController)['setQRCodeURI'] = uri => {
```
ignaciosantise, line 330: the uri is not only for the QR code but for all connections using WalletConnect protocol. maybe `setURI` or `setWalletConnectURI` works better?

packages/solana/src/connectors/walletConnectConnector.ts
```diff
@@ -7,7 +7,7 @@ import { UniversalProviderFactory } from './universalProvider.js'
 import { BaseConnector } from './baseConnector.js'
 
 import type { Signer } from '@solana/web3.js'
-import type UniversalProvider from '@walletconnect/universal-provider'
```
ignaciosantise, line 10: why this?

glitch-txs: This actually breaks in some bundlers (default import of UP), it should be a name import.

packages/scaffold-ui/src/views/w3m-connecting-wc-view/index.ts
```diff
@@ -33,7 +30,11 @@ export class W3mConnectingWcView extends LitElement {
   public constructor() {
     super()
     this.initializeConnection()
-    this.interval = setInterval(this.initializeConnection.bind(this), ConstantsUtil.TEN_SEC_MS)
+    /*
+     * For 15minutes, the relay handles retries and reconnections internally.
+     * We re-initialize the connection after that
+     */
+    this.interval = setInterval(this.initializeConnection.bind(this), 15 * 60 * 1000)
```
chris13524, line 37: This is the Sign SDK that handles this, and I would be concerned that depending on this behavior is dangerous. E.g. if the SDK is updated to a version w/o this

packages/solana/src/connectors/walletConnectConnector.ts
chris13524: @glitch-txs can we make a lint or tsconfig role to disallow this?

PR #2689, opened 2024-08-09
fix(multi-account): account switch on wagmi

enesozturk:
# Description

Our Wagmi client have a connection status mismatching issue while switching between accounts and re-calling the `syncAccount` method. The `status` value is getting `undefined`, thus, the new account is not properly set.

This PR introduces changes to fix this along with some function props for multi-chain functionality. The multi-chain prop update is not fixing something, it just makes sure that we are okay when we switch to multi-chain architecture.

## Type of change

# Associated Issues

For GH issues: closes #...

# Showcase (Optional)

# Checklist


packages/ethers/src/client.ts
```diff
@@ -836,7 +839,10 @@ export class Web3Modal extends Web3ModalScaffold {
         EthersStoreUtil.setProvider(config.injected)
         EthersStoreUtil.setStatus('connected')
         EthersStoreUtil.setIsConnected(true)
-        this.setAllAccounts(addresses.map(address => ({ address, type: 'eoa' })))
+        this.setAllAccounts(
+          addresses.map(address => ({ address, type: 'eoa' })),
+          this.chain
```
enesozturk, line 844: These are just to make sure we are passing the active chain to the scaffold, this will help us when we switch to multi-chain architecture.

packages/wagmi/src/client.ts
```diff
@@ -460,7 +462,8 @@ export class Web3Modal extends Web3ModalScaffold {
     chainId,
     connector,
     addresses,
-    status
+    status,
+    resetAccount = true
```
enesozturk, line 466: Added optional `resetAccount` param - this is something we could discuss. Do you think we should reset the account state when switch to another account? If you see `ChainController.resetAccount`, we are resetting bunch of things and if we call `this.resetAccount` before setting the new account, this is first rendering the unconnected `w3m-button`, then shows account button. Like there is a glitch while switching accounts. I didn't want to have that.

@tomiir @ganchoradkov like to see your thoughts

packages/wagmi/src/client.ts
```diff
@@ -498,15 +503,20 @@ export class Web3Modal extends Web3ModalScaffold {
...
       this.hasSyncedConnectedAccount = true
     } else if (isDisconnected && this.hasSyncedConnectedAccount) {
-      this.resetAccount()
+      if (resetAccount) {
+        this.resetAccount()
+      }
       this.resetWcConnection()
       this.resetNetwork()
-      this.setAllAccounts([])
+      this.setAllAccounts([], this.chain)
```
enesozturk, line 519: Same with the ethers. Passing adapter's chain info to the scaffold.

tomiir (approved):

packages/wagmi/src/client.ts
tomiir: Im wondering if we really need to reset the account. Maybe we should more carefully update values instead

enesozturk: I agree with this tbh. There are bunch of values for the account state like the followings. We just need to be double sure that not resetting them wouldn't cause issues on different features. Tried to list them like that. Will make update accordingly:

```js
isConnected: false, // OK - will be updated
smartAccountDeployed: false, // in email/social we don't have multi-account yet, so should be OK to not reset
currentTab: 0, // OK
caipAddress: undefined, // OK - will be updated
address: undefined, // OK - will be updated
balance: undefined, // OK - will be updated
balanceSymbol: undefined, // OK - will be updated
profileName: undefined, // OK - will be updated
profileImage: undefined, // OK - will be updated
addressExplorerUrl: undefined, // OK - will be updated 
tokenBalance: [],  // related with Swaps, need to reset
connectedWalletInfo: undefined, // OK - will be updated 
preferredAccountType: undefined, // in email/social we don't have multi-account yet, so should be OK to not reset
socialProvider: undefined, // in email/social we don't have multi-account yet, so should be OK to not reset
socialWindow: undefined, // in email/social we don't have multi-account yet, so should be OK to not reset
farcasterUrl: undefined // in email/social we don't have multi-account yet, so should be OK to not reset
```

But another issue is that `syncAccount` is called for different purposes, the calls other than account change might need to reset all?

enesozturk: Refactored the ,syncAccount` and removed the unnecessary `resetAccount` and `status` prop which is doing the same with the `isConnected` and `isDisconnected` , see: https://wagmi.sh/core/api/actions/getAccount#isconnecting-isreconnecting-isconnected-isdisconnected

Tian-Hun:
Account switching is failing to function correctly in both my local @web3modal/wagmi 5.1.6 setup and the [web3modal-laboratory](https://lab.web3modal.com/library/wagmi-no-email) online testing platform.

PR #2690, opened 2024-08-09
refactor: standardize solana provider adapters

zoruka:
# Description

This PR brings a refactor for Solana client by adapting `WalletConnectConnector` and `StandardWalletAdapter` classes into the `Provider` interface. 

## The differences

- By using the Provider interface we are allowed to remove the different treatment of adapters and connectors we have being doing in the client and have the diferences between the connection methods treated by each Provider exclusively;
- The client should treat the state handling through the Events emitted by the Providers and we must enforce that each Provider have the correct event triggers being called ensuring the correct behavior of the client;
- The implementation of each Provider is based over composition pattern, which allows us to create unit test suits for the providers that validates if the Provider execution is correct;
- We are allowed to add more Providers easily without needing to modify the behavior the client code, this structure should be used to easily add the AuthProvider (email and socials).

This PR is also removing code that was not in use anymore.

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# Associated Issues

For GH issues: closes #...

# Showcase (Optional)

# Checklist


packages/solana/src/providers/WalletConnectProvider.ts
```diff
@@ -0,0 +1,254 @@
...
+
+export type WalletConnectProviderConfig = {
+  provider: UniversalProvider
+  chains: Chain[]
+}
+
+export class WalletConnectProvider extends ProviderEventEmitter implements Provider {
+  public readonly name = 'WalletConnect'
+  public readonly type = 'WALLET_CONNECT'
+  public readonly icon = 'https://avatars.githubusercontent.com/u/37784886'
```
glitch-txs, line 23: Do we fetch this from Cloud? @enesozturk

packages/solana/src/providers/WalletConnectProvider.ts
```diff
@@ -0,0 +1,254 @@
...
+      this.session = await this.provider.connect({
+        optionalNamespaces: {
+          solana: {
+            chains: this.getRequestedChainsWithDeprecated(),
+            methods: [
+              'solana_signMessage',
+              'solana_signTransaction',
+              'solana_signAndSendTransaction'
+            ],
+            events: [],
```
glitch-txs, line 88: Should we add the accountChanged event here?

packages/solana/src/providers/WalletConnectProvider.ts
```diff
@@ -0,0 +1,254 @@
...
+    }
+
+    return Transaction.from(decodedTransaction) as T
+  }
+
+  public async signAndSendTransaction<T extends AnyTransaction>(
+    transaction: T,
+    sendOptions?: SendOptions
+  ) {
+    const serializedTransaction = this.serializeTransaction(transaction)
```
glitch-txs, line 138: Probably we would like to check if this method was approved in the session before calling it? To make sure the wallet does support it.

glitch-txs: We could fallback on sign maybe? I'm not sure...

packages/solana/src/providers/WalletConnectProvider.ts
zoruka: I have tried to add the event but then it doesn't connect, as main code also have this issue I decided to not tackle it in the refactor as may increase even more the PR size.

packages/solana/src/providers/WalletConnectProvider.ts
zoruka: Is there a way to check if the wallet support the feature through the UniversalProvider?

packages/solana/src/providers/WalletConnectProvider.ts
zoruka: This was already in use before https://github.com/WalletConnect/web3modal/blob/main/packages/solana/src/client.ts#L421, but we may change it if is better.

apps/laboratory/src/components/Solana/SolanaWriteContractTest.tsx
```diff
@@ -26,7 +26,7 @@ export function SolanaWriteContractTest() {
     const PROGRAM_ID = new PublicKey(detectProgramId(currentChain?.chainId ?? ''))
 
     try {
-      if (!walletProvider || !address) {
+      if (!walletProvider?.publicKey) {
```
enesozturk, line 29: @zoruka Is this the only way to check if user is connected / have address? Would this be something unified for other chains as well? (AFAIK the `publicKey` is a concept only on Solana)

packages/solana/exports/chains.ts
```diff
@@ -1 +0,0 @@
-export { solana, solanaDevnet, solanaTestnet } from '../src/utils/chains.js'
```
enesozturk, line 1: Why don't we export them here anymore?

packages/solana/src/providers/WalletConnectProvider.ts
enesozturk: @zoruka Can we use this one: https://imagedelivery.net/_aTEfDRm7z3tKgu9JhfeKA/05338e12-4f75-4982-4e8a-83c67b826b00/md (from our CDN service)

packages/solana/tests/WalletStandardProvider.test.ts
```diff
@@ -0,0 +1,124 @@
...
+  SolanaSignTransaction
+} from '@solana/wallet-standard-features'
+import { TestConstants } from './util/TestConstants'
+import { mockLegacyTransaction, mockVersionedTransaction } from './mocks/Transaction'
+import { WalletStandardFeatureNotSupportedError } from '../src/providers/shared/Errors'
+
+describe('WalletStandardProvider specific tests', () => {
+  let wallet = mockWalletStandard()
+  let getActiveChain = vi.fn(() => TestConstants.chains[0])
+  let sut = new WalletStandardProvider({
```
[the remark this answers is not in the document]
zoruka: sut means "system under test", do you suggest another name, or just `walletStandardProvider`?

packages/solana/src/utils/chains.ts
```diff
@@ -26,3 +27,9 @@ export const solanaDevnet = {
   rpcUrl: 'https://rpc.walletconnect.org/v1',
   chain: ConstantsUtil.CHAIN.SOLANA
 }
+
+export const solanaChains = {
+  'solana:mainnet': solana,
+  'solana:testnet': solanaTestnet,
+  'solana:devnet': solanaDevnet
+} as Record<`${string}:${string}`, Chain>
```
enesozturk, line 35: Maybe this?
```suggestion
} as Record<`solana:${"mainnet"|"testnet"|"devnet"}`, Chain>
```

apps/laboratory/src/components/Solana/SolanaWriteContractTest.tsx
zoruka: This is related to type assertion for `publicKey` , not exactly to check if is connected or not. To check the connection status we should use the `isConnected` method from `useWeb3ModalAccount`.

packages/solana/tests/WalletStandardProvider.test.ts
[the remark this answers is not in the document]
enesozturk: I didn't know about system under test approach which sounds a good name, I'd prefer `walletStandardProvider` which is more understandable for the team IMO

packages/solana/src/utils/chains.ts
zoruka: This is going to break type assertions forward, maybe we don't need to have as strict?

apps/laboratory/src/components/Solana/SolanaWriteContractTest.tsx
zoruka: Btw, I have removed the address checking because it was not in use besides this conditional.

packages/solana/tests/WalletStandardProvider.test.ts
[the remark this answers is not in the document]
zoruka: let me update it

packages/solana/src/utils/chains.ts
enesozturk: Got it, that's cool then

enesozturk (approved):

packages/solana/src/providers/WalletConnectProvider.ts
glitch-txs: yes, `provider.session.namespaces` something like that

PR #2704, opened 2024-08-15
fix: solana qa round

zoruka:
# Description

This PR brings somes fixes after a QA round of tests:
- Correct handling of available chains by wallet;
- Improved parsing of deprecated chains;
- Fix for `SolanaWriteContractTest`;
- Bring back previous base64 encoding of the transaction as described in our RPC specs.

## Type of change

# Associated Issues

For GH issues: closes #...

# Showcase (Optional)

# Checklist

### DIFF
--- a/apps/laboratory/src/components/Solana/SolanaWriteContractTest.tsx
+++ b/apps/laboratory/src/components/Solana/SolanaWriteContractTest.tsx
@@ -4,9 +4,10 @@ import {
   SystemProgram,
   PublicKey,
   Keypair,
-  Transaction,
   TransactionInstruction,
-  LAMPORTS_PER_SOL
+  LAMPORTS_PER_SOL,
+  VersionedTransaction,
+  TransactionMessage
 } from '@solana/web3.js'
 import { useWeb3ModalAccount, useWeb3ModalProvider } from '@web3modal/solana/react'
 
@@ -42,7 +43,7 @@ export function SolanaWriteContractTest() {
         throw Error('Not enough SOL in wallet')
       }
 
-      const allocIx: TransactionInstruction = SystemProgram.createAccount({
+      const allocIx = SystemProgram.createAccount({
         fromPubkey: walletProvider.publicKey,
         newAccountPubkey: counter,
         lamports: await connection.getMinimumBalanceForRentExemption(COUNTER_ACCOUNT_SIZE),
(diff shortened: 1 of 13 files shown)


apps/laboratory/src/components/Solana/SolanaWriteContractTest.tsx
```diff
@@ -62,11 +63,33 @@ export function SolanaWriteContractTest() {
...
+      )
 
-      await walletProvider.signAndSendTransaction(tx)
+      transaction.sign([counterKeypair])
+
+      const transactionId = await walletProvider.sendTransaction(transaction, connection, {
+        preflightCommitment: 'confirmed'
+      })
+
+      // We should consider using `connection.confirmTransaction` instead of polling when websocket support is added
```
enesozturk, line 82: @zoruka Can we track this improvement with a Linear ticket?

packages/solana/src/providers/WalletConnectProvider.ts
```diff
@@ -179,14 +182,46 @@ export class WalletConnectProvider extends ProviderEventEmitter implements Provi
...
+    return this.session?.namespaces['solana']?.chains || []
   }
 
   private serializeTransaction(transaction: AnyTransaction) {
-    return base58.encode(transaction.serialize({ verifySignatures: false }))
+    /*
+     * We should consider serializing the transaction to base58 as it is the solana standard.
+     * But our specs requires base64 right now:
+     * https://docs.walletconnect.com/advanced/multichain/rpc-reference/solana-rpc#solana_signtransaction
+     */
```
enesozturk, line 223: Same here for improvement ticket 🙏🏽

packages/solana/src/client.ts
```diff
@@ -390,36 +385,47 @@ export class Web3Modal extends Web3ModalScaffold {
...
-    SolStoreUtil.setProvider(provider)
-    this.setAddress(address)
+      // Check if the provider supports the current chain or switch to the first supported chain
+      const connectionChain =
+        provider.chains.find(chain => chain.chainId === SolStoreUtil.state.currentChain?.chainId) ||
+        provider.chains[0]
 
-    window?.localStorage.setItem(SolConstantsUtil.WALLET_ID, provider.name)
+      if (connectionChain) {
+        await this.switchNetwork(
```
enesozturk, line 398: Why do we do this? Would this be unexpected UX for the users? Is this have some bad cases (ie. like not having a chain provided)?

enesozturk (approved):
@zoruka Left some comments

packages/solana/src/client.ts
zoruka: Because the wallet may connect and nothing works because the wallet doesn't support the chain, it is the case of connecting with Trust Wallet and a couple other wallets that only supports mainnet. Then we are just switch to any of the supported chains or stop the connection.

PR #2709, opened 2024-08-16
feat: solana auth provider

zoruka:
# Description

This PR adds the Solana `AuthProvider` and setup wallet features for secure site. It goes alongside with https://github.com/WalletConnect/secure-web3modal/pull/163.

Edit:

The Solana's Auth feature will be disabled until secure site update is ready.

## Type of change

# Associated Issues

For Linear issues: Closes APKT-723, Closes APKT-715

# Showcase (Optional)

https://walletconnect.slack.com/files/U07BR253G1M/F07JF002B5W/example.mov

# Checklist


packages/core/src/controllers/ConnectorController.ts
```diff
@@ -104,5 +89,28 @@ export const ConnectorController = {
...
+  },
+
+  syncIfAuthConnector(connector: Connector | AuthConnector) {
+    if (connector.id !== 'w3mAuth') {
+      return
+    }
+
+    const authConnector = connector as AuthConnector
+
+    const optionsState = snapshot(OptionsController.state) as typeof OptionsController.state
```
tomiir, line 101: nit: why not `getSnapshot` as the other store?

packages/solana/src/client.ts
```diff
@@ -281,19 +290,17 @@ export class Web3Modal extends Web3ModalScaffold {
   private async syncAccount() {
     const address = SolStoreUtil.state.address
     const chainId = SolStoreUtil.state.currentChain?.chainId
-    const isConnected = SolStoreUtil.state.isConnected
-    this.resetAccount()
 
-    if (isConnected && address && chainId) {
+    if (address && chainId) {
```
tomiir, line 294: is this ok? why remove the isConnected check?

packages/solana/src/client.ts
```diff
@@ -332,6 +339,11 @@ export class Web3Modal extends Web3ModalScaffold {
 
   public async switchNetwork(caipNetwork: CaipNetwork) {
     const caipChainId = caipNetwork.id
+
+    if (this.provider instanceof AuthProvider) {
+      await this.provider.switchNetwork(caipChainId)
+    }
```
tomiir, line 345: is this required? can't we init the provider with a network directly?

packages/solana/src/client.ts
```diff
@@ -365,7 +376,8 @@ export class Web3Modal extends Web3ModalScaffold {
           imageUrl: chainImages?.[chain.chainId],
           chain: this.chain
         })
-        if (isConnected && address) {
+
+        if (address) {
```
tomiir, line 380: same, why remove the connected check?

packages/solana/src/client.ts
```diff
@@ -394,38 +406,98 @@ export class Web3Modal extends Web3ModalScaffold {
...
       this.setLoading(false)
     }
   }
 
   private watchProvider(provider: Provider) {
+    /*
+     * The auth RPC request handlers should be moved to the primary scaffold (Web3ModalScaffold).
+     * They are replicated in wagmi and ethers clients and the behavior should be kept the same
+     * between any client.
+     */
```
tomiir, line 437: Agreed. @enesozturk @svenvoskamp we should apply this in MC directly I'd think (`AppKit` class)

packages/solana/src/client.ts
```diff
@@ -394,38 +406,98 @@ export class Web3Modal extends Web3ModalScaffold {
...
+      } else {
+        super.open()
+        // eslint-disable-next-line no-console
+        console.error(W3mFrameRpcConstants.RPC_METHOD_NOT_ALLOWED_MESSAGE, {
+          method: request.method
+        })
+        setTimeout(() => {
+          this.showErrorMessage(W3mFrameRpcConstants.RPC_METHOD_NOT_ALLOWED_UI_MESSAGE)
+        }, 300)
+        // Provider.rejectRpcRequests()
```
tomiir, line 465: remove or uncomment? I think there's no need to cancel ALL requests at this point

packages/solana/src/client.ts
```diff
@@ -394,38 +406,98 @@ export class Web3Modal extends Web3ModalScaffold {
...
+    // eslint-disable-next-line func-style
+    const rpcSuccessHandler = (_response: W3mFrameTypes.FrameEvent) => {
+      if (super.isTransactionStackEmpty()) {
+        super.close()
+      } else {
+        super.popTransactionStack()
+      }
+    }
+
+    // eslint-disable-next-line func-style
```
tomiir, line 478: any reason to disable linter?

packages/solana/src/client.ts
```diff
@@ -441,6 +513,9 @@ export class Web3Modal extends Web3ModalScaffold {
     provider.on('disconnect', disconnectHandler)
     provider.on('accountsChanged', accountsChangedHandler)
     provider.on('connect', accountsChangedHandler)
+    provider.on('auth_rpcRequest', rpcRequestHandler)
+    provider.on('auth_rpcSuccess', rpcSuccessHandler)
+    provider.on('auth_rpcError', rpcErrorHandler)
```
tomiir, line 518: where are you emmiting these? why did you end up doing this here? 
do we need to update other clients?

packages/solana/src/client.ts
zoruka: It is redundant with `address` and it blocking to sync account while inside the `connect` function that only sets `isConnected` when the full connection is done

packages/solana/src/providers/AuthProvider.ts
```diff
@@ -0,0 +1,203 @@
...
+    const result = await this.provider.switchNetwork(args)
+    this.emit('chainChanged', args as string)
+
+    return result
+  }
+
+  // -- Private ------------------------------------------- //
+  private getPublicKey<Required extends boolean>(
+    required?: Required
+  ): Required extends true ? PublicKey : PublicKey | undefined {
```
tomiir, line 161: what is this response type?

packages/wallet/src/W3mFrameSchema.ts
```diff
@@ -310,6 +310,48 @@ export const RpcEthSendTransactionRequest = z.object({
...
+  params: z.object({
+    transaction: z.string()
+  })
+})
+
+export const RpcSolanaSignAndSendTransactionRequest = z.object({
+  method: z.literal('solana_signAndSendTransaction'),
+  params: z.object({
+    transaction: z.string(),
+    // Options should match https://solana-labs.github.io/solana-web3.js/types/SendOptions.html
```
tomiir, line 332: any reason to base it off solana-web3js specfically?

packages/solana/src/client.ts
zoruka: but what if the user changes the network? the secure site will remain with old network

packages/solana/src/client.ts
zoruka: It has redundancy with `address` and blocks `connect` function for setting `isConnected` only after everything is ok in the connection. Do you think this is not the correct behavior?

packages/solana/src/client.ts
zoruka: without it would be required to declare the function with `function name() {` which trigger ts errors about `this` and `super` as it loses the binding scope.

packages/solana/src/client.ts
zoruka: in this structure all events are emitted by the providers and the client behavior is handled through these events:

- https://github.com/WalletConnect/web3modal/pull/2709/files/d6f1b9ac0a71d95a6eab05f7181fa597bdccadf2#diff-7292affc7810dedb06b2c96d9e1b0552120ec1f525697b37f93cccb8fafdb2c9R177-R199
- https://github.com/WalletConnect/web3modal/pull/2709/files/d6f1b9ac0a71d95a6eab05f7181fa597bdccadf2#diff-7292affc7810dedb06b2c96d9e1b0552120ec1f525697b37f93cccb8fafdb2c9R153
- https://github.com/WalletConnect/web3modal/pull/2709/files/d6f1b9ac0a71d95a6eab05f7181fa597bdccadf2#diff-7292affc7810dedb06b2c96d9e1b0552120ec1f525697b37f93cccb8fafdb2c9R71
- https://github.com/WalletConnect/web3modal/pull/2709/files/d6f1b9ac0a71d95a6eab05f7181fa597bdccadf2#diff-7292affc7810dedb06b2c96d9e1b0552120ec1f525697b37f93cccb8fafdb2c9R79

The idea is splitting client behavior from provider execution, so we have less replicated code for injected wallet, wallet connect, auth and so on.

packages/solana/src/providers/AuthProvider.ts
zoruka: if `required` is true then the function must return the `PublicKey` type, else it can return `PublicKey | undefined`

packages/wallet/src/W3mFrameSchema.ts
zoruka: It is what exposes the exact options for sending a transaction to Solana and what is commonly used for interact with Solana. Do you think we should change it?

packages/solana/exports/options.ts
```diff
@@ -0,0 +1,5 @@
+import type { AppKitOptions } from '@web3modal/base'
```
enesozturk, line 1: Why we needed to export these things under `*/options`?

packages/core/src/controllers/ConnectorController.ts
```diff
@@ -31,6 +31,7 @@ export const ConnectorController = {
   },
 
   setConnectors(connectors: ConnectorControllerState['connectors']) {
+    connectors.forEach(this.syncIfAuthConnector)
```
enesozturk, line 34: Why do we need to do this? When we switch to multiple adapters, this will run the `syncIfAuthConnector` function for every connector we have. Is it okay? What about if the `wagmi/ethers/ethers5`'s auth connector? It'll pass the if clause right?ç

tomiir (approved):

enesozturk (approved):

PR #2715, opened 2024-08-19
fix: remove auth connector from ethers and check for socials length

tomiir:
# Description
Auth connector was being injected in ethers even when setting `email: false`

## Type of change

# Associated Issues

For Linear issues: Closes APKT-874

# Checklist

- [] Code in this PR is covered by automated tests (Unit tests, E2E tests)


packages/wagmi/src/utils/defaultWagmiCoreConfig.ts
```diff
@@ -76,7 +76,7 @@ export function defaultWagmiConfig({
     ...auth
   }
 
-  if (mergedAuth.email || mergedAuth.socials) {
+  if (mergedAuth.email || mergedAuth.socials?.length) {
```
enesozturk, line 79: Why don't we allow users to have email login if they don't set social options? Cannot they use only email?

enesozturk (approved):

svenvoskamp (approved):

apps/laboratory/src/utils/DataUtil.ts
```diff
@@ -61,6 +66,11 @@ export const ethersSdkOptions = [
     title: 'Email and Social',
     link: '/library/ethers-email/',
     description: 'Configuration using ethers and implementing email and social login'
+  },
+  {
```
enesozturk, line 70: What about `ethers5` no email?

enesozturk: We could even test solana since it's enabled for Solana as well

tomiir: Removing ethers5 email until we get it working properly

Let me change this for solana

tomiir: It doesn't really have a working email tbh. LEaving it out until we have it stable. 
Adding solana too 🫡

PR #2734, opened 2024-08-23
refactor/separation of dependencies

enesozturk:
# Description

### Unnecessary dependencies

Moving adapter specific dependencies to `devDependencies`. We are maintaining multiple adapters under `base` package and this approach require us to install different clients (wagmi, ethers, ethers5, solana packages). To keep the dependencies separated and not include them in the base package, we need to move them to `devDependencies`.

### Module Resolution for types

`typesVersions` field allows you to specify different type definitions based on the TypeScript version or based on the module resolution context.

When you define path exports in your `package.json`, it means you are providing specific entry points for your package’s modules. This approach is helpful for optimizing bundling and tree-shaking in modern bundlers. However, TypeScript needs to know where to find the type definitions for these specific paths.

## Type of change

# Associated Issues

For GH issues: closes #...

# Showcase (Optional)

# Checklist


Cali93 (approved):
LGTM 💪

svenvoskamp (approved):

packages/base/package.json
```diff
@@ -50,21 +57,57 @@
...
     }
   },
-  "scripts": {
-    "build:clean": "rm -rf dist",
-    "build": "tsc --build tsconfig.build.json",
-    "watch": "tsc --watch",
-    "typecheck": "tsc --noEmit",
-    "lint": "eslint . --ext .js,.jsx,.ts,.tsx"
+  "typesVersions": {
+    "*": {
```
tomiir, line 61: What is this sub level?

enesozturk: https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#version-selection-with-typesversions

The "*" is acting as a wildcard which means these type definitions apply to all versions of the package. Since we don't need to separate them between versions, we are okay to go with this way.

Following the approach we are doing in `packages/wagmi/package.json`, the following would work as well I think:

```js
"typesVersions": {
  "react": [
    "./dist/types/exports/react/index.d.ts"
  ],
  "vue": [
    "./dist/types/exports/vue.d.ts"
  ]
}
```

PR #2775, opened 2024-08-30
fix: update CSP

chris13524:
- Updates the CSP to enforce
- Removes `script-src unsafe-inline` as this is totally unsafe
- ~~Removes `.com` API endpoints to enforce the use of the `.org` endpoints~~
  - Will do in follow-up PR as looks like we are still using .com
- Adds `*` to `img-src` to allow loading ENS avatar images from any URL


apps/laboratory/next.config.mjs
```diff
@@ -1,13 +1,18 @@
 /** @type {import('next').NextConfig} */
-// Keep in-sync with https://docs.walletconnect.com/advanced/security/content-security-policy
+const SHAKRA_UI = `'sha256-e7MRMmTzLsLQvIy1iizO1lXf7VWYoQ6ysj5fuUzvRwE='`
+/*
+ * Keep in-sync with https://docs.walletconnect.com/advanced/security/content-security-policy
```
tomiir, line 4: what happens if there's a mismatch

chris13524: We risk having docs that aren't accurate or CSP requirements that aren't tested

enesozturk (approved):

PR #2778, opened 2024-09-01
fix/missing action functions exports from clients

enesozturk:
# Description

While migrating to the multi-chain architecture, there are few getter, action and listeners methods that the new modal instances don't expose but was available for the previous clients. This PR introduces changes with the required type definition tricks to have type safe method expose. 

The reason we needed to use generics for AppKit and `ChainAdapter` types is that each client's following methods have different type of parameters and returns. When users use `@web3modal/ethers` or `@web3modal/solana`, the TS should know which type it'll take an return for these methods. For that to resolve, we need to use generics while exporting the backwards compatibility support.

```
getError
getChainId
getAddress
switchNetwork
getIsConnected
getWalletProvider
getWalletProviderType
subscribeProvider
```

This PR also introduces:
- Few changes to example apps, adds new example app for Ethers to test those methods.
- Fixing multi-address switching issue on Wagmi

## Type of change

# Associated Issues

For GH issues: closes #...

# Showcase (Optional)

# Checklist


svenvoskamp (approved):

packages/base/src/client.ts
```diff
@@ -303,7 +344,11 @@ export class AppKit {
     AssetUtil.getConnectorImage(connector)
 
   // -- Private ------------------------------------------------------------------
-  private async initControllers(options: AppKitOptions) {
+  private async initControllers(
+    options: AppKitOptions & {
```
tomiir, line 348: Why remove adapters from type and add it manually here?

tomiir (approved):

packages/base/src/client.ts
enesozturk: Bc we need to pass different types to the `adapters` array since the v5 is only have single adapter. And that's how we are fixing the missing exposed functions that I explained in the PR description. So our backwards compatible versions will be sharing same types between adapter <-> base.

0xBURP:
but AppKit.subscribeProvider is still unavailable in @web3modal/ethers 5.1.7?

PR #2788, opened 2024-09-03
chore: run Playwright on self-hosted runners

chris13524:
# Description

Runs Playwright tests on our self-hosted runners to attempt to reduce flakiness. GitHub runners have network flakiness that tests/SDK currently does not gracefully handle.

## Type of change


tomiir (approved):
Let's try

PR #2831, opened 2024-09-10
chore: test ethers verify API + AppKit in iframe verify test case

chris13524:
# Description

Adds missing test coverage for Verify API on the Ethers SDK.

Also adds test coverage for https://github.com/WalletConnect/cloud-app/pull/1509

Remaining work:

- [ ] Add parameter so Cloud App's execution of these tests can specify a custom Verify Server: https://github.com/WalletConnect/cloud-app/pull/1509

Fixes APKT-1445

## Type of change


apps/laboratory/tests/shared/fixtures/w3m-fixture.ts
```diff
@@ -30,6 +30,18 @@ export const testM = timingFixture.extend<ModalFixture>({
     await use(modalPage)
   }
 })
+export const testMEthers = timingFixture.extend<ModalFixture>({
```
tomiir, line 33: How is this different fromcthe default testM? / why does it need a special case?

tomiir (approved):

apps/laboratory/next.config.mjs
```diff
@@ -8,18 +8,20 @@ const SHAKRA_UI = `'sha256-e7MRMmTzLsLQvIy1iizO1lXf7VWYoQ6ysj5fuUzvRwE='`
...
   font-src 'self' https://fonts.gstatic.com;
   connect-src 'self' https://react-wallet.walletconnect.com https://rpc.walletconnect.com https://rpc.walletconnect.org https://relay.walletconnect.com https://relay.walletconnect.org wss://relay.walletconnect.com wss://relay.walletconnect.org https://pulse.walletconnect.com https://pulse.walletconnect.org https://api.web3modal.com https://api.web3modal.org wss://www.walletlink.org https://o1095249.ingest.sentry.io https://quote-api.jup.ag;
-  frame-src 'self' https://verify.walletconnect.com https://verify.walletconnect.org https://secure.walletconnect.com https://secure.walletconnect.org https://secure.reown.com https://widget.solflare.com/ ${secureSiteDomain}/;
+  frame-src 'self' https://verify.walletconnect.com https://verify.walletconnect.org https://secure.walletconnect.com https://secure.walletconnect.org https://secure.reown.com https://widget.solflare.com/ ${secureSiteDomain};
   object-src 'none';
   base-uri 'self';
   form-action 'self';
-  frame-ancestors 'none';
+  frame-ancestors ${verifyApiNestedIframesTestOuterDomain};
```
tomiir, line 24: do we need to conditionally add this on test env?

chris13524: iirc adding it conditionally was tricky because the tests run against `NODE_ENV` production, so it didn't know if it was building for tests or not

Didn't bother going deeper I don't think, so hardcoded an intentionally-long domain here which wouldn't be valid anyway

PR #2861, opened 2024-09-17
refactor: local storage keys to appkit + network switch fixes

enesozturk:
# Description

Switched to `@appkit/*` prefix for the local storage keys.

## Type of change

# Associated Issues

For GH issues: closes #...

# Showcase (Optional)

# Checklist


packages/common/src/utils/SafeLocalStorage.ts
```diff
@@ -24,7 +32,7 @@ export const SafeLocalStorage = {
     value: SafeLocalStorageItems[Key]
   ): void {
     if (isSafe()) {
-      localStorage.setItem(key, JSON.stringify(value))
+      localStorage.setItem(key, value)
```
enesozturk, line 35: We shouldn't do the parsing here. An example;

We have `@appkit/active_caip_network` key, which is expected to store `CaipNetwork` type. 

If we set `'@appkit/active_caip_network': string` on `SafeLocalStorageItems`, the `getItem`'s `catch` will return type error bc, if somehow the parsing fails, it'll return `string` which is not what that key stores.

So the `stringify`ing and `parse`ing should be abstracted from this `SafeLocalStorage` utility.

packages/adapters/wagmi/src/client.ts
```diff
@@ -230,10 +230,6 @@ export class WagmiAdapter implements ChainAdapter {
 
     this.networkControllerClient = {
       switchCaipNetwork: async caipNetwork => {
-        SafeLocalStorage.setItem(
-          SafeLocalStorageKeys.ACTIVE_CAIP_NETWORK,
-          JSON.stringify(caipNetwork)
-        )
```
tomiir, line 236: is this intended if so, where are we persisting the netowkr?

packages/adapters/solana/src/client.ts
```diff
@@ -140,10 +140,6 @@ export class SolanaAdapter implements ChainAdapter {
     this.networkControllerClient = {
       switchCaipNetwork: async caipNetwork => {
         if (caipNetwork) {
-          SafeLocalStorage.setItem(
-            SafeLocalStorageKeys.ACTIVE_CAIP_NETWORK,
-            JSON.stringify(caipNetwork)
-          )
```
tomiir, line 146: same

packages/adapters/wagmi/src/client.ts
enesozturk: We are persisting the network on where we update the state, we shouldn't be calling it from different places. 

Like if you do;
```js
// pseudo code
ChainController.state.activeCaipNetwork = something
```

You can update localStroge here as well
```js
// pseudo code
ChainController.state.activeCaipNetwork = something
localStorage.setItem("activeCaipNetwork, something) 
```

packages/common/src/utils/SafeLocalStorage.ts
```diff
@@ -24,23 +34,29 @@ export const SafeLocalStorage = {
...
+  ): SafeLocalStorageItems[Key] | undefined {
     if (isSafe()) {
       const value = localStorage.getItem(key)
 
       if (value) {
         try {
           return JSON.parse(value)
         } catch {
-          return value
+          return undefined
```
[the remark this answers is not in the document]
enesozturk: Nope, checked everthing we read. We are trying to use undefined for falsy values instead of `null`. I remember we've been discussed this long ago (maybe last year). So for such values we should avoid `null` and use `undefined` (not sure what is the strong why)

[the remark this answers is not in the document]
tomiir: I prefer to have `undefined` personally as it prevents you from doing `| null` everywhere else so ok for me

packages/common/src/utils/SafeLocalStorage.ts
```diff
@@ -24,23 +34,29 @@ export const SafeLocalStorage = {
     value: SafeLocalStorageItems[Key]
   ): void {
     if (isSafe()) {
-      localStorage.setItem(key, JSON.stringify(value))
+      if (typeof value === 'string') {
+        localStorage.setItem(key, value)
+      } else {
+        localStorage.setItem(key, JSON.stringify(value))
+      }
```
zoruka, line 41: We need to keep parsing string to JSON string, otherwise it will fail when parsing back

packages/common/src/utils/SafeLocalStorage.ts
tomiir, line 54: I say we add a warning log or similar here, so that we know there was an error loading from localstorage

zoruka:
> @enesozturk @tomiir @zoruka are you going to add some logic to migrate from old keys to new ones? If not, i think you'll probably mess some connections from dapps migrating from the old sdk to the new one

I think old connections will fail to parse when attempted to read the values in the `JSON.parse` and then it will result as if the localStorage where empty. I guess this will only happen one time

zoruka (approved):

PR #2882, opened 2024-09-19
fix: deeplinks not working

magiziz:
# Description

After connecting with a wallet on mobile deeplinks will fail when trying to sign a message or tx. 
This is happening because [this.appkit.getIsConnectedState()](https://github.com/reown-com/appkit/blob/main/packages/appkit/src/universal-adapter/client.ts#L517) is always `false`.

## Type of change

# Associated Issues

For Linear issues: Closes APKT-1133

# Checklist

### DIFF
--- a/packages/adapters/ethers5/src/client.ts
+++ b/packages/adapters/ethers5/src/client.ts
@@ -829,7 +829,6 @@ export class Ethers5Adapter {
       return
     }
 
-    this.appKit?.setIsConnected(true, this.chainNamespace)
     this.appKit?.setPreferredAccountType(
       preferredAccountType as W3mFrameTypes.AccountType,
       this.chainNamespace
--- a/packages/appkit/src/universal-adapter/client.ts
+++ b/packages/appkit/src/universal-adapter/client.ts
@@ -528,7 +528,7 @@ export class UniversalAdapterClient {
           this.syncConnectedWalletInfo()
           await Promise.all([this.appKit?.setApprovedCaipNetworksData(chainNamespace)])
         }
-        
+
         this.syncAccounts()
       })
     } else {
--- a/packages/appkit/src/universal-adapter/client.ts
+++ b/packages/appkit/src/universal-adapter/client.ts
@@ -526,9 +526,10 @@ export class UniversalAdapterClient {
           this.appKit?.setPreferredAccountType(preferredAccountType, chainNamespace)
           this.appKit?.setCaipAddress(address, chainNamespace)
           this.syncConnectedWalletInfo()
-          this.syncAccounts()
           await Promise.all([this.appKit?.setApprovedCaipNetworksData(chainNamespace)])
         }
+        
+        this.syncAccounts()
       })
     } else {
       this.appKit?.resetWcConnection()
--- a/packages/common/src/utils/SafeLocalStorage.ts
+++ b/packages/common/src/utils/SafeLocalStorage.ts
@@ -8,7 +8,7 @@ export type SafeLocalStorageItems = {
   '@appkit/connected_social': string
   '@appkit/connected_social_username': string
   '@appkit/recent_wallets': string
-  WALLETCONNECT_DEEPLINK_CHOICE: { href: string; name: string }
+  '@appkit/deeplink': string
 }
 
 export const SafeLocalStorageKeys = {
@@ -21,7 +21,7 @@ export const SafeLocalStorageKeys = {
   CONNECTED_SOCIAL: '@appkit/connected_social',
   CONNECTED_SOCIAL_USERNAME: '@appkit/connected_social_username',
   RECENT_WALLETS: '@appkit/recent_wallets',
-  DEEPLINK_CHOICE: 'WALLETCONNECT_DEEPLINK_CHOICE'
+  DEEPLINK: '@appkit/deeplink'
 } as const
 
 export const SafeLocalStorage = {
@@ -30,29 +30,14 @@ export const SafeLocalStorage = {
     value: SafeLocalStorageItems[Key]
   ): void {
     if (isSafe()) {
-      localStorage.setItem(
-        key,
-        key === SafeLocalStorageKeys.DEEPLINK_CHOICE ? JSON.stringify(value) : (value as string)
-      )
+      localStorage.setItem(key, value)
     }
   },
   getItem<Key extends keyof SafeLocalStorageItems>(
     key: Key
   ): SafeLocalStorageItems[Key] | undefined {
     if (isSafe()) {
-      const value = localStorage.getItem(key)
-
-      if (value) {
-        if (key === SafeLocalStorageKeys.DEEPLINK_CHOICE) {
-          try {
-            return JSON.parse(value)
-          } catch {
-            return undefined
-          }
-        }
-
-        return value as SafeLocalStorageItems[Key]
-      }
+      return localStorage.getItem(key) || undefined
     }
 
     return undefined
--- a/packages/common/src/utils/SafeLocalStorage.ts
+++ b/packages/common/src/utils/SafeLocalStorage.ts
@@ -8,7 +8,8 @@ export type SafeLocalStorageItems = {
   '@appkit/connected_social': string
   '@appkit/connected_social_username': string
   '@appkit/recent_wallets': string
-  '@appkit/deeplink': string
+  // Hardcoding the value since @walletconnect/universal-provider package uses it
+  WALLETCONNECT_DEEPLINK_CHOICE: string
 }
 
 export const SafeLocalStorageKeys = {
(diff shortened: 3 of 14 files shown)


packages/appkit/src/universal-adapter/client.ts
```diff
@@ -414,6 +414,10 @@ export class UniversalAdapterClient {
       }
     }
 
+    if (this.appKit?.getCaipAddress() && !this.appKit?.getIsConnectedState()) {
+      this.appKit?.setIsConnected(true, this.appKit?.getActiveChainNamespace() || 'eip155')
+    }
```
magiziz, line 419: I wish i could put this logic in a handler, but we're [syncing account](https://github.com/reown-com/appkit/blob/main/packages/appkit/src/universal-adapter/client.ts#L417) before listening to our [handlers](https://github.com/reown-com/appkit/blob/main/packages/appkit/src/universal-adapter/client.ts#L418). 

I tried to refactor it so it's the other way around, but that introduces more issues.

magiziz:
Seems like something that got merged in main broke this PR

packages/appkit/src/universal-adapter/client.ts
```diff
@@ -464,6 +469,8 @@ export class UniversalAdapterClient {
       SafeLocalStorage.removeItem(SafeLocalStorageKeys.WALLET_ID)
       SafeLocalStorage.removeItem(SafeLocalStorageKeys.ACTIVE_CAIP_NETWORK)
 
+      AccountController.setIsConnected(false, this.appKit?.getActiveChainNamespace() || 'eip155')
```
enesozturk, line 472: These `setIsConnected` calls should be already removed completely, is this branch up to date?

packages/appkit/src/universal-adapter/client.ts
```diff
@@ -423,6 +424,10 @@ export class UniversalAdapterClient {
       JSON.stringify(this.appKit?.getCaipNetwork())
     )
 
+    if (this.appKit?.getCaipAddress() && !this.appKit?.getIsConnectedState()) {
```
enesozturk, line 427: We have `getIsConnected` and `getIsConnectedState`, they seem to do the same things, we should remove one of them, and the other should return `ChainController.state.activeCaipAddress` , if there is an address being set we are connected, otherwise we are not. And we stopped using `setIsConnected`

packages/adapters/ethers5/src/client.ts
```diff
@@ -829,7 +829,6 @@ export class Ethers5Adapter {
       return
     }
 
-    this.appKit?.setIsConnected(true, this.chainNamespace)
```
magiziz, line 832: Getting rid of `setIsConnected` here since we'll use `ChainController.state.activeCaipAddress` to see if a user is connected

packages/common/src/utils/SafeLocalStorage.ts
```diff
@@ -30,14 +30,22 @@ export const SafeLocalStorage = {
...
+      const value = localStorage.getItem(key)
+
+      if (value) {
+        try {
+          return JSON.parse(value)
+        } catch {
+          return undefined
+        }
+      }
     }
```
magiziz, line 49: Refactoring the `SafeLocalStorage` here due to the `DEEPLINK_CHOICE` changes

packages/appkit/src/universal-adapter/client.ts
magiziz: Mhmm yeah the branch wasn't up to date

packages/common/src/utils/SafeLocalStorage.ts
```diff
@@ -30,14 +30,29 @@ export const SafeLocalStorage = {
...
+        if (key === SafeLocalStorageKeys.DEEPLINK_CHOICE) {
+          try {
+            return JSON.parse(value)
+          } catch {
+            return undefined
+          }
+        }
+
+        return value as SafeLocalStorageItems[Key]
+      }
```
magiziz, line 55: I wanted to use `JSON.parse` and `JSON.stringify` to not make the code complex, but if the devs were to upgrade their appkit dependency it'd cause issues with their current local storage. 

Instead i applied this logic to `SafeLocalStorageKeys.DEEPLINK_CHOICE` only and the rest of `SafeLocalStorageKeys` should return as is to make it backward compatible.

packages/common/src/utils/SafeLocalStorage.ts
```diff
@@ -8,7 +8,7 @@ export type SafeLocalStorageItems = {
   '@appkit/connected_social': string
   '@appkit/connected_social_username': string
   '@appkit/recent_wallets': string
-  '@appkit/deeplink_choice': string
+  WALLETCONNECT_DEEPLINK_CHOICE: { href: string; name: string }
```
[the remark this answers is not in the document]
enesozturk: @magiziz I'm confused why we don't follow the same naming pattern?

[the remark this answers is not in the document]
magiziz: Since universal provider uses `WALLETCONNECT_DEEPLINK_CHOICE` local storage key to initiate a deep link. We could push a PR to universal provider repo to use same naming pattern. Lmk what you think.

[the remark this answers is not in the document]
enesozturk: Okay if that's the case we should add a JSDoc comment on top of this key IMO. That'a a different case.

But can we still keep returning string and do the parsing outside of it?

packages/common/src/utils/SafeLocalStorage.ts
```diff
@@ -21,7 +21,7 @@ export const SafeLocalStorageKeys = {
   CONNECTED_SOCIAL: '@appkit/connected_social',
   CONNECTED_SOCIAL_USERNAME: '@appkit/connected_social_username',
   RECENT_WALLETS: '@appkit/recent_wallets',
-  DEEPLINK_CHOICE: '@appkit/deeplink_choice'
+  DEEPLINK: '@appkit/deeplink'
```
magiziz, line 24: Changed this key to prevent getting issues with the current local storage value

packages/core/tests/utils/StorageUtil.test.ts
```diff
@@ -45,8 +45,8 @@ describe('StorageUtil', () => {
     it('should set WalletConnect deep link in localStorage', () => {
       const deepLink = { href: 'https://example.com', name: 'Example Wallet' }
       StorageUtil.setWalletConnectDeepLink(deepLink)
-      const savedDL = SafeLocalStorage.getItem(SafeLocalStorageKeys.DEEPLINK_CHOICE)
-      expect(savedDL).toBe(deepLink.href)
+      const savedDL = SafeLocalStorage.getItem(SafeLocalStorageKeys.DEEPLINK)
+      expect(savedDL).toBe(JSON.stringify({ href: deepLink.href, name: deepLink.name }))
```
magiziz, line 49: Since we store primitive value it'll return as pure string instead of an object.

packages/core/src/utils/StorageUtil.ts
```diff
@@ -5,19 +5,22 @@ import { ChainController } from '../controllers/ChainController.js'
...
 // -- Utility -----------------------------------------------------------------
 export const StorageUtil = {
-  setWalletConnectDeepLink({ href }: { href: string; name: string }) {
+  setWalletConnectDeepLink({ name, href }: { href: string; name: string }) {
     try {
-      SafeLocalStorage.setItem(SafeLocalStorageKeys.DEEPLINK_CHOICE, href)
+      const deeplink = JSON.stringify({ href, name })
+      SafeLocalStorage.setItem(SafeLocalStorageKeys.DEEPLINK, deeplink)
+      // Required by @walletconnect/universal-provider to handle deep links
+      localStorage.setItem('WALLETCONNECT_DEEPLINK_CHOICE', deeplink)
```
tomiir, line 13: This needs to stay as-is so that the UP sign client can fetch it. 

Still, should use `SafeLocalStorage`

packages/common/src/utils/SafeLocalStorage.ts
```diff
@@ -8,7 +8,8 @@ export type SafeLocalStorageItems = {
   '@appkit/connected_social': string
   '@appkit/connected_social_username': string
   '@appkit/recent_wallets': string
-  '@appkit/deeplink': string
+  // Hardcoding the value since @walletconnect/universal-provider package uses it
+  WALLETCONNECT_DEEPLINK_CHOICE: string
```
tomiir, line 12: suggestion:
```
// DO NOT CHANGE: @walletconnect/universal-provider requires us to set this specific key
//  This value is a stringified version of { href: stiring; name: string }
```

tomiir (approved):

PR #2895, opened 2024-09-20
Migrate import assertions to with

KeM1aL:


enesozturk:
To provide wide veriety or Node environments it's better to not use any of these syntactic sugars. We recently resolved this problem in #3036.

We appreciate and thank you for your support @KeM1aL!

PR #2920, opened 2024-09-24
fix: use usdt abi for usdt transactions

ignaciosantise:
# Description

USDT ABI does not follow `erc20ABI`, so the send flow doesn't work. The send function needs to use a specific abi for usdt

## Type of change

# Associated Issues
https://github.com/wevm/wagmi/issues/2749

# Checklist


packages/common/src/utils/ConstantsUtil.ts
```diff
@@ -15,5 +15,6 @@ export const ConstantsUtil = {
     eip155: 'Ethereum',
     solana: 'Solana',
     polkadot: 'Polkadot'
-  } as const satisfies Record<ChainNamespace, string>
+  } as const satisfies Record<ChainNamespace, string>,
+  USDT_CONTRACT_ADDRESS: '0xdac17f958d2ee523a2206206994597c13d831ec7'
```
enesozturk, line 19: Where do we get this address? Should this be required to maintained in the future? What it represents?

Can you add a JSDoc on top of it with a description and a resource (if any)?

packages/core/src/controllers/SendController.ts
```diff
@@ -225,15 +225,17 @@ export const SendController = {
         params.receiverAddress &&
         params.tokenAddress
       ) {
+        const tokenAddress = CoreHelperUtil.getPlainAddress(
+          params.tokenAddress as CaipAddress
+        ) as `0x${string}`
```
enesozturk, line 230: We should move `0x${string}` type definition to `common/TypeUtil.ts` and remove from all the duplications in the code - but not for this PR

enesozturk (approved):

packages/common/src/utils/ConstantsUtil.ts
ignaciosantise: @enesozturk That's Tether USD contract address, it's needed to make usdt transactions. In SendController, its the `tokenAddress`.

The address is immutable so it doesn't require maintenance but now that i see, each network has a different address, so this will only work for mainnet 🤔 

We'll need to think another solution

ignaciosantise: maybe checking the name & symbol?

packages/common/src/contracts/usdt.ts
```diff
@@ -0,0 +1,192 @@
+export const usdtABI = [
+  {
+    type: 'event',
+    name: 'Approval',
+    inputs: [
+      {
```
tomiir, line 6: do we really need the full abi? can we use the compact one for just the transfer method?

packages/common/src/utils/ConstantsUtil.ts
tomiir: Weird how it only works on just a few chains. 
Might be worth adding them all

packages/common/src/contracts/usdt.ts
ignaciosantise: done

packages/common/src/utils/ContractUtil.ts
```diff
@@ -0,0 +1,14 @@
+import { erc20ABI } from '../contracts/erc20.js'
+import { usdtABI } from '../contracts/usdt.js'
+import { ConstantsUtil } from './ConstantsUtil.js'
+
+export const ContractUtil = {
+  getABI: (tokenAddress: string) => {
```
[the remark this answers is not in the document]
ignaciosantise: done

packages/common/src/utils/ConstantsUtil.ts
ignaciosantise: I added the ones i've found. Can you guys re-check the list? and see if you can find another one? I couldn't find a good list

packages/common/src/contracts/usdt.ts
ignaciosantise: i guess we could only use the compact abi for erc20, right?

tomiir (approved):

PR #2929, opened 2024-09-25
feat: debug mode / alert

magiziz:
# Description

We're now showing more friendly console errors when a user provides invalid app configuration. Additionally we've added a `debug` option to `createAppKit` to display UI alerts when needed. This will make the dev experience much better 🚀

Example.

```tsx
const modal = createAppKit({
  adapters: [/* Adapters */],
  networks,
  defaultNetwork: mainnet,
  projectId: 'YOUR_PROJECT_ID',
  debug: true // Optional - defaults to false
})
```

## Type of change

# Associated Issues

For Linear issues: Closes APKT-1056

# Showcase (Optional)

https://github.com/user-attachments/assets/a5bae1b1-7c14-46c1-996a-67a1286553d4

https://github.com/user-attachments/assets/d5c9f162-4b99-405c-9e47-e606cf586dd2

# Checklist


packages/wallet/src/W3mFrameProvider.ts
```diff
@@ -451,15 +462,36 @@ export class W3mFrameProvider {
...
+      W3mFrameConstants.APP_GET_SOCIAL_REDIRECT_URI,
+      W3mFrameConstants.APP_GET_FARCASTER_URI
+    ]
+      .map(replaceEventType)
+      .includes(type)
+
+    if (shouldCheckForTimeout && this.onTimeout) {
+      // 15 seconds timeout
+      timer = setTimeout(this.onTimeout, 15_000)
+    }
```
magiziz, line 490: There are some events that don't respond at all, so it wasn't a good idea to have a timeout for all of those. 

Instead i added the necessary ones + i added a 15 second timeout before showing the alert UI.

packages/ui/src/utils/ThemeUtil.ts
```diff
@@ -521,6 +571,10 @@ function createRootStyles(themeVariables?: ThemeVariables) {
         --wui-color-success-glass-060: rgba(38, 217, 98, 0.6);
         --wui-color-success-glass-080: rgba(38, 217, 98, 0.8);
 
+        --wui-color-success-glass-reown-020: rgba(48, 164, 107, 0.2);
+
+        --wui-color-warning-glass-reown-020: rgba(243, 161, 63, 0.2);
```
magiziz, line 576: Since we're using reown colors / backgrounds for the alerts i've included `reown` name here to make it easier.

packages/adapters/ethers/src/client.ts
```diff
@@ -1112,7 +1109,19 @@ export class EthersAdapter {
...
+        onTimeout: () => {
+          AlertController.open(ErrorUtil.ALERT_ERRORS.INVALID_APP_CONFIGURATION, 'error')
+          // eslint-disable-next-line no-console
+          console.error(
+            ErrorUtil.ALERT_ERRORS.ORIGIN_NOT_WHITELISTED_SOCIALS.replace(
+              '{{origin}}',
+              window.origin
+            )
+          )
+        }
```
magiziz, line 1123: I really wanted to add this logic in [W3mFrameProvider](https://github.com/reown-com/appkit/blob/main/packages/wallet/src/W3mFrameProvider.ts), but then i can't use `AlertController`.

Because i'm running to an issue cyclic dependency issue where `@reown/appkit-wallet` is used in `@reown/appkit-core`, so we can't do the same thing the other way around with `@reown/appkit-core`. 

You can try adding `@reown/appkit-core` [here](https://github.com/reown-com/appkit/blob/main/packages/wallet/package.json) and you'll get the error.

packages/appkit-utils/src/WalletConnectLoggerUtil.ts
```diff
@@ -0,0 +1,20 @@
+import { pino } from '@walletconnect/logger'
+
+export const WalletConnectLoggerUtil = {
```
lukaisailovic, line 3: We should rename this

packages/appkit-utils/src/ErrorUtil.ts
```diff
@@ -0,0 +1,10 @@
+export const ErrorUtil = {
+  UniversalProviderErrors: {
+    UNAUTHORIZED_DOMAIN_NOT_ALLOWED: 'Unauthorized: origin not allowed'
+  },
+  ALERT_ERRORS: {
+    INVALID_APP_CONFIGURATION: 'Invalid App Configuration',
+    ORIGIN_NOT_WHITELISTED: `Origin {{origin}} not found on Allowlist - update configuration`,
+    ORIGIN_NOT_WHITELISTED_SOCIALS: `Origin {{origin}} not found on Allowlist - update configuration to enable social login`
+  }
+}
```
lukaisailovic, line 10: is there a better way to handle the replace of origin or other params? 
Right now you have a param here and then you replace it in the code. Thats easy to miss and overlook by someone else. 

What if we change the errors to be functions instead that accept the required params and make it type safe?

Open to another opinion here.

packages/appkit-utils/src/WalletConnectLoggerUtil.ts
magiziz: Updated 🙏

packages/appkit-utils/src/ErrorUtil.ts
magiziz: Good idea. Updated it to use function instead of string 🙏

packages/appkit/src/client.ts
```diff
@@ -416,9 +417,18 @@ export class AppKit {
...
     }
   ) {
+    OptionsController.setDebug(options.debug)
     OptionsController.setProjectId(options.projectId)
     OptionsController.setSdkVersion(options.sdkVersion)
 
+    if (!options.projectId) {
+      AlertController.open('Project ID Not Configured', 'error')
+      // eslint-disable-next-line no-console
+      console.error('Project ID Not Configured - update configuration')
```
lukaisailovic, line 427: Should we  use a custom logger for this, also should we have this error in the consts?

packages/ui/src/utils/ThemeUtil.ts
```diff
@@ -750,6 +833,14 @@ export const colorStyles = css`
     color: var(--wui-color-fg-300);
   }
 
+  .wui-color-fg-325 {
+    color: var(--wui-color-fg-325);
```
enesozturk, line 837: This `325` sounds something new. It's from the new design system or old one? If new one, is it okay to have it together with old values?

enesozturk: Okay saw your previous comment

packages/scaffold-ui/src/partials/w3m-alertbar/styles.ts
```diff
@@ -0,0 +1,14 @@
+import { css } from 'lit'
+
+export default css`
+  :host {
+    display: block;
+    position: absolute;
+    opacity: 1;
```
enesozturk, line 7: Do we need `opacity` to be set true? Since it's already 1 by default

packages/scaffold-ui/src/partials/w3m-alertbar/styles.ts
```diff
@@ -0,0 +1,14 @@
...
+    top: 11px;
+    left: 16px;
```
enesozturk, line 9: Can we use spacing variables for horizontal values? Also why the top value doesn't follow our spacing system? Maybe we need to validate with the designer?

packages/scaffold-ui/src/partials/w3m-alertbar/styles.ts
```diff
@@ -0,0 +1,14 @@
...
+    right: 16px;
+    opacity: 0;
```
enesozturk, line 11: Oh I see this now, we can remove the previous opacity value then

packages/scaffold-ui/src/partials/w3m-alertbar/index.ts
```diff
@@ -0,0 +1,107 @@
...
+        ],
+        {
+          duration: 150,
+          fill: 'forwards',
+          easing: 'ease'
+        }
+      )
+      this.style.cssText = `pointer-events: none`
+    }
+  }
```
enesozturk, line 100: Can we achieve this with CSS animations? I'd prefer to go CSS only way which is nativa way to run animations and more performant.

packages/adapters/wagmi/src/connectors/AuthConnector.ts
```diff
@@ -76,7 +77,14 @@ export function authConnector(parameters: AuthParameters) {
 
     async getProvider() {
       if (!this.provider) {
-        this.provider = W3mFrameProviderSingleton.getInstance(parameters.options.projectId)
+        this.provider = W3mFrameProviderSingleton.getInstance({
+          projectId: parameters.options.projectId,
+          onTimeout: () => {
+            AlertController.open(ErrorUtil.ALERT_ERRORS.INVALID_APP_CONFIGURATION, 'error')
+            // eslint-disable-next-line no-console
+            console.error(ErrorUtil.ALERT_ERRORS.originNotWhitelistedSocials(window.origin))
```
enesozturk, line 85: Why are the console and Alert component messages are different? Cannot they be same and we could instead console.error inside of the `open` function?

packages/adapters/wagmi/src/client.ts
```diff
@@ -217,10 +217,6 @@ export class WagmiAdapter implements ChainAdapter {
   }
 
   public construct(appKit: AppKit, options: AppKitOptions) {
-    if (!options.projectId) {
```
enesozturk, line 220: Where do we check this now? Wagmi adapter requires projectId to be set unlike the other adapters?

packages/adapters/ethers/src/client.ts
```diff
@@ -1112,7 +1109,14 @@ export class EthersAdapter {
 
   private async syncAuthConnector(projectId: string, bypassWindowCheck = false) {
     if (bypassWindowCheck || typeof window !== 'undefined') {
-      this.authProvider = W3mFrameProviderSingleton.getInstance(projectId)
+      this.authProvider = W3mFrameProviderSingleton.getInstance({
+        projectId,
+        onTimeout: () => {
+          AlertController.open(ErrorUtil.ALERT_ERRORS.INVALID_APP_CONFIGURATION, 'error')
+          // eslint-disable-next-line no-console
+          console.error(ErrorUtil.ALERT_ERRORS.originNotWhitelistedSocials(window.origin))
```
enesozturk, line 1117: Why are the console and Alert component messages different? Can they not be the same and could we instead use `console.error` inside of the open function?

enesozturk (commented):
Left some comments

packages/adapters/wagmi/src/client.ts
magiziz: Yeah so we don't construct the parameters unless `projectId` has been provided. 

I've implemented a logic [here](https://github.com/reown-com/appkit/blob/feat/debug/packages/appkit/src/client.ts#L424-L430) that does an early return if `projectId` is not provided which prevents adapters to construct.

packages/scaffold-ui/src/partials/w3m-alertbar/index.ts
magiziz: We have same animation in our snackbar component [here](https://github.com/reown-com/appkit/blob/main/packages/scaffold-ui/src/partials/w3m-snackbar/index.ts) (`<w3m-snackbar>`). Shall we still go for the CSS only way or follow the same pattern like other components ?

packages/scaffold-ui/src/partials/w3m-alertbar/styles.ts
magiziz: Good catch, we should probably do the same for [snackbar](https://github.com/reown-com/appkit/blob/main/packages/scaffold-ui/src/partials/w3m-snackbar/styles.ts#L9).

packages/scaffold-ui/src/partials/w3m-alertbar/styles.ts
magiziz: Oh yeah we don't need it 👍

packages/ui/src/utils/ThemeUtil.ts
magiziz: Yeah i wanted to have it with old values, but wasn't sure if it was going to impact the whole modal UI so kept it simple.

packages/adapters/wagmi/src/client.ts
enesozturk: But the thing is, adapter classes and Appkit class are initialized independently. 

For example for ethers example, we don't already pass projectId to EthersAdapter class, we need to pass it only to AppKit.
Bot for wagmi example, they need to pass projectId to WagmiAdapter as well: https://github.com/reown-com/appkit/blob/main/apps/laboratory/src/pages/library/wagmi.tsx#L16

enesozturk: > if projectId is not provided which prevents adapters to construct

Even though we pass to the AppKit, there is possibility for users that they might forget passing projectId to WagmiAdapter still

magiziz: Makes sense. I pushed a change [here](https://github.com/reown-com/appkit/pull/2929/commits/69e649c1ac3f4ba2a66bf855d9d59ad0dc96d1a4) 🙏

apps/laboratory/src/pages/library/ethers-debug-mode.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { EthersModalInfo } from '../../components/Ethers/EthersModalInfo'
+import { EthersAdapter } from '@reown/appkit-adapter-ethers'
+import { siweConfig } from '../../utils/SiweUtils'
+import { arbitrum, mainnet, optimism, polygon, zkSync, sepolia } from '@reown/appkit/networks'
+
+const ethersAdapter = new EthersAdapter()
+
+const modal = createAppKit({
+  adapters: [ethersAdapter],
+  networks: [arbitrum, mainnet, optimism, polygon, zkSync, sepolia],
```
enesozturk, line 14: Can we use `ConstantsUtil.EvmNetworks` as with the other examples?

apps/laboratory/src/pages/library/ethers5-debug-mode.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { Ethers5Tests } from '../../components/Ethers/Ethers5Tests'
+import { Ethers5ModalInfo } from '../../components/Ethers/Ethers5ModalInfo'
+import { Ethers5Adapter } from '@reown/appkit-adapter-ethers5'
+import { arbitrum, mainnet, optimism, polygon, zkSync, sepolia } from '@reown/appkit/networks'
+
+const ethersAdapter = new Ethers5Adapter()
+
+const modal = createAppKit({
+  adapters: [ethersAdapter],
+  networks: [arbitrum, mainnet, optimism, polygon, zkSync, sepolia],
```
enesozturk, line 15: Same here

apps/laboratory/src/pages/library/solana-debug-mode.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { SolanaAdapter } from '@reown/appkit-adapter-solana/react'
+
+import { ThemeStore } from '../../utils/StoreUtil'
+import { solana, solanaDevnet, solanaTestnet } from '@reown/appkit/networks'
+import { AppKitButtons } from '../../components/AppKitButtons'
+import { ConstantsUtil } from '../../utils/ConstantsUtil'
+import { SolanaTests } from '../../components/Solana/SolanaTests'
+import { HuobiWalletAdapter, SolflareWalletAdapter } from '@solana/wallet-adapter-wallets'
+
+const networks = [solana, solanaTestnet, solanaDevnet]
```
enesozturk, line 11: And `ConstantsUtil.SolanaNetworks` here

apps/laboratory/src/pages/library/wagmi-debug-mode.tsx
```diff
@@ -0,0 +1,45 @@
...
+import { AppKitButtons } from '../../components/AppKitButtons'
+import { WagmiTests } from '../../components/Wagmi/WagmiTests'
+import { ThemeStore } from '../../utils/StoreUtil'
+import { arbitrum, mainnet, optimism, polygon, zkSync, sepolia } from '@reown/appkit/networks'
+import { WagmiModalInfo } from '../../components/Wagmi/WagmiModalInfo'
+import { WagmiAdapter } from '@reown/appkit-adapter-wagmi'
+
+const queryClient = new QueryClient()
+
+const networks = [mainnet, optimism, polygon, zkSync, arbitrum, sepolia]
```
enesozturk, line 13: `ConstantsUtil.EvmNetworks`

packages/adapters/wagmi/src/client.ts
```diff
@@ -218,7 +222,7 @@ export class WagmiAdapter implements ChainAdapter {
 
   public construct(appKit: AppKit, options: AppKitOptions) {
     if (!options.projectId) {
-      throw new Error('appkit:initialize - projectId is undefined')
+      return
```
enesozturk, line 225: Here seems we removed the error throw this time, is it okay to do, how can we warn users?

packages/appkit/src/client.ts
```diff
@@ -416,9 +417,18 @@ export class AppKit {
       sdkVersion: SdkVersion
     }
   ) {
+    OptionsController.setDebug(options.debug)
     OptionsController.setProjectId(options.projectId)
     OptionsController.setSdkVersion(options.sdkVersion)
 
+    if (!options.projectId) {
+      AlertController.open(ErrorUtil.ALERT_ERRORS.PROJECT_ID_NOT_CONFIGURED, 'error')
```
enesozturk, line 425: As another solution, maybe we could map short and long messages in the `ErrorUtil` constant, and inside the `.open` method, use short message for AlertController state, and use long error for `console.error`. 

Somehow I still think we could lower the LoC and use single `console.error` and lint override, wdyt? @magiziz

packages/adapters/wagmi/src/client.ts
magiziz: Yeah i think that's fine, we're checking it in the constructor [here](https://github.com/reown-com/appkit/blob/a6d099085dcd23ef90b90153fe0c7199780689fd/packages/adapters/wagmi/src/client.ts#L147-L149).

packages/appkit/src/client.ts
magiziz: Good idea. I pushed a change where i use `shortMessage` and `longMessage` in `ErrorUtil`.

enesozturk (approved):

PR #2935, opened 2024-09-25
fix: v5 default options

zoruka:
# Description

This PR sets default AppKit options into a constant and changes `enableAnalytics` to true by default.

## Type of change

# Associated Issues

For GH issues: closes #...

# Showcase (Optional)

# Checklist

### DIFF
--- a/packages/base/src/client.ts
+++ b/packages/base/src/client.ts
@@ -390,8 +390,8 @@ export class AppKit<AdapterStoreState = unknown, SwitchNetworkParam = unknown> {
     OptionsController.setPrivacyPolicyUrl(options.privacyPolicyUrl)
     OptionsController.setCustomWallets(options.customWallets)
     OptionsController.setEnableAnalytics(options.enableAnalytics)
-    OptionsController.setOnrampEnabled(options.enableOnramp !== false)
-    OptionsController.setEnableSwaps(options.enableSwaps !== false)
+    OptionsController.setOnrampEnabled(options.enableOnramp)
+    OptionsController.setEnableSwaps(options.enableSwaps)
 
     if (options.metadata) {
       OptionsController.setMetadata(options.metadata)
--- a/packages/core/src/controllers/ApiController.ts
+++ b/packages/core/src/controllers/ApiController.ts
@@ -267,7 +267,7 @@ export const ApiController = {
       ApiController.fetchNetworkImages(),
       ApiController.fetchConnectorImages()
     ]
-    if (OptionsController.state.enableAnalytics === undefined) {
+    if (OptionsController.state.enableAnalytics) {
       promises.push(ApiController.fetchAnalyticsConfig())
     }
     state.prefetchPromise = Promise.race([Promise.allSettled(promises)])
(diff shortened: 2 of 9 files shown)


arein (approved):

packages/base/src/client.ts
```diff
@@ -390,8 +390,8 @@ export class AppKit<AdapterStoreState = unknown, SwitchNetworkParam = unknown> {
     OptionsController.setPrivacyPolicyUrl(options.privacyPolicyUrl)
     OptionsController.setCustomWallets(options.customWallets)
     OptionsController.setEnableAnalytics(options.enableAnalytics)
-    OptionsController.setOnrampEnabled(options.enableOnramp !== false)
-    OptionsController.setEnableSwaps(options.enableSwaps !== false)
+    OptionsController.setOnrampEnabled(options.enableOnramp)
```
enesozturk, line 393: @zoruka Onramp and Swaps will be enabled by default. So when they don't provide these props, they'll be `undefined`. To be able to set them `true`, we need `!== false` options to enable them

packages/core/src/controllers/ApiController.ts
```diff
@@ -267,7 +267,7 @@ export const ApiController = {
       ApiController.fetchNetworkImages(),
       ApiController.fetchConnectorImages()
     ]
-    if (OptionsController.state.enableAnalytics === undefined) {
+    if (OptionsController.state.enableAnalytics) {
```
enesozturk, line 270: Same here, we unles they set it to false, we could enable it like the following
```suggestion
    if (OptionsController.state.enableAnalytics !== false) {
```

packages/core/src/controllers/OptionsController.ts
```diff
@@ -176,23 +188,33 @@ export const OptionsController = {
     state.metadata = metadata
   },
 
-  setOnrampEnabled(enableOnramp: OptionsControllerState['enableOnramp']) {
+  setOnrampEnabled(
+    enableOnramp: OptionsControllerState['enableOnramp'] = ConstantsUtil.DEFAULT_FEATURES.onramp
```
enesozturk, line 192: When you already pass the default value, do you need the type definition?
```suggestion
    enableOnramp = ConstantsUtil.DEFAULT_FEATURES.onramp
```

enesozturk (changes requested):

packages/base/src/client.ts
zoruka: I did add the default parameter in the [constant file](https://github.com/reown-com/appkit/pull/2935/files#diff-5db4770c881137b47d85a901a5f7b972e5845d481e09084aa31cdd653cdce0a4R205-R206) so we don't need hardcoded values like that. Are you sure this is correct?

packages/core/src/controllers/OptionsController.ts
zoruka: We need it because the param has multiple types and the constant just is one of them.

packages/core/src/controllers/ApiController.ts
zoruka: It is the same idea from before, leaving `undefined` as a positive value is miss leading and carries implicit logic. Are you sure we should change to that?

packages/core/src/controllers/OptionsController.ts
zoruka: We could force type the constant, but it is going to be circular dependency

PR #2957, opened 2024-09-27
fix: v5 solana check installed extensions

zoruka:
# Description

Check for installed extensions in "All Wallets" list when using Solana Adapter.

## Type of change

# Associated Issues

For GH issues: closes #...

# Showcase (Optional)

# Checklist

### DIFF
--- a/packages/base/adapters/solana/web3js/client.ts
+++ b/packages/base/adapters/solana/web3js/client.ts
@@ -261,8 +261,11 @@ export class SolanaWeb3JsClient implements ChainAdapter<SolStoreUtilState, CaipN
 
       formatUnits: () => '',
 
-      checkInstalled: (ids: string[] = []) =>
-        ids.some(id => this.availableProviders.some(provider => provider.name === id))
+      checkInstalled: (ids: string[] = []) => {
+        const availableIds = new Set(this.availableProviders.map(provider => provider.name))
+
+        return ids.some(id => availableIds.has(id))
+      }
     }
   }
 
--- a/packages/base/adapters/solana/web3js/client.ts
+++ b/packages/base/adapters/solana/web3js/client.ts
@@ -259,7 +259,10 @@ export class SolanaWeb3JsClient implements ChainAdapter<SolStoreUtilState, CaipN
 
       parseUnits: () => BigInt(0),
 
-      formatUnits: () => ''
+      formatUnits: () => '',
+
+      checkInstalled: (ids: string[] = []) =>
+        ids.some(id => this.availableProviders.some(provider => provider.name === id))
     }
   }
 
--- a/packages/base/adapters/solana/web3js/providers/WalletConnectProvider.ts
+++ b/packages/base/adapters/solana/web3js/providers/WalletConnectProvider.ts
@@ -81,7 +81,7 @@ export class WalletConnectProvider extends ProviderEventEmitter implements Provi
       return acc
     }, {})
 
-    if (this.provider.session) {
+    if (this.provider.session?.namespaces['solana']) {
       this.session = this.provider.session
     } else {
       this.provider.on('display_uri', this.onUri)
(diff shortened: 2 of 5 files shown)


packages/base/adapters/solana/web3js/providers/WalletConnectProvider.ts
```diff
@@ -81,7 +81,7 @@ export class WalletConnectProvider extends ProviderEventEmitter implements Provi
       return acc
     }, {})
 
-    if (this.provider.session) {
+    if (this.provider.session?.namespaces['solana']) {
```
zoruka, line 84: This fixes a edge case where the session only have 'eip155' namespace

packages/base/adapters/solana/web3js/client.ts
```diff
@@ -259,7 +259,10 @@ export class SolanaWeb3JsClient implements ChainAdapter<SolStoreUtilState, CaipN
 
       parseUnits: () => BigInt(0),
 
-      formatUnits: () => ''
+      formatUnits: () => '',
+
+      checkInstalled: (ids: string[] = []) =>
+        ids.some(id => this.availableProviders.some(provider => provider.name === id))
```
tomiir, line 265: This is n2, don't think it's a problem but you could do
```
      checkInstalled: (ids: string[] = []) => {
          const availableProviders = new Set(this.availableProviders.map(provider => provider.id))
          return id.some(id => availableProviders.has(id)
      }
```

packages/base/adapters/solana/web3js/providers/WalletStandardProvider.ts
```diff
@@ -65,17 +65,16 @@ export class WalletStandardProvider extends ProviderEventEmitter implements Prov
...
   public get type() {
-    const FILTER_OUT_ADAPTERS = ['Trust']
-
-    if (FILTER_OUT_ADAPTERS.includes(this.wallet.name)) {
-      return 'EXTERNAL'
-    }
-
-    return 'ANNOUNCED'
+    return 'ANNOUNCED' as const
   }
```
tomiir, line 78: why remov this?

packages/scaffold-ui/src/partials/w3m-connecting-wc-browser/index.ts
```diff
@@ -29,15 +29,20 @@ export class W3mConnectingWcBrowser extends W3mConnectingWidget {
...
       this.error = false
       const { connectors } = ConnectorController.state
-      const announcedConnector = connectors.find(
-        c => c.type === 'ANNOUNCED' && c.info?.rdns === this.wallet?.rdns
+
+      const connector = connectors.find(
+        c =>
+          (c.type === 'ANNOUNCED' && c.info?.rdns === this.wallet?.rdns) ||
+          c.type === 'INJECTED' ||
+          c.name === this.wallet?.name
```
tomiir, line 37: should you match by ID instead?

packages/base/adapters/solana/web3js/providers/WalletStandardProvider.ts
zoruka: I think this was added in an attempt to fix duplicated Trust wallets, which now I've changed the fix in the lines above. And this was making Trust appear without "Installed" tag like in the image 
image

packages/scaffold-ui/src/partials/w3m-connecting-wc-browser/index.ts
zoruka: I think this way makes more explicit that we are matching the name. In case of Solana wallets we would need this to be `c.id === this.wallet?.name` because the id in Solana refers to the wallet name as the Wallet Standard doesn't support ids and the extension do not expose a id.

tomiir (approved):

PR #3073, opened 2024-10-11
chore: add max timeouts for frame requests

magiziz:
# Description

Edited maximum timeouts for frame requests. We currently have our timeout duration on 15 seconds, but we don't throw an error and the timeout duration is low as well. We instead decided to increase the timeout duration to 25 - 30 seconds and also abort the request if it takes that long.

## Type of change

# Associated Issues

For Linear issues: Closes APKT-1147

# Showcase (Optional)

Here is an example if the request takes more than 30 seconds.

image

# Checklist

### DIFF
--- a/apps/laboratory/playwright.config.ts
+++ b/apps/laboratory/playwright.config.ts
@@ -21,7 +21,7 @@ export default defineConfig<ModalFixture>({
   expect: {
     timeout: getValue(60, 15) * 1000
   },
-  timeout: 60 * 1000,
+  timeout: 90 * 1000,
 
   use: {
     /* Base URL to use in actions like `await page.goto('/')`. */
--- a/apps/laboratory/tests/email.spec.ts
+++ b/apps/laboratory/tests/email.spec.ts
@@ -11,7 +11,8 @@ let page: ModalWalletPage
 let validator: ModalWalletValidator
 let context: BrowserContext
 let browserPage: Page
-/* eslint-enable init-declarations */
+let email: Email
+let tempEmail: string
 
 // -- Setup --------------------------------------------------------------------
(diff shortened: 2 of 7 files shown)


apps/laboratory/playwright.config.ts
```diff
@@ -21,7 +21,7 @@ export default defineConfig<ModalFixture>({
   expect: {
     timeout: getValue(60, 15) * 1000
   },
-  timeout: 60 * 1000,
+  timeout: 90 * 1000,
```
enesozturk, line 24: Why do we need increase this? For which situations 60 was not enough? Might it be something we need to improve?

magiziz, line 24: I increased the test timeout to not cause conflicts with other tests, since i added a test which takes 30+ seconds.

apps/laboratory/tests/email.spec.ts
```diff
@@ -105,3 +108,10 @@ emailTest('it should disconnect correctly', async () => {
   await page.disconnect()
   await validator.expectDisconnected()
 })
+
+emailTest('it should abort request if it takes more than 30 seconds', async () => {
+  await page.page.context().setOffline(true)
+  await page.loginWithEmail(tempEmail, false)
+  await page.page.waitForTimeout(30_000)
```
enesozturk, line 115: Confused how this works. we are making test environment to wait 30secs, why?

apps/laboratory/playwright.config.ts
magiziz: Yeah i increased the test timeout to not cause conflicts with other tests, since i added a test which takes 30+ seconds.

enesozturk (approved):

PR #3076, opened 2024-10-11
:feat refactor adapter clients

svenvoskamp:
# Description

Introducing new adapter foundation. All adapters will be handled through the adapter blueprint. All state will be handled on AppKit side. 

## Type of change

- [x]  New feature (non-breaking change which adds functionality)

# Associated Issues

Closes https://linear.app/reown/issue/APKT-1441/[bug]-not-showing-warningerrors-when-changing-manually-in-mm-to-a

# Showcase (Optional)

# Checklist

# Todo 

* Adding tests 
* Ironing out issues with multi-chain and UPA
* Adding all methods to adapters


packages/appkit/src/adapters/ChainAdapterBlueprint.ts
```diff
@@ -0,0 +1,426 @@
...
+ * @template Connector - The type of connector extending ChainAdapterConnector
+ */
+export abstract class AdapterBlueprint<
+  Connector extends ChainAdapterConnector = ChainAdapterConnector
+> {
+  public namespace: ChainNamespace | undefined
+  public caipNetworks?: CaipNetwork[]
+  public projectId?: string
+
+  protected avaiableConnectors: Connector[] = []
```
magiziz, line 29: Should we change this to `connectors`

apps/laboratory/src/pages/library/ethers.tsx
```diff
@@ -17,9 +17,10 @@ const modal = createAppKit({
   defaultNetwork: mainnet,
   projectId: ConstantsUtil.ProjectId,
   features: {
-    analytics: true
-  },
-  customWallets: ConstantsUtil.CustomWallets
```
tomiir, line 22: is this ok?

packages/adapters/ethers/src/client.ts
```diff
@@ -1,114 +1,34 @@
...
-
-  public siweControllerClient = this.options?.siweConfig
-
-  public tokens = HelpersUtil.getCaipTokens(this.options?.tokens)
-
-  public defaultCaipNetwork: CaipNetwork | undefined = undefined
-
-  public adapterType: AdapterType = 'ethers'
+  constructor() {
+    super({})
```
tomiir, line 29: is the empty object required?

packages/common/src/utils/SafeLocalStorage.ts
```diff
@@ -25,7 +27,8 @@ export const SafeLocalStorageKeys = {
   CONNECTED_SOCIAL: '@appkit/connected_social',
   CONNECTED_SOCIAL_USERNAME: '@appkit/connected_social_username',
   RECENT_WALLETS: '@appkit/recent_wallets',
-  DEEPLINK_CHOICE: 'WALLETCONNECT_DEEPLINK_CHOICE'
```
magiziz, line 28: We need to keep `DEEPLINK_CHOICE` as `WALLETCONNECT_DEEPLINK_CHOICE` since universal provider package uses that value.

packages/adapters/ethers/src/client.ts
```diff
@@ -175,889 +95,394 @@ export class EthersAdapter {
...
 
-        if (!provider) {
-          throw new Error('Provider is undefined')
-        }
+      return { address: result }
+    }
 
-        return await EthersMethods.writeContract(data, provider, address, Number(caipNetwork?.id))
-      },
+    return { address: false }
```
tomiir, line 199: is this ok? I know it was like this before but wouldn't having an empty string here be a more logical response?

packages/adapters/ethers/src/client.ts
```diff
@@ -175,889 +95,394 @@ export class EthersAdapter {
...
 
-        if (!address) {
-          throw new Error('Address is undefined')
-        }
+  public async getEnsAddress(
+    params: AdapterBlueprint.GetEnsAddressParams
+  ): Promise<AdapterBlueprint.GetEnsAddressResult> {
+    const { name, appKit } = params
+    if (appKit) {
+      const result = await EthersMethods.getEnsAddress(name, appKit)
```
tomiir, line 194: why does this receive appkit as a parameter?

packages/adapters/ethers/src/client.ts
```diff
@@ -175,889 +95,394 @@ export class EthersAdapter {
...
     })
 
-    if (this.ethersConfig) {
-      this.syncConnectors(this.ethersConfig)
-    }
+    this.listenProviderEvents(selectedProvider)
 
-    if (typeof window !== 'undefined') {
-      this.listenConnectors(true)
+    if (!accounts[0]) {
```
tomiir, line 230: can accounts be `undefined`? this will crash if so

packages/adapters/ethers/src/client.ts
```diff
@@ -175,889 +95,394 @@ export class EthersAdapter {
...
-        provider: this.EIP6963Providers.find(p => p.info.name === walletName)?.provider
-      }
+    if (!this.caipNetworks || !provider) {
+      throw new Error(
+        'UniversalAdapter:connectWalletConnect - caipNetworks or provider is undefined'
+      )
     }
 
-    const activeConfig = providerConfigs[walletId as unknown as keyof typeof providerConfigs]
+    provider.on('display_uri', (uri: string) => {
```
tomiir, line 287: @ganchoradkov does this impact in any way your UP changes?

tomiir, line 287: shouldn't this be done on the universal adapter level ? are we not using universal adapter here anymore?

packages/adapters/ethers/src/client.ts
```diff
@@ -175,889 +95,394 @@ export class EthersAdapter {
...
+    switch (params.providerType) {
+      case 'WALLET_CONNECT':
+      case 'AUTH':
+        await params.provider.disconnect()
+        break
+      case 'ANNOUNCED':
+        await this.revokeProviderPermissions(params.provider as Provider)
+        break
+      case 'EXTERNAL':
+        await this.revokeProviderPermissions(params.provider as Provider)
```
tomiir, line 385:
```suggestion
      case 'ANNOUNCED':
      case 'EXTERNAL':
        await this.revokeProviderPermissions(params.provider as Provider)
```

packages/adapters/ethers/src/client.ts
```diff
@@ -175,889 +95,394 @@ export class EthersAdapter {
...
-        this.appKit?.close()
-      } else {
-        this.appKit?.popTransactionStack(true)
-      }
-    }
-  }
+    if (caipNetwork) {
+      const jsonRpcProvider = new JsonRpcProvider(caipNetwork.rpcUrls.default.http[0], {
+        chainId: caipNetwork.id as number,
+        name: caipNetwork.name
```
tomiir, line 400: might be harder to maintain, but wyt about having this an attribute of the adapter? I mean every adapter will have an underlying provider when connected right?

packages/adapters/ethers/src/client.ts
```diff
@@ -175,889 +95,394 @@ export class EthersAdapter {
...
+  public async switchNetwork(params: AdapterBlueprint.SwitchNetworkParams): Promise<void> {
+    const { caipNetwork, provider, providerType } = params
+    if (providerType === 'WALLET_CONNECT') {
+      ;(provider as UniversalProvider).setDefaultChain(String(`eip155:${String(caipNetwork.id)}`))
+    } else if (providerType === 'AUTH') {
+      const authProvider = provider as W3mFrameProvider
+      await authProvider.switchNetwork(caipNetwork.id)
+      await authProvider.connect({
+        chainId: caipNetwork.id
+      })
```
tomiir, line 482: you can technically directly connect here, using new chain ID!

packages/adapters/ethers5/src/tests/client.test.ts
```diff
@@ -1,960 +1,7 @@
...
-      client['syncConnectedWalletInfo']()
-
-      expect(mockAppKit.setConnectedWalletInfo).toHaveBeenCalledWith(
-        { name: 'Coinbase Wallet', icon: 'coinbase-icon-url' },
-        'eip155'
-      )
-    })
+describe('Ethers5 Client', () => {
+  it('should pass', () => {
+    // Empty test to ensure test suite runs
```
tomiir, line 5: did you move these tests to somewhere else?

packages/adapters/solana/src/client.ts
```diff
@@ -1,699 +1,451 @@
...
-      if (caipAddress && isSolanaAddress && isSolanaNetwork) {
-        this.syncAccount({
-          address: CoreHelperUtil.getPlainAddress(caipAddress),
-          caipNetwork
-        })
+          if (state.data.properties?.name === 'Coinbase Wallet' && !('coinbaseSolana' in window)) {
+            const href = window.location.href
+            window.location.href = `https://go.cb-w.com/dapp?cb_url=${href}`
+          }
+        }
```
tomiir, line 66: we should maybe abstract these to something like `handleMobileWalletRedirection` or similar

packages/adapters/solana/src/client.ts
```diff
@@ -1,699 +1,451 @@
...
-          if (state.data.properties?.name === 'Coinbase Wallet' && !('coinbaseSolana' in window)) {
-            const href = window.location.href
-            window.location.href = `https://go.cb-w.com/dapp?cb_url=${href}`
-          }
+        if (status?.value) {
+          clearInterval(interval)
+          resolve()
         }
-      }
+      }, 1000)
```
tomiir, line 237: does this interval ever get cleaned?

packages/adapters/solana/src/client.ts
```diff
@@ -1,699 +1,451 @@
...
-        this.chainNamespace
-      )
-      await this.syncNetwork(address)
-    } else {
-      this.appKit?.resetWcConnection()
-      this.appKit?.resetNetwork(this.chainNamespace)
-      this.appKit?.resetAccount(this.chainNamespace)
-    }
+  public formatUnits(): string {
+    return ''
```
tomiir, line 250: are these ok ?

packages/adapters/solana/src/client.ts
```diff
@@ -1,699 +1,451 @@
...
-    const rpcSuccessHandler = (_response: W3mFrameTypes.FrameEvent) => {
-      if (!this.appKit) {
-        return
-      }
+    provider.on('disconnect', disconnectHandler)
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-expect-error
+    provider.on('accountsChanged', accountsChangedHandler)
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-expect-error
```
tomiir, line 348: why the disables?

packages/adapters/solana/src/tests/client.test.ts
```diff
@@ -1,234 +1,7 @@
...
-      expect(mockProvider.on).toHaveBeenCalledWith('disconnect', expect.any(Function))
-      expect(mockProvider.on).toHaveBeenCalledWith('accountsChanged', expect.any(Function))
-      expect(mockProvider.on).toHaveBeenCalledWith('connect', expect.any(Function))
-      expect(mockProvider.on).toHaveBeenCalledWith('auth_rpcRequest', expect.any(Function))
-      expect(mockProvider.on).toHaveBeenCalledWith('auth_rpcSuccess', expect.any(Function))
-      expect(mockProvider.on).toHaveBeenCalledWith('auth_rpcError', expect.any(Function))
-    })
+describe('Solana Client', () => {
+  it('should pass', () => {
+    // Empty test to ensure test suite runs
```
tomiir, line 5: same?

packages/adapters/wagmi/src/index.ts
```diff
@@ -2,8 +2,5 @@ import '@reown/appkit-polyfills'
```
tomiir, line 2: is wagmi not impemented?

packages/core/src/controllers/ChainController.ts
```diff
@@ -182,7 +151,7 @@ export const ChainController = {
   setChainAccountData(
     chain: ChainNamespace | undefined,
     accountProps: Partial<AccountControllerState>,
-    replaceState = true
+    replaceState = false
```
tomiir, line 154: is this prop even used anymore?

packages/scaffold-ui/src/views/w3m-account-settings-view/index.ts
```diff
@@ -79,7 +79,7 @@ export class W3mAccountSettingsView extends LitElement {
   // -- Render -------------------------------------------- //
   public override render() {
     if (!this.address) {
-      throw new Error('w3m-account-settings-view: No account provided')
+      // Throw new Error('w3m-account-settings-view: No account provided')
```
tomiir, line 82: is this ok?

packages/appkit/package.json
```diff
@@ -61,6 +61,11 @@
       "types": "./dist/types/exports/auth-provider.d.ts",
       "import": "./dist/esm/exports/auth-provider.js",
       "default": "./dist/esm/exports/auth-provider.js"
+    },
+    "./adapters": {
+      "types": "./dist/types/exports/adapters.d.ts",
+      "import": "./dist/esm/exports/adapters.js",
+      "default": "./dist/esm/exports/adapters.js"
```
tomiir, line 68: is this ok? won't this duplicate the adapters ? They are currently exposed as `@reown/appkit-adapter-whatever`

packages/appkit/src/adapters/ChainAdapterBlueprint.ts
tomiir: in any case `avaiableConnectors` => `availableConnectors`

packages/appkit/src/adapters/ChainAdapterBlueprint.ts
```diff
@@ -0,0 +1,426 @@
...
+   * Formats a bigint value into a decimal string with the specified number of decimals.
+   * @param {AdapterBlueprint.FormatUnitsParams} params - Parameters including value and decimals
+   * @returns {AdapterBlueprint.FormatUnitsResult} The formatted decimal string
+   */
+  public abstract formatUnits(
+    params: AdapterBlueprint.FormatUnitsParams
+  ): AdapterBlueprint.FormatUnitsResult
+}
+
+export namespace AdapterBlueprint {
```
tomiir, line 286: nit: we could move this to a different file

packages/appkit/src/client.ts
```diff
@@ -80,19 +132,38 @@ export class AppKit {
...
+  private async initialize(
+    options: AppKitOptions & {
+      adapters?: ChainAdapter[]
+    } & {
+      sdkVersion: SdkVersion
+    }
+  ) {
+    this.caipNetworks = this.extendCaipNetworks(options)
+
+    this.createAuthProvider()
```
tomiir, line 152: doesn't this need an await too ?

packages/appkit/src/client.ts
```diff
@@ -80,19 +132,38 @@ export class AppKit {
...
+    this.defaultCaipNetwork = this.extendDefaultCaipNetwork(options)
+    this.initControllers(options)
+    this.chainAdapters = await this.createAdapters(
+      options.adapters as unknown as AdapterBlueprint[]
+    )
+    await this.initChainAdapters()
+    this.syncRequestedNetworks()
+    await this.initOrContinue()
+    await this.syncExistingConnection()
+    this.version = options.sdkVersion
```
tomiir, line 164: do these need to be serial? can we Promise.all any of these to save some time?

tomiir: nit: I'd separate in code the attribute setters first and then the initializer calls whenever possible (just so it's a bit clearer)

packages/appkit/src/client.ts
```diff
@@ -575,40 +644,667 @@ export class AppKit {
...
+        ChainController.state.activeChain === chainNamespace &&
+        ChainController.state.activeCaipNetwork?.id
+      ) {
+        this.syncAccount({
+          address,
+          chainId: ChainController.state.activeCaipNetwork.id
+        })
+      }
+    })
+  }
```
tomiir, line 986: glorious 👨‍🍳

packages/appkit/src/client.ts
```diff
@@ -575,40 +644,667 @@ export class AppKit {
...
+      }
+    }
+  }
+
+  private async syncReownName(address: string) {
+    try {
+      const registeredWcNames = await this.getReownName(address)
+      if (registeredWcNames[0]) {
+        const wcName = registeredWcNames[0]
+        this.setProfileName(wcName.name, ChainController.state.activeChain as ChainNamespace)
```
tomiir, line 1127: might be able to remove the if/else by setting `wcName.name || null`

packages/appkit/src/client.ts
```diff
@@ -575,40 +644,667 @@ export class AppKit {
...
+
+  private async syncExistingConnection() {
+    const connectedConnector = SafeLocalStorage.getItem(
+      SafeLocalStorageKeys.CONNECTED_CONNECTOR
+    ) as ConnectorType
+    const connectedNamespace = SafeLocalStorage.getItem(SafeLocalStorageKeys.CONNECTED_NAMESPACE)
+
+    if (connectedConnector === 'WALLET_CONNECT' && connectedNamespace) {
+      this.syncWalletConnectAccount()
+    } else if (connectedConnector && connectedConnector !== 'w3mAuth' && connectedNamespace) {
```
tomiir, line 1159: do we have constants for these ids?

packages/appkit/src/client.ts
```diff
@@ -575,40 +644,667 @@ export class AppKit {
...
+        adapters[namespace].construct({
+          namespace,
+          projectId: this.options?.projectId,
+          networks: this.caipNetworks
+        })
+        if (this.universalProvider) {
+          adapters[namespace].setUniversalProvider(this.universalProvider)
+        }
+        if (this.authProvider) {
+          adapters[namespace].setAuthProvider(this.authProvider)
```
tomiir, line 1254: could it happen that authprovider is not suported on one of these blueprints / networks ?

tomiir (commented):
Awesome work my dude, a lot of comments cause PR is huge.
Only big concern: let's not lose the tests 🙏

packages/appkit/src/adapters/ChainAdapterBlueprint.ts
```diff
@@ -0,0 +1,426 @@
...
+  switchNetwork: { address?: string; chainId: number | string }
+}
+type EventCallback<T extends EventName> = (data: EventData[T]) => void
+
+/**
+ * Abstract class representing a chain adapter blueprint.
+ * @template Connector - The type of connector extending ChainAdapterConnector
+ */
+export abstract class AdapterBlueprint<
+  Connector extends ChainAdapterConnector = ChainAdapterConnector
```
zoruka, line 23: We are lacking to use this generic to fulfill the abstraction. This should receive what is the interface that changes from each network. For example, in Solana we have [signAndSendTransaction](https://docs.reown.com/advanced/multichain/rpc-reference/solana-rpc#solana_signandsendtransaction) method and in Bitcoin we will have [signPsbt](https://docs.reown.com/advanced/multichain/rpc-reference/bitcoin-rpc#signpsbt). The `Connector` interface should extend the `ChainAdapterConnector` and add the extra methods that the specific chain has. Then when adding each `Connector` it should be added through an adaption that maps how it executes stuff into a same interface for every `Connector` of a given network.

Right now we are not using this generic abstraction and `Connector['provider']` type is just unknown which leads into castings like `as unknown as Something` and this is probably going to lead to issues in the future.

packages/appkit/src/adapters/ChainAdapterBlueprint.ts
```diff
@@ -0,0 +1,426 @@
...
+  export type SyncConnectionParams = {
+    id: string
+    namespace: ChainNamespace
+    rpcUrl: string
+  }
+
+  export type SignMessageParams = {
+    message: string
+    address: string
+    provider?: AppKitConnector['provider']
```
zoruka, line 334: `provider` is in the blueprint prototype, why do we need to pass it as a param?

packages/appkit/src/adapters/ChainAdapterBlueprint.ts
```diff
@@ -0,0 +1,426 @@
...
+  ): Promise<AdapterBlueprint.SendTransactionResult>
+
+  /**
+   * Writes a contract transaction.
+   * @param {AdapterBlueprint.WriteContractParams} params - Parameters including receiver address, token amount, token address, from address, method, and ABI
+   * @returns {Promise<AdapterBlueprint.WriteContractResult>} Object containing the transaction hash
+   */
+  public abstract writeContract(
+    params: AdapterBlueprint.WriteContractParams
+  ): Promise<AdapterBlueprint.WriteContractResult>
```
zoruka, line 256: Following up the previous comment, I think `writeContract`, `getEnsAddress`, `parseUnits`, `formatUnits` and so on should be part of the Connector abstraction and not part of the ChainAdapterBlueprint. I mean this because those methods are specifics of EVM networks and should not be needed in every chain.

We could have interfaces like `EVMConnector` or `SolanaConnector` or `BitcoinConnector` extending `ChainAdapterConnector` and implementing the specific methods of each chain, this will be later the `connector` present in this Blueprint prototype and what we expose for our users. With these interfaces we can create classes like `class ExtensionWallet extends EVMConnector` and this specific class would handle extensions to do what the connector should do. This is what would create abstraction between Wallet Connect, Extension Wallet or Auth and also would really help to not need to keep using type casting everywhere.

packages/appkit/src/adapters/ChainAdapterBlueprint.ts
zoruka: As far as I know, to keep namespace typing and class with same name they must be in the same file.

packages/appkit/src/adapters/ChainAdapterBlueprint.ts
```diff
@@ -0,0 +1,426 @@
...
+export abstract class AdapterBlueprint<
+  Connector extends ChainAdapterConnector = ChainAdapterConnector
+> {
+  public namespace: ChainNamespace | undefined
+  public caipNetworks?: CaipNetwork[]
+  public projectId?: string
+
+  protected avaiableConnectors: Connector[] = []
+  protected connector?: Connector
+  protected provider?: Connector['provider']
```
zoruka, line 31: I think we are using the abstraction wrong here as `Connector['provider']` is `unknown` and we should setup generics for this.

zoruka (commented):
Great work! I've let some comments that I think are really important for the concept of this refactor, please take a look with love 🙏

apps/laboratory/src/pages/library/ethers.tsx
enesozturk: + we'd better to left the `socials` and `email` values default

packages/appkit/src/adapters/ChainAdapterBlueprint.ts
enesozturk, line 29:
```suggestion
  protected availableConnectors: Connector[] = []

```

packages/adapters/ethers/src/client.ts
```diff
@@ -1,114 +1,34 @@
...
-
-declare global {
-  interface Window {
-    ethereum?: Record<string, unknown>
-  }
-}
+import type { W3mFrameProvider } from '@reown/appkit-wallet'
+import { EthersMethods } from './utils/EthersMethods.js'
 
 interface Info {
```
enesozturk, line 14: We could remove this and use `Connector['info']` instead? Same for `ethers5` client too.

packages/appkit/src/adapters/ChainAdapterConnector.ts
```diff
@@ -0,0 +1,6 @@
+import type { CaipNetwork } from '@reown/appkit-common'
+import type { Connector } from '@reown/appkit-core'
+
+export interface ChainAdapterConnector extends Connector {
+  chains: CaipNetwork[]
```
enesozturk, line 5: We might need to use `AppKitNetwork` instead of `CaipNetwork`. `AppKitNetwork` includes both Viem network and `CaipNetwork`. Both `createAppKit` and `WagmiAdapter` should be able to handle Viem networks as well.

We are extending networks while initializing the AppKit

packages/adapters/ethers/src/client.ts
svenvoskamp: It is a derived class, so the constructor needs a super call

packages/adapters/ethers/src/client.ts
svenvoskamp: If you look at the `createAdapters` you see that if there is an blueprint, we will create the adapter for it with the specific namespace. If there are no adapters, we will create an UniversalAdapter instance for this namespace. All adapters will be using the same UniversalProvider instance.

We decided to go this way with Felipe, so that everything will be treated in the same way and we're not overcomplicating the codebase (wagmi issues, too much if else statements, etc)

packages/adapters/ethers/src/client.ts
svenvoskamp: Could you elaborate?

packages/adapters/ethers5/src/tests/client.test.ts
svenvoskamp: Tests are WIP

packages/adapters/solana/src/tests/client.test.ts
svenvoskamp: Tests are WIP

packages/appkit/src/client.ts
svenvoskamp: It could be, but the only thing is that we set the underlying frameProvider for the adapter. What the adapter does with this provider, can be defined in the client code

packages/appkit/package.json
svenvoskamp: We need this to import the blueprint class

packages/appkit-utils/src/ConstantsUtil.ts
```diff
@@ -9,13 +9,20 @@ export const ConstantsUtil = {
   SAFE_CONNECTOR_ID: 'safe',
   LEDGER_CONNECTOR_ID: 'ledger',
   EIP6963_CONNECTOR_ID: 'eip6963',
-  AUTH_CONNECTOR_ID: 'w3mAuth',
+  AUTH_CONNECTOR_ID: 'ID_AUTH',
```
tomiir, line 12: why change this?

packages/adapters/solana/src/client.ts
```diff
@@ -1,699 +1,457 @@
...
+   *
+   * These methods are supported only on `wagmi` and `ethers` since the Solana SDK does not support them in the same way.
+   * These function definition is to have a type parity between the clients. Currently not in use.
+   */
+  // eslint-disable-next-line @typescript-eslint/require-await
+  public async getEnsAddress(
+    params: AdapterBlueprint.GetEnsAddressParams
+  ): Promise<AdapterBlueprint.GetEnsAddressResult> {
+    return { address: params.name }
+  }
```
tomiir, line 157: we should probably abstract a bit further (different PR) so that these things are kind of an extension to the blueprint rather than on the blueprint itself

tomiir (approved):

tomiir:
This is missing a changeset!

packages/appkit-utils/src/ConstantsUtil.ts
svenvoskamp: It was interfering with the providerType, so I think this is better to distinguish type and id

magiziz (approved):
🚀

PR #3086, opened 2024-10-15
fix: ethers/ethers5 infinite requests

magiziz:
# Description

There's a bug where if you switch to any network other than ethereum you'll get a lot of network requests sometimes up to 50k.

This happens because the ethers/ethers5 adapters switches caip address every second. It goes back and forth between the ethereum caip address and polygon caip address if you're connected to polygon. This results into hitting the rpc endpoint loads of times.

image

## Type of change

# Associated Issues

For GH issues: closes #...

# Checklist


packages/adapters/ethers/src/client.ts
```diff
@@ -631,9 +631,12 @@ export class EthersAdapter {
       if (provider) {
         const { addresses, chainId } = await EthersHelpersUtil.getUserInfo(provider)
         const firstAddress = addresses?.[0]
-        const caipAddress = `${this.chainNamespace}:${chainId}:${firstAddress}` as CaipAddress
+        const caipNetwork = this.caipNetworks.find(c => c.id === chainId) ?? this.caipNetworks[0]
+        const caipAddress =
+          `${this.chainNamespace}:${caipNetwork?.id}:${firstAddress}` as CaipAddress
```
tomiir, line 636: Have you tried switching from solana to evm? doesn't that break this cause of the "hardcoded" `this.chainNamespace` ?

tomiir, line 636: why not use the caipNetwork id directly?

tomiir (approved):

packages/adapters/ethers/src/client.ts
```diff
@@ -631,9 +631,12 @@ export class EthersAdapter {
       if (provider) {
         const { addresses, chainId } = await EthersHelpersUtil.getUserInfo(provider)
```
tomiir, line 632: why source the infor from here ?

magiziz: I think it's because during the construct of the adapter we'd need to get the address and chain id directly from the provider so users are still connected even if they refresh the page. The logic happens [here](https://github.com/reown-com/appkit/blob/main/packages/adapters/ethers/src/client.ts#L522-L524).

enesozturk (approved):

packages/adapters/ethers/src/client.ts
tomiir: No, I meant `${caipNetwork.namespace}:${caipNetwork?.id}:${firstAddress}` or `${caipNetwork.caipNetworkId}:${firstAddress}`

(unsure about proper names but you get what I mean)

PR #3096, opened 2024-10-16
refactor: token balance error handling

magiziz:
# Description

Refactored token balance error message and ensured that token balances are only fetched again after 5 seconds if the token balance API fails.

## Type of change

# Showcase (Optional)

image

# Checklist


apps/laboratory/tests/email.spec.ts
```diff
@@ -100,9 +100,18 @@ emailTest('it should show names feature only for EVM networks', async ({ library
 emailTest('it should show loading on page refresh', async () => {
   await page.page.reload()
   await validator.expectConnectButtonLoading()
+  await page.page.waitForTimeout(5000)
```
tomiir, line 103: is this required?

packages/core/src/controllers/AccountController.ts
```diff
@@ -239,7 +248,10 @@ export const AccountController = {
         SwapController.setBalances(SwapApiUtil.mapBalancesToSwapTokens(response.balances))
       }
     } catch (error) {
-      SnackController.showError('Failed to fetch token balance')
+      if (chainId && state.tokenFetchFailureChainIds) {
+        state.tokenFetchFailureChainIds.add(chainId)
```
tomiir, line 252: Are we not clearing this? What happens if there's a true flake in the request?
I think we should silently retry after some time

apps/laboratory/tests/email.spec.ts
magiziz: Yeah it'd need to finish loading before i can open the modal again

enesozturk: But we don't know when the loading will finish right? Cannot be wait for an element to be visible instead?

tomiir: I think best way here is to add a `validator.expectConnectButtonReady` or similar that does 
```
    const connectButton = this.page.getByTestId('connect-button')
    await expect(connectButton).toContainText('Connect', {
      timeout: MAX_WAIT
    }))
``` 
That way you do not use arbitrary timeouts

magiziz: Good idea, let me add that 🙏

tomiir (approved):

enesozturk (approved):

PR #3100, opened 2024-10-16
fix: wagmi smart account excessive network requests

magiziz:
# Description

When switching to smart account the wagmi adapter causes excessive network requests that's not needed. This results into getting 429 errors.

## Type of change

# Associated Issues

For Linear issues: Closes APKT-1302

# Showcase (Optional)

errors

# Checklist


magiziz:
We went from having 74 requests to 14 requests. These 14 requests are not spam requests and can be reduced from secure site if needed.

image

packages/adapters/wagmi/src/client.ts
```diff
@@ -745,15 +746,6 @@ export class WagmiAdapter implements ChainAdapter {
...
           this.appKit?.setLoading(true)
-          const connectors = getConnectors(this.wagmiConfig)
-          const currentConnector = connectors.find(c => c.id === connector.id)
-
-          if (currentConnector) {
-            await reconnect(this.wagmiConfig, {
-              connectors: [currentConnector]
-            })
-            this.appKit?.setLoading(false)
-          }
```
magiziz, line 756: I removed this since if we're in a `reconnecting` status we don't need to reconnect again since Wagmi already does that for us [here](https://github.com/wevm/wagmi/blob/main/packages/core/src/actions/reconnect.ts#L80-L82).

packages/adapters/wagmi/src/connectors/AuthConnector.ts
```diff
@@ -68,12 +73,14 @@ export function authConnector(parameters: AuthParameters) {
       await provider.disconnect()
     },
 
-    async getAccounts() {
-      const provider = await this.getProvider()
-      const { address } = await provider.connect()
-      config.emitter.emit('change', { accounts: [address as Address] })
```
magiziz, line 74: We've been calling `provider.connect` again which does more network requests, instead we can store the accounts internally.

packages/adapters/wagmi/src/client.ts
```diff
@@ -1012,10 +1004,6 @@ export class WagmiAdapter implements ChainAdapter {
         }
       })
 
-      provider.onIsConnected(() => {
-        provider.connect()
-      })
```
magiziz, line 1017: Whenever we check if we're connected we call `provider.connect`, which resulted into more excessive network calls

tomiir: We should check if this is required for previously conected connections I think

zoruka: I wonder why we have this only for wagmi client

enesozturk (approved):

tomiir (approved):

PR #3144, opened 2024-10-24
chore: optimize connection handling for connectors

magiziz:
# Description

There is an issue where if you refresh your page even tho you've connected with e.g metamask it'll take too much time to finish loading. This is because we're always reconnecting to auth connector even tho we haven't connected with it before.

For this change we'll make sure that we're only reconnecting to auth connector if a user has already connected before.

## Type of change

# Associated Issues

For Linear issues: Closes APKT-1358

# Showcase (Optional)

**Before:**

https://github.com/user-attachments/assets/e5ca1521-e18b-4e10-ba5c-8cfb8a0b8abb

**After:**

https://github.com/user-attachments/assets/8a2600da-98d3-4197-a424-ded5678b048f

# Checklist

### DIFF
--- a/packages/adapters/wagmi/src/client.ts
+++ b/packages/adapters/wagmi/src/client.ts
@@ -974,11 +974,7 @@ export class WagmiAdapter implements ChainAdapter {
     bypassWindowCheck = false
   ) {
     if (bypassWindowCheck || (typeof window !== 'undefined' && connector)) {
-      this.appKit?.setLoading(true)
       const provider = (await connector.getProvider()) as W3mFrameProvider
-      const isLoginEmailUsed = provider.getLoginEmailUsed()
-
-      this.appKit?.setLoading(isLoginEmailUsed)
 
       provider.onRpcRequest((request: W3mFrameTypes.RPCRequest) => {
         if (W3mFrameHelpers.checkIfRequestExists(request)) {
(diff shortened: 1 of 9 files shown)


packages/adapters/wagmi/src/client.ts
```diff
@@ -974,11 +975,7 @@ export class WagmiAdapter implements ChainAdapter {
     bypassWindowCheck = false
   ) {
     if (bypassWindowCheck || (typeof window !== 'undefined' && connector)) {
-      this.appKit?.setLoading(true)
       const provider = (await connector.getProvider()) as W3mFrameProvider
-      const isLoginEmailUsed = provider.getLoginEmailUsed()
-
-      this.appKit?.setLoading(isLoginEmailUsed)
```
magiziz, line 981: We don't need to set loading true or false here since we can already do that in `this.syncAccount` when wagmi gives us back status of `'reconnecting'`

packages/adapters/wagmi/src/connectors/AuthConnector.ts
```diff
@@ -105,9 +105,8 @@ export function authConnector(parameters: AuthParameters) {
 
     async isAuthorized() {
       const provider = await this.getProvider()
-      const { isConnected } = await provider.isConnected()
 
-      return isConnected
+      return Promise.resolve(provider.getLoginEmailUsed())
```
magiziz, line 109: We can use the local storage value to see if someone has connected before. Since `isAuthorized` function checks whether the connector has connected previously and is still authorized.

This means when wagmi is trying to reconnect if `isAuthorized` returns false it'll ignore the connection ([here](https://github.com/wevm/wagmi/blob/4825ef4a901c7281e1ca69d2f3378d167108c98a/packages/core/src/actions/reconnect.ts#L77-L78) is the code from wagmi)

arein (approved):
awesome

PR #3153, opened 2024-10-28
chore: override metadata values

magiziz:
# Description

We should be able to override metadata values instead of using window / document properties only.

## Type of change

# Associated Issues

For Linear issues: Closes APKT-1366

# Checklist

### DIFF
--- a/packages/appkit/src/client.ts
+++ b/packages/appkit/src/client.ts
@@ -446,8 +446,10 @@ export class AppKit {
     }
 
     this.adapters = options.adapters
-
-    this.setMetadata(options)
+    
+    if (!options.metadata) {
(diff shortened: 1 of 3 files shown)


packages/appkit/src/client.ts
```diff
@@ -446,8 +446,10 @@ export class AppKit {
     }
 
     this.adapters = options.adapters
-
-    this.setMetadata(options)
+    
+    if (!options.metadata) {
```
tomiir, line 450: Wouldn't it be better to have a default metadata value and always set the metadata? 

Also, if there IS a metadata, I don't see where it's being set

tomiir (approved):

PR #3172, opened 2024-10-30
feat: siwx solana verifier

zoruka:
# Description

- Adds Solana verifier into the list of verifiers of SIWX.
- Removes `messageVersion` from SIWX because it actually don't make sense since verifiers are targeted to verify the bytes and not exactly what is inside of it.

## Type of change

# Associated Issues

For Linear issues: Closes APKT-1386
For GH issues: closes #...

# Showcase (Optional)

# Checklist


packages/siwx/src/core/SIWXVerifier.ts
```diff
@@ -3,13 +3,9 @@ import type { SIWXSession } from '@reown/appkit-core'
 
 export abstract class SIWXVerifier {
   public abstract readonly chainNamespace: ChainNamespace
-  public abstract readonly messageVersion: string
```
zoruka, line 6: I'm removing this because it actually is not correct since the base verifiers are not validating the message content but the bytes generated by the signature in the string. When we create specific verifiers that checks the message content, then `shouldVerify` method should be overriden with the specific logic in the specific class.

packages/siwx/src/verifiers/SolanaVerifier.ts
```diff
@@ -0,0 +1,22 @@
+import { SIWXVerifier } from '../core/SIWXVerifier.js'
+import type { SIWXSession } from '@reown/appkit-core'
+import nacl from 'tweetnacl'
+import bs58 from 'bs58'
+
+export class SolanaVerifier extends SIWXVerifier {
+  public readonly chainNamespace = 'solana'
```
enesozturk, line 7: If possible we could import ConstantsUtil from `common` package and do this like
```suggestion
  public readonly chainNamespace = ConstantsUtil.CHAIN.SOLANA
```

enesozturk (approved):

packages/siwx/src/configs/DefaultSIWX.ts
```diff
@@ -11,7 +11,7 @@ const DEFAULTS = {
       getNonce: async () => Promise.resolve(Math.round(Math.random() * 10000).toString())
     }),
 
-  getDefaultVerifiers: () => [new EIP155Verifier()],
+  getDefaultVerifiers: () => [new EIP155Verifier(), new SolanaVerifier()],
```
tomiir, line 14: nit: by looking at this, wouldn't it be better to have `{ eip155: EIP155Verifier, solana: SolanaVerifier }`?

packages/siwx/src/core/SIWXVerifier.ts
tomiir: so if different message versions are used, verifiers are responsible to handle this change. 
aka if they change the message version, they should update their verifier?

tomiir (approved):

packages/siwx/src/configs/DefaultSIWX.ts
zoruka: My thought on this was to allow combine multiple verifiers in a same chain if for some reason that's needed. For example if validating signature and something in server side specific for the chain. Do you think this makes sense?

packages/siwx/src/core/SIWXVerifier.ts
zoruka: If the Verifier have a validation for something inside the message then yes, but the verifiers I have implemented so far only validates the resultant string signature, then it makes no sense to check for the version.

PR #3175, opened 2024-10-31
refactor: sa gas availability check logics for swaps

enesozturk:
# Description

In swaps, we have logic to check if user have enough balance to pay the gas; `SwapController.hasInsufficientToken`. For specifically to smart accounts, we've returning direct value for gas availability if the account is smart account.

This logic directly returning `true` and it have two issues:
- Bypasses the balance check for swap (not for paying gas)
- It should be return `false` to make the swap available, returning `true` means user has insufficient token as function name describes. 

## Type of change

# Associated Issues

For Linear issues: Closes APKT-1288
For GH issues: closes #...

# Showcase (Optional)

# Checklist

### DIFF
--- a/packages/core/src/controllers/SwapController.ts
+++ b/packages/core/src/controllers/SwapController.ts
@@ -186,13 +186,6 @@ export const SwapController = {
       !state.sourceToken?.decimals ||
       !NumberUtil.bigNumber(state.sourceTokenAmount).isGreaterThan(0)
     const invalidSourceTokenAmount = !state.sourceTokenAmount
-    console.log(
-      '>>> SWAP',
-      caipAddress,
-      invalidToToken,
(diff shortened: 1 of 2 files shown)


magiziz (approved):

packages/core/src/controllers/SwapController.ts
```diff
@@ -186,6 +186,13 @@ export const SwapController = {
       !state.sourceToken?.decimals ||
       !NumberUtil.bigNumber(state.sourceTokenAmount).isGreaterThan(0)
     const invalidSourceTokenAmount = !state.sourceTokenAmount
+    console.log(
```
tomiir, line 189: log

tomiir (approved):

PR #3179, opened 2024-10-31
feat: improvements in siwx flow

zoruka:
# Description

- Add SIWX handler for account switch
- Clean up SIWX initialization
- Fixes for multichain switches

## Type of change

# Associated Issues

For Linear issues: Closes APKT-1320
For GH issues: closes #...

# Showcase (Optional)

# Checklist


packages/adapters/solana/src/client.ts
```diff
@@ -477,14 +477,14 @@ export class SolanaAdapter implements ChainAdapter {
         provider.chains.find(chain => chain.caipNetworkId === caipNetworkId) || provider.chains[0]
 
       if (connectionChain) {
+        ProviderUtil.setProvider(this.chainNamespace, provider)
+        this.provider = provider
+
```
tomiir, line 482: why was this needed?

packages/appkit/src/universal-adapter/client.ts
```diff
@@ -246,12 +247,35 @@ export class UniversalAdapterClient {
...
+          signature = (response as { signature: string }).signature
+        } else {
+          signature = await provider.request(
+            {
+              method: 'personal_sign',
+              params: [message, address]
+            },
+            ChainController.state.activeCaipNetwork?.caipNetworkId
+          )
+        }
```
tomiir, line 276: hmmmm won't we need an if for every chain here? 
should we have a unified `signMessage` method that abstract this somehow ?

packages/adapters/solana/src/client.ts
zoruka: It was a leftover attempt to keep states synced as the provider was being get as undefined at some point due to the async task on network switch. I'm going to revert it.

packages/appkit/src/universal-adapter/client.ts
zoruka: Theoretically this should have been handled by the Solana Adapter, but when Solana adapter is not added as an adapter option it will use UPA and this case was not being handled correctly, that's why I had to add this conditional. If there is different implementation for each network then it UPA will need to handle all of them. I think the way to do it would have these methods implemented somewhere else and then call these methods from this function, what do you think?

tomiir (approved):

PR #3186, opened 2024-11-04
refactor: deprecate w3m naming on public components

enesozturk:
# Description

Updates public component names with `appkit-*` prefix and drops `w3m-*` with backwards compatibility.

The `w3m-*` components are still available but they are rendering `appkit-*` components inside instead of duplicating the components.

## Type of change

# Associated Issues

For GH issues: closes #...

# Showcase (Optional)

# Checklist


packages/appkit/src/library/react/index.ts
```diff
@@ -23,11 +26,17 @@ type ThemeVariablesOptions = AppKitOptions['themeVariables']
...
   namespace JSX {
     interface IntrinsicElements {
+      'appkit-button': Pick<
+        AppKitButton,
+        'size' | 'label' | 'loadingLabel' | 'disabled' | 'balance'
+      >
+      'appkit-connect-button': Pick<AppKitConnectButton, 'size' | 'label' | 'loadingLabel'>
+      'appkit-account-button': Pick<AppKitAccountButton, 'disabled' | 'balance'>
+      'appkit-network-button': Pick<AppKitNetworkButton, 'disabled'>
       'w3m-connect-button': Pick<W3mConnectButton, 'size' | 'label' | 'loadingLabel'>
```
enesozturk, line 36: For the newer versions we can drop the old component TS supports, this would caurage users to switch to `<appkit-* />` ones

magiziz (approved):

zoruka (approved):

apps/laboratory/src/components/AppKitInfo.tsx
```diff
@@ -30,7 +30,7 @@ export function AppKitInfo({ caipAddress, address, chainId, clientId }: AppKitIn
             <Heading size="xs" textTransform="uppercase" pb="2">
               CAIP Address
             </Heading>
-            <Text data-testid="appkit-caip-address">{caipAddress}</Text>
+            <Text data-testid="w3m-caip-address">{caipAddress}</Text>
```
zoruka, line 33: should this stay `appkit-caip-address`?

apps/laboratory/tests/shared/pages/ModalPage.ts
```diff
@@ -422,7 +422,7 @@ export class ModalPage {
   }
 
   async switchNetworkWithNetworkButton(networkName: string) {
-    const networkButton = this.page.getByTestId('w3m-network-button')
+    const networkButton = this.page.getByTestId('wui-network-button')
```
tomiir, line 425: Is this ok ? Shouldnt it be appkit?

apps/laboratory/tests/shared/validators/ModalValidator.ts
```diff
@@ -83,7 +83,7 @@ export class ModalValidator {
   }
 
   async expectCaipAddressHaveCorrectNetworkId(caipNetworkId: CaipNetworkId) {
-    const address = this.page.getByTestId('appkit-caip-address')
+    const address = this.page.getByTestId('w3m-caip-address')
```
tomiir, line 86: Also here

packages/ui/src/composites/wui-network-button/index.ts
```diff
@@ -21,7 +21,7 @@ export class WuiNetworkButton extends LitElement {
   // -- Render -------------------------------------------- //
   public override render() {
     return html`
-      <button data-testid="w3m-network-button" ?disabled=${this.disabled}>
+      <button data-testid="wui-network-button" ?disabled=${this.disabled}>
```
tomiir, line 24: Sane?

apps/laboratory/tests/shared/pages/ModalPage.ts
enesozturk: Yeap, I think we don't need to pass data test id to `w3m-network-button`, the `wui-network-button` already have inside, so this is safe to do

apps/laboratory/tests/shared/validators/ModalValidator.ts
enesozturk: This is just aligning the test ids on laboratory with the other test id's - these are not the ids in AppKit itself. 

I'd like to change all but that wouldn't make sense to replace all `w3m` calls in this PR I guess

packages/ui/src/composites/wui-network-button/index.ts
enesozturk: Fixing the data test id since it's better to have `wui-` instaed of `w3m-`

apps/laboratory/src/components/AppKitInfo.tsx
enesozturk: Since this is laboratory it's not mandatory - it's just annoying that others have `w3m-` and only this one has `appkit-`. I've just aligned them with the other laboratory elements. Something outside of the AppKit

tomiir (approved):

PR #3194, opened 2024-11-04
chore: siwe to siwx migration

zoruka:
# Description

- Replace the whole SIWE logic inside AppKit package with SIWX;
- Let all SIWX code present in AppKit inside `SIWXUtil` object centralizing the logic spread around the code;
- Correction into default implementation of SIWX `LocalStorage` to validate getting sessions.

## Type of change

# Associated Issues

For Linear issues: Closes APKT-1408
For GH issues: closes #...

# Showcase (Optional)

# Checklist


packages/adapters/solana/src/client.ts
```diff
@@ -187,7 +188,7 @@ export class SolanaAdapter extends AdapterBlueprint {
     )
 
     return {
-      signature: new TextDecoder().decode(signature)
+      signature: bs58.encode(signature)
```
tomiir, line 191: are we changing signature schemes?

packages/appkit/src/client.ts
```diff
@@ -699,78 +708,20 @@ export class AppKit {
...
+          (await this.universalProvider?.client?.core?.crypto?.getClientId()) || null
+        )
+
+        const isAuthenticated =
+          this.universalProvider &&
+          (await SIWXUtil.universalProviderAuthenticate({
+            universalProvider: this.universalProvider,
+            chains: this.caipNetworks?.map(network => network.caipNetworkId) || [],
+            methods: OPTIONAL_METHODS
+          }))
```
tomiir, line 721: nit: don't like nested inline definitions

tomiir: also what happens with [] chains?

packages/appkit/src/universal-adapter/client.ts
```diff
@@ -60,12 +63,32 @@ export class UniversalAdapter extends AdapterBlueprint {
...
+        {
+          method: 'personal_sign',
+          params: [message, address]
+        },
+        ChainController.state.activeCaipNetwork?.caipNetworkId
+      )
+    }
 
-    return { signature: signature as `0x${string}` }
+    return { signature }
```
tomiir, line 91: the more I think about it, this method should not exist on the adapters

packages/core/src/utils/SIWXUtil.ts
```diff
@@ -1,4 +1,256 @@
...
+    return OptionsController.state.siwx
+  },
+  async initializeIfEnabled() {
+    const siwx = OptionsController.state.siwx
+    const address = CoreHelperUtil.getPlainAddress(ChainController.getActiveCaipAddress())
+    const network = ChainController.getActiveCaipNetwork()
+
+    if (!(siwx && address && network)) {
+      return
+    }
```
tomiir, line 26: should we log anything here?

packages/core/src/utils/SIWXUtil.ts
```diff
@@ -1,4 +1,256 @@
...
+  },
+  async requestSignMessage() {
+    const siwx = OptionsController.state.siwx
+    const address = CoreHelperUtil.getPlainAddress(ChainController.getActiveCaipAddress())
+    const network = ChainController.getActiveCaipNetwork()
+    const client = ConnectionController._getClient()
+
+    if (!(siwx && address && network && client)) {
+      throw new Error('SIWX is not enabled')
+    }
```
tomiir, line 61: why throw here but not on the initialization?

packages/core/src/utils/SIWXUtil.ts
```diff
@@ -1,4 +1,256 @@
...
+
+      if (properties.isSmartAccount) {
+        SnackController.showError('This application might not support Smart Accounts')
+      } else {
+        SnackController.showError('Signature declined')
+      }
+
+      EventsController.sendEvent({
+        type: 'track',
+        event: 'SIWX_AUTH_ERROR',
```
tomiir, line 105: are these events tracked? else we need to notify ata so they start consuming them

packages/core/src/utils/SIWXUtil.ts
```diff
@@ -1,4 +1,256 @@
...
+        properties
+      })
+
+      // eslint-disable-next-line no-console
+      console.error('SWIXUtil:requestSignMessage', error)
+    }
+  },
+  async cancelSignMessage() {
```
tomiir, line 113: do you need to update session state?

packages/core/src/utils/SIWXUtil.ts
```diff
@@ -1,4 +1,256 @@
...
+
+    if (!siwx || namespaces.size !== 1) {
+      return false
+    }
+
+    // Ignores chainId and account address to get other message data
+    const siwxMessage = await siwx.createMessage({
+      chainId: '' as CaipNetworkId,
+      accountAddress: ''
+    })
```
tomiir, line 173: how does this work ?

packages/core/src/utils/SIWXUtil.ts
```diff
@@ -1,4 +1,256 @@
...
+            uri: cacao.p.aud,
+            domain: cacao.p.domain,
+            nonce: cacao.p.nonce,
+            version: cacao.p.version || siwxMessage.version,
+            expirationTime: cacao.p.exp,
+            statement: cacao.p.statement,
+            issuedAt: cacao.p.iat,
+            notBefore: cacao.p.nbf,
+            requestId: cacao.p.requestId,
+            resources: cacao.p.resources
```
tomiir, line 211: anything against spreading `cacao.p` and overriding wherever needed?

packages/scaffold-ui/src/utils/NetworkUtil.ts
```diff
@@ -1,22 +1,9 @@
...
-
-      if (shouldSignOut) {
-        await SIWEController.signOut()
-      }
-
-      RouterUtil.navigateAfterNetworkSwitch()
-    } else {
-      RouterUtil.navigateAfterNetworkSwitch()
-    }
+    return Promise.resolve()
```
tomiir, line 7: why keep this a promise¿

packages/adapters/solana/src/client.ts
zoruka: no this decoding was originally wrong, but we were not using it before. This was decoding bytes into utf-8 which would result in wrong characters, later was not possible to verify the signature.

packages/core/src/utils/SIWXUtil.ts
zoruka: I don't think so, this is just ignoring siwx if it is not enabled

packages/core/src/utils/SIWXUtil.ts
zoruka: because this function should not be called if siwx was not enabled/initialized

packages/core/src/utils/SIWXUtil.ts
zoruka: hmmm I'm not sure, how can I make sure this is tracked or not?

packages/core/src/utils/SIWXUtil.ts
zoruka: `ConnectionController.disconnect` will call the session revoke, inside AppKit siwx is stateless. Maybe I should bring the code from `ConnectionController.disconnect` into another function of `SIWXUtil`, wdyt?

packages/core/src/utils/SIWXUtil.ts
zoruka: At this point we don't have these data yet because the user is not connected, `universalProvider.authenticate` requires the other data from `siwxMessage`, then we create the message without the not available fields to get those params from the user SIWXConfig

packages/appkit/src/universal-adapter/client.ts
zoruka: I'm not sure what to do about this 😞 if we don't do it, it will break the signature for solana

packages/scaffold-ui/src/utils/NetworkUtil.ts
zoruka: This is not in use 🤔, I'll remove it

apps/laboratory/tests/multichain/multichain-ethers-solana-email-siwe.spec.ts
```diff
@@ -59,7 +59,7 @@ test('it should switch to different namespace', async () => {
   const chainName = 'Solana'
 
   await page.switchNetwork(chainName)
-  await page.openNetworks()
+  await validator.expectUnauthenticated()
```
zoruka, line 62: In this case the expected behavior of after network switch is to keep the switch network view opened, right? I've changed to `expectUnauthenticated` because when switching to non evm with siwe, the current behavior is no validation of signature so unauthenticated.

apps/laboratory/tests/multichain/multichain-ethers-solana-siwe.spec.ts
```diff
@@ -28,6 +28,9 @@ test.beforeAll(async ({ browser }) => {
   await modalPage.load()
   await modalPage.qrCodeFlow(modalPage, walletPage)
   await modalValidator.expectConnected()
+  await modalPage.promptSiwe()
+  await walletPage.handleRequest({ accept: true })
+  await modalValidator.expectAuthenticated()
```
zoruka, line 33: This won't expect to prompt siwe because before we were filtering out the solana chain ids when trying to authenticate, as we are not doing it anymore and the OneClickAuth feature won't accept multiple namespaces, the sign message step is now needed for multichain auth.

https://github.com/reown-com/appkit/blob/main/packages/appkit/src/client.ts#L722-L724

apps/laboratory/tests/multichain/multichain-ethers-solana-siwe.spec.ts
```diff
@@ -53,5 +56,5 @@ test('it should switch networks and sign siwe', async () => {
 test('it should switch to Solana and validate chain title', async () => {
   const chainName = 'Solana'
   await modalPage.switchNetwork(chainName)
-  await modalValidator.expectSwitchChainView(chainName.toLocaleLowerCase())
+  await modalValidator.expectSwitchedNetworkOnNetworksView(chainName)
```
zoruka, line 59: Again, if the expected behavior is to keep switch chain view open after the chain is switched, this test was wrong.

zoruka (commented):
I've let some comments why changing the tests as this concerns me and we might check if these changes makes sense

cc @tomiir

apps/laboratory/tests/multichain/multichain-ethers-solana-siwe.spec.ts
tomiir: how was it passing before?

magiziz (approved):
🙌

elix1er:
wen release?

PR #3278, opened 2024-11-19
hotfix: fix broken state swaps

svenvoskamp:
# Description

hotfix: fix broken state swaps

## Type of change

# Associated Issues

 Closes [APKT-xxx](https://linear.app/reown/issue/APKT-1489/[bug]-appkit-swaps-are-broken)
# Showcase (Optional)

# Checklist


packages/core/src/controllers/SwapController.ts
```diff
@@ -202,7 +202,7 @@ export const SwapController = {
       invalidSourceTokenAmount,
       availableToSwap:
         caipAddress && !invalidToToken && !invalidSourceToken && !invalidSourceTokenAmount,
-      isAuthConnector: type === 'AUTH'
+      isAuthConnector: type === 'ID_AUTH'
```
tomiir, line 205: was 'AUTH' stil a valid type?

magiziz, line 205: Should we maybe add `ID_AUTH` to a constants file ?

magiziz (approved):

packages/core/src/controllers/SwapController.ts
tomiir: this should be type checked already else it'd say that it would always be false 🤔 
but we should use the constant regardless IMO

svenvoskamp: `AUTH` and `ID_AUTH` shares the same type as `ConnectorType`

authConnectorId = `ID_AUTH`
authConnectorType = `AUTH`

This is confusing, so we should definitely improve this

PR #3286, opened 2024-11-20
fix: remove condition on SIWE sign out mapping

zoruka:
# Description

This PR is removing the condition for SIWE `signOut` call so it keeps backwards compatibility as this was not being checked before.

## Type of change

# Associated Issues

For Linear issues: Closes APKT-1497
For GH issues: closes #...

# Showcase (Optional)

# Checklist


packages/siwe/src/mapToSIWX.ts
```diff
@@ -96,18 +106,21 @@ export function mapToSIWX(siwe: AppKitSIWEClient): SIWXConfig {
...
       }
-
-      throw new Error('Failed to sign out')
     },
 
     async setSessions(sessions) {
       if (sessions.length === 0) {
-        await siwe.methods.signOut()
+        try {
+          await siwe.methods.signOut()
```
tomiir, line 120: how come one is siwe.signout and the other is siwe.methods.signout()

zoruka: it should be `siwe.methods`, good catch. This happens because of how [this class](https://github.com/reown-com/appkit/blob/main/packages/siwe/src/client.ts#L49) is built.

tomiir (approved):

PR #3299, opened 2024-11-21
:fix switch network improvements & wagmi to recognize wc session

svenvoskamp:
# Description

Resolve cases where wagmi doesn't recognize wc session and unable to switch from unsupported network"

## Type of change

# Associated Issues

For Linear issues: Closes [APKT-xxx](https://linear.app/reown/issue/DEC-256/connecting-wallet-thats-not-already-on-op-sepolia-results-in-no)
For GH issues: closes #...

# Showcase (Optional)

# Checklist


packages/adapters/wagmi/src/connectors/UniversalConnector.ts
```diff
@@ -276,14 +276,14 @@ export function walletConnect(
...
-        }
-
         await provider.request({
           method: 'wallet_switchEthereumChain',
           params: [{ chainId: numberToHex(chainId) }]
         })
+
+        if (chainToSwitch?.caipNetworkId) {
+          provider.setDefaultChain(chainToSwitch?.caipNetworkId as string)
+        }
```
tomiir, line 286: Why move to after?  How does it change things?

packages/appkit/src/client.ts
```diff
@@ -1282,10 +1283,12 @@ export class AppKit {
...
           } catch (error) {
-            adapter?.switchNetwork({
-              provider: this.universalProvider,
-              caipNetwork: ChainController.state.activeCaipNetwork as CaipNetwork
-            })
+            if (adapter?.reconnect) {
+              adapter?.reconnect({
+                id: 'walletConnect',
+                type: 'WALLET_CONNECT'
+              })
```
tomiir, line 1290: can we add a comment on why this is required? not sure why this or a switch network would be required if a connection fails? 🤔

packages/adapters/wagmi/src/connectors/UniversalConnector.ts
svenvoskamp: For some wallets (e.g MM) will emit a `chainChanged` event when this is called and the chainId is not the active chainId. So thats why we first need to switch to the correct chain.

packages/appkit/src/client.ts
svenvoskamp: Added a comment:

`Handle edge case where wagmi detects existing connection but lacks to complete UniversalProvider instance. Connection attempt fails due to already connected state - reconnect to restore provider state.`

tomiir (approved):

PR #3379, opened 2024-12-03
remove axios deps from experimental package and refactor code

KannuSingh:
# Description

- Remove axios deps from experimental package 
- Refactored cosigner Service implementation and test code

## Type of change

# Associated Issues

For GH issues: closes #...

# Showcase (Optional)

# Checklist


lukaisailovic (approved):

packages/experimental/src/smart-session/utils/CosignerService.ts
```diff
@@ -38,26 +37,67 @@ export async function sendCoSignerRequest<
...
-      headers
-    })
 
-    return response.data
+    // Construct query string
+    const queryString = new URLSearchParams({
+      ...queryParams
+    }).toString()
+
+    const fullUrl = queryString ? `${url}?${queryString}` : url
```
enesozturk, line 48: Can we use the URL class to construct the fullUrl instead of manually concatenating strings? Using the URL class would ensure proper encoding of query parameters and make the code more robust and readable. For example:

```js
const fullUrl = new URL(url);
Object.entries(queryParams).forEach(([key, value]) => {
  fullUrl.searchParams.append(key, value);
});
```

This approach avoids potential issues with improperly encoded characters and handles edge cases gracefully.

@KannuSingh

packages/experimental/src/smart-session/utils/CosignerService.ts
```diff
@@ -38,26 +37,67 @@ export async function sendCoSignerRequest<
...
+
+    // If response is empty or has no content
+    if (
+      response.status === 204 ||
+      contentLength === '0' ||
+      !contentType ||
+      !contentType.includes('application/json')
+    ) {
+      return undefined as TResponse
+    }
```
enesozturk, line 82: If we are going to use this logics in different requests, can we move it to a named function to encapsulate the logic for readability and reusability? 

For example:
```js
function isResponseEmptyOrInvalid(response: Response, contentType: string | null, contentLength: string | null): boolean {
  return (
    response.status === 204 || // No content
    contentLength === '0' || // Explicit zero-length
    !contentType || // No content type
    !contentType.includes('application/json') // Non-JSON response
  );
}
...
if (isResponseEmptyOrInvalid(response, contentType, contentLength)) {
  return undefined as TResponse;
}
```

packages/experimental/src/smart-session/utils/CosignerService.ts
```diff
@@ -38,26 +37,67 @@ export async function sendCoSignerRequest<
...
+
+    // Check if the response is ok (status in the range 200-299)
+    if (!response.ok) {
+      // Try to parse error response
+      const errorData = await response.text()
+      throw new CoSignerApiError(response.status, errorData)
+    }
+
+    // Handle void responses explicitly
+    const contentLength = response.headers.get('content-length')
```
enesozturk, line 71: Should we handle `null` or `undefined` cases for this to ensure they have valid values?

```js
const contentLength = response.headers.get('content-length') || '0';
const contentType = response.headers.get('content-type') || '';
```

KannuSingh: The method `isResponseEmptyOrInvalid` will handle this, so might not need to add valid values.

enesozturk (approved):

PR #3382, opened 2024-12-03
:feat listen to transactions for every adapter.

svenvoskamp:
# Description

Currently when a transaction has been made in AppKit, the balance won't update. This is a blocker for several clients. With this PR we are listening to transactionEvents and fetch the balance on these events.

wagmi: for wagmi we are using https://wagmi.sh/core/api/actions/watchPendingTransactions
ethers: for ethers we are using the `pending` event 
solana: For Solana we need to emit the event in the provider itself

## Type of change

# Associated Issues

For Linear issues: Closes APKT-820
For GH issues: closes #...

# Showcase (Optional)

# Checklist


packages/adapters/ethers/src/client.ts
```diff
@@ -479,6 +480,19 @@ export class EthersAdapter extends AdapterBlueprint {
...
+  private listenPendingTransactions(provider: Provider) {
+    const browserProvider = new BrowserProvider(provider)
+
+    try {
+      browserProvider.on('pending', () => {
+        this.emit('pendingTransactions')
+      })
+    } catch (error) {
+      // eslint-disable-next-line no-console
+      console.error('Error listening to pending transactions', error)
```
lukaisailovic, line 492: Should we use `@walletconnect/logger` with new `somethingLogger` instance?

packages/appkit/src/client.ts
```diff
@@ -1244,6 +1250,12 @@ export class AppKit {
       }
     })
 
+    adapter?.on('pendingTransactions', () => {
+      if (AccountController.state.address && ChainController.state.activeCaipNetwork?.id) {
+        this.updateBalance()
+      }
+    })
+
```
lukaisailovic, line 1258: Should we do optional chaining here? I wonder if there is a better way to handle this. We should check if undefined, or assert somewhere earlier in the flow

packages/adapters/ethers/src/client.ts
magiziz, line 492: Maybe we can use [AlertController](https://github.com/reown-com/appkit/blob/main/packages/core/src/controllers/AlertController.ts) for users who use debug mode ?

svenvoskamp: Using AlertController as suggested by @magiziz

magiziz (approved):

PR #3460, opened 2024-12-13
refactor: disconnect logic and state update for multiple adapter use cases

enesozturk:
# Description

- Refactors's disconnect functionality to make sure it's calling the connectionController's `disconnect` method
  - Moves all adapters iteration from adapter specific code (from `appkit/client` to `ChainController`)
- Makes sure it's removing account and network data for each adapter

## Type of change

# Associated Issues

For GH issues: closes #...

# Showcase (Optional)

# Checklist

### DIFF
--- a/packages/appkit/src/client.ts
+++ b/packages/appkit/src/client.ts
@@ -904,20 +904,12 @@ export class AppKit {
         const provider = ProviderUtil.getProvider<UniversalProvider | Provider | W3mFrameProvider>(
           ChainController.state.activeChain as ChainNamespace
         )
-
         const providerType =
           ProviderUtil.state.providerIds[ChainController.state.activeChain as ChainNamespace]
 
         await adapter?.disconnect({ provider, providerType })
 
         this.setStatus('disconnected', ChainController.state.activeChain as ChainNamespace)
-
-        StorageUtil.deleteConnectedConnector()
-        StorageUtil.deleteActiveCaipNetworkId()
-
-        ChainController.state.chains.forEach(chain => {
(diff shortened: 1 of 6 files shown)


packages/appkit/src/client.ts
```diff
@@ -904,20 +904,12 @@ export class AppKit {
...
         await adapter?.disconnect({ provider, providerType })
 
         this.setStatus('disconnected', ChainController.state.activeChain as ChainNamespace)
-
-        StorageUtil.deleteConnectedConnector()
-        StorageUtil.deleteActiveCaipNetworkId()
-
-        ChainController.state.chains.forEach(chain => {
-          this.resetAccount(chain.namespace as ChainNamespace)
-        })
```
magiziz, line 920: Not needed ?

packages/core/src/controllers/ChainController.ts
```diff
@@ -519,5 +520,48 @@ export const ChainController = {
         allAccounts: []
       })
     )
+  },
+
+  async disconnect() {
+    try {
+      const disconnectResults = await Promise.allSettled(
```
magiziz, line 527: Do we need `Promise.allSettled` ? Maybe `Promise.all` is enough i don't think it'll throw an error since we have try and catch block.

packages/appkit/src/client.ts
enesozturk: They are handled on the `ChainController` side, this function should do things for one namespace at a time instead of iterating adapters. I moved similar logic to `ChainController`

svenvoskamp (approved):

PR #3484, opened 2024-12-17
chore: adapter client improvements

magiziz:
# Description

We’re currently facing hydration and connector issues because of how our adapters are initialized. This PR cleanup will make sure that both adapters and connectors are prepared right at the start of the appkit client initialization rather than waiting for some async operations to complete.

Also we no longer need to dynamically import the siwe package in the appkit client as this is now handled at the SIWE client level 🙏

## Type of change

# Associated Issues

For Linear issues: Closes APKT-1376

# Checklist


packages/adapters/wagmi/src/tests/client.test.ts
```diff
@@ -97,6 +101,38 @@ describe('WagmiAdapter', () => {
       expect(adapter.namespace).toBe('eip155')
     })
 
+    it('should set wagmi connectors', () => {
+      const addWagmiConnector = vi.spyOn(adapter, 'addWagmiConnector' as any)
```
zoruka, line 105: This is a nit and maybe hard thing, but I think we should not test private methods like that, we should test private methods through the public methods that call them, thats how it happens in the practice. https://softwareengineering.stackexchange.com/a/100966

packages/appkit/src/client.ts
```diff
@@ -1750,12 +1695,12 @@ export class AppKit {
     return this.chainAdapters?.[namespace]
   }
 
+  private isSafe() {
+    return typeof window !== 'undefined'
+  }
```
zoruka, line 1700: we have an util for this https://github.com/reown-com/appkit/blob/main/packages/core/src/utils/CoreHelperUtil.ts#L38-L40

zoruka (approved):
nice work! let some nit comments

packages/appkit/src/client.ts
```diff
@@ -1889,18 +1833,26 @@ export class AppKit {
...
 
+  private async createConnectorsForAdapter(namespace: ChainNamespace) {
+    await this.createUniversalProviderForAdapter(namespace)
+    this.createAuthProviderForAdapter(namespace)
+  }
+
+  private onConnectors(chainNamespace: ChainNamespace) {
+    const adapter = this.getAdapter(chainNamespace)
+
+    adapter?.on('connectors', this.setConnectors.bind(this))
```
svenvoskamp, line 1844: We have this implementation now for `wagmi` but can we also use this listener for Ethers and Solana? I know that the connectors are 'ready' when we want to retrieve them, but I think it is a better DX if all the adapters work the same, so can we also emit connectors from these adapters?

packages/adapters/wagmi/src/client.ts
```diff
@@ -358,40 +356,49 @@ export class WagmiAdapter extends AdapterBlueprint {
...
+    this.addConnector({
+      id: connector.id,
+      explorerId: PresetsUtil.ConnectorExplorerIds[connector.id],
+      imageUrl: options?.connectorImages?.[connector.id] ?? connector.icon,
+      name: PresetsUtil.ConnectorNamesMap[connector.id] ?? connector.name,
+      imageId: PresetsUtil.ConnectorImageIds[connector.id],
+      type: PresetsUtil.ConnectorTypesMap[connector.type] ?? 'EXTERNAL',
+      info:
+        connector.id === ConstantsUtil.INJECTED_CONNECTOR_ID ? undefined : { rdns: connector.id },
+      provider,
```
tomiir, line 382: what happens if provider is undefined?

packages/adapters/wagmi/src/tests/client.test.ts
tomiir: Yeah, maybe we should just initialize an appkit instance and see if it's calling this internally as it should

magiziz: Good idea, yeah i'll do that.

zoruka:
The pr is missing changeset

packages/appkit/src/client.ts
tomiir: Yes, all adapters should work the same. 
Also, what happens when all adapters fire this concurrently?  Could race conditions happen here?

packages/appkit/src/adapters/ChainAdapterBlueprint.ts
```diff
@@ -155,6 +161,8 @@ export abstract class AdapterBlueprint<
 
       return true
     })
+
+    this.emit('connectors', this.availableConnectors)
```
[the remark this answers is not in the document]
magiziz: I think it's fine since we already filter them [here](https://github.com/reown-com/appkit/blob/chore/adapter-improvement/packages/appkit/src/adapters/ChainAdapterBlueprint.ts#L154-L163) and it [ConnectorController](https://github.com/reown-com/appkit/blob/chore/adapter-improvement/packages/core/src/controllers/ConnectorController.ts#L35-L44) as well.

packages/adapters/wagmi/src/client.ts
magiziz: This is how we do it atm, we don't add `provider` in wagmi adapter [here](https://github.com/reown-com/appkit/blob/main/packages/adapters/wagmi/src/client.ts#L383-L393). This is because we're getting the connector id and find the connector in wagmi config instead, you can check it [here](https://github.com/reown-com/appkit/blob/main/packages/adapters/wagmi/src/client.ts#L441)

packages/appkit/src/client.ts
magiziz: Yeah they are exactly the same already. Here is an example from [ethers](https://github.com/reown-com/appkit/blob/main/packages/adapters/ethers/src/client.ts#L249). Once `this.addConnector` gets called we'll emit `'connectors'` event. 

Also @tomirr i don't think any race condition could happen because we always prepare listeners [here](https://github.com/reown-com/appkit/blob/chore/adapter-improvement/packages/appkit/src/client.ts#L1850), before syncing connectors [here](https://github.com/reown-com/appkit/blob/chore/adapter-improvement/packages/appkit/src/client.ts#L1852).

tomiir (approved):

svenvoskamp (approved):

tomiir (approved):

PR #3527, opened 2024-12-23
:fix reload iframe after aborting farcaster

svenvoskamp:
# Description

## Type of change

# Associated Issues

For GH issues: closes #...

# Showcase (Optional)

# Checklist


.changeset/pre.json
```diff
@@ -0,0 +1,49 @@
+{
```
enesozturk, line 1: Guess we need to remove this?

packages/core/src/controllers/RouterController.ts
```diff
@@ -185,6 +189,9 @@ export const RouterController = {
   },
 
   goBack() {
+    const shouldReload =
```
enesozturk, line 192: Would it make sense to set this value inside the setTimeout in case of potential closure issues? Defining the variable outside might lead to problems if the value of `shouldReload` is modified elsewhere in the code before the setTimeout executes

enesozturk (approved):

packages/core/src/controllers/RouterController.ts
svenvoskamp: If we would add in the setTimeout, the `this.state.view` is not accurate anymore, so that is why I added this before. The value of `shouldReload` won't change in the code below.

PR #3541, opened 2024-12-26
fix: wc sync network

tomiir:
# Description

Fixes issue where WC connections while on a network not supported by the wallet would cause AppKit to default to an unsupported network

## Type of change

# Associated Issues

Closes APKT-1202

# Checklist


rkhadem4 (approved):

datis56 (approved):

packages/adapters/wagmi/src/client.ts
```diff
@@ -213,21 +213,20 @@ export class WagmiAdapter extends AdapterBlueprint {
...
 
             this.emit('accountChanged', {
-              address: accountData.address,
-              chainId: accountData.chainId
+              address: accountData.address
             })
           }
 
-          if (accountData.chainId) {
+          if (accountData.chainId !== prevAccountData?.chainId) {
```
enesozturk, line 229: What about the other clients? Is it going to have same functionality for the other clients?

tomiir: Given wagmi triggers all of these updates at the same time, this is to be able to individually re-fire the events. On others adapters they are handled individually. SAdly there's no way to get the "previous" data else I would apply the same optimization there :\

packages/appkit/src/client.ts
```diff
@@ -1249,21 +1249,25 @@ export class AppKit {
...
+        chainId: ChainController.state.activeCaipNetwork?.id,
+        chain: namespace
       })
       this.setLoading(false)
       this.setStatus('connected', namespace)
     } else {
       this.setLoading(false)
-      this.setStatus('disconnected', namespace)
-      StorageUtil.removeConnectedNamespace(namespace)
+      if (namespace) {
```
enesozturk, line 1267: Should we instead wrap parent block with this if, bc in all conditions (if and elses) we are using namespace, if it's not truthy, we shouldn't run any of those logics?

packages/appkit/src/client.ts
```diff
@@ -1406,10 +1410,25 @@ export class AppKit {
...
+
+      // We try and find the address for this network in the session object.
+
+      const activeCaipNetworkId = ChainController.state.activeCaipNetwork?.id
+
+      const caipAddress =
+        namespaceAccounts.find(account => {
+          const [, chainId] = account.split(':')
+
+          return chainId === activeCaipNetworkId?.toString()
```
enesozturk, line 1424: Is chainId and caipNetworkId same? One is `1`, the other one is `eip155:1` if I understood the code right

packages/appkit/src/client.ts
```diff
@@ -1406,10 +1410,25 @@ export class AppKit {
...
+        }) || namespaceAccounts[0]
 
       if (caipAddress) {
+        if (caipAddress.split(':').length !== 3) {
```
enesozturk, line 1428: Seems like we could have a utility in common (or another pck) for this instead of doing here, wdyt? We have `CaipNetworkUtil` in `appkit-utils`

packages/appkit/src/client.ts
```diff
@@ -1406,10 +1410,25 @@ export class AppKit {
...
+          throw new Error('Invalid caip address')
+        }
+        const [, chainId, address] = caipAddress.split(':')
```
enesozturk, line 1431: Same for this one, we should already have a parser util function

packages/appkit/src/client.ts
```diff
@@ -1543,11 +1549,31 @@ export class AppKit {
     this.setStatus('connected', chainNamespace)
 
     if (chainIdToUse && chainNamespace === activeNamespace) {
-      const caipNetwork = this.caipNetworks?.find(n => n.id.toString() === chainIdToUse.toString())
-      const fallBackCaipNetwork = this.caipNetworks?.find(n => n.chainNamespace === chainNamespace)
-      this.setCaipNetwork(caipNetwork || fallBackCaipNetwork)
+      let caipNetwork = this.caipNetworks?.find(n => n.id.toString() === chainIdToUse.toString())
```
enesozturk, line 1552: We we need to parse id's to string? Isn't safer to not do it? On top as I see `chainIdToUse` is coming from `params` which it should be number for EVM already - just to double check

Preferably we should avoid type casting IMO, wdyt?

packages/appkit/src/client.ts
```diff
@@ -1543,11 +1549,31 @@ export class AppKit {
...
+      const shouldSupportsAllNetworks = ChainController.getNetworkProp(
+        'supportsAllNetworks',
+        chainNamespace
+      )
+
+      if (!shouldSupportsAllNetworks) {
+        // Connection can be requested for a chain that is not supported by the wallet so we need to use approved networks here
+        const caipNetworkIds = this.getApprovedCaipNetworkIds() || []
+        const caipNetworkId = caipNetworkIds.find(
+          id => id.split(':')[1] === chainIdToUse.toString()
```
enesozturk, line 1563: Same for using utility, we could also benefit type's of the id's in the utility, would be really nice to return number for EVM and string for others, this would allow us to drop type casting

magiziz (approved):

packages/appkit/src/client.ts
tomiir: Did and also abstracted all the event listeners to their own function, this was getting weird

packages/appkit/src/client.ts
tomiir: bad naming, will change

packages/appkit/src/client.ts
tomiir: They are both `string | number` ! `syncAccount` could be triggered from for example wagmi which would set `chainId` to a number and this condition would not match cause our internal caipnetworks are strings

packages/appkit/src/client.ts
tomiir: utility already responds everything in string form, uni¡fying types so it's good imo

PR #3558, opened 2024-12-30
feat: add email / social info to `useAppKitAccount` hook

magiziz:
# Description

Added email, username and some other auth info to `useAppKitAccount` hook

## Type of change

# Associated Issues

For Linear issues: Closes APKT-1410

# Showcase (Optional)

Screenshot 2024-12-30 at 15 38 20

# Checklist


packages/appkit/src/client.ts
```diff
@@ -1223,6 +1240,9 @@ export class AppKit {
       await provider.getSmartAccountEnabledNetworks()
       this.setLoading(false)
     })
+    provider.onSocialConnected(({ userName }) => {
+      this.setUsername(userName, ChainController.state.activeChain as ChainNamespace)
+    })
```
magiziz, line 1245: We can only get the username when connecting to a social. This means we'd have to store the username in local storage like we currently do and use that value in our `AccountController.state.username` state.

Magic doesn't give us back username when calling `this.magic.getUser()` in secure site

apps/laboratory/src/components/AppKitInfo.tsx
```diff
@@ -55,6 +59,42 @@ export function AppKitInfo({ caipAddress, address, chainId, clientId }: AppKitIn
...
+          )}
+
+          <Box>
+            <Heading size="xs" textTransform="uppercase" pb="2">
+              Smart Account Status
+            </Heading>
+            <Text data-testid="w3m-sa-account-status">
+              {isSmartAccountDeployed ? 'Deployed' : 'Not Deployed'}
+            </Text>
+          </Box>
```
tomiir, line 97: maybe nit but can we make a component for this?  like `InfoRow` or something

packages/appkit/src/client.ts
tomiir: wym? wouldn't `connect` have the username when reconnecting?

packages/appkit/src/client.ts
```diff
@@ -490,6 +496,14 @@ export class AppKit {
     AccountController.setProfileImage(profileImage, chain)
   }
 
+  public setEmail: (typeof AccountController)['setEmail'] = (email, chain) => {
+    AccountController.setEmail(email, chain)
+  }
+
+  public setUsername: (typeof AccountController)['setUsername'] = (username, chain) => {
+    AccountController.setUsername(username, chain)
```
tomiir, line 504: this is only for social accounts right? wouldn't it make sense to store the user object directly?

packages/core/exports/react.ts
```diff
@@ -20,14 +20,23 @@ export function useAppKitNetworkCore(): Pick<
...
+
   const { activeCaipAddress } = useSnapshot(ChainController.state)
 
   return {
     caipAddress: activeCaipAddress,
     address: CoreHelperUtil.getPlainAddress(activeCaipAddress),
     isConnected: Boolean(activeCaipAddress),
-    status
+    status,
+    user: {
```
tomiir, line 34: shouldn't this be part of wallet info instead? it's not really user data but rather wallet data which has user info from auth/email/SA

tomiir: nvm

tomiir, line 34: I'd call this embedded wallet info or something like that

packages/core/exports/vue.ts
```diff
@@ -23,9 +29,29 @@ export function useAppKitAccount() {
...
+    state.value.user.accountType = val ?? null
+  })
+
+  const unsubscribeEmail = AccountController.subscribeKey('email', val => {
+    state.value.user.email = val ?? null
+  })
+
+  const unsubscribeUsername = AccountController.subscribeKey('username', val => {
+    state.value.user.username = val ?? null
+  })
```
tomiir, line 46: they are all updated at the same time (on connection) so I'd keep a single object for this and we serialize all data as is to the hook. wyt?

packages/core/src/controllers/AccountController.ts
```diff
@@ -144,6 +146,14 @@ export const AccountController = {
     ChainController.setAccountProp('profileImage', profileImage, chain)
   },
 
+  setEmail(email: AccountControllerState['email'], chain: ChainNamespace) {
+    ChainController.setAccountProp('email', email, chain)
+  },
+
+  setUsername(username: AccountControllerState['username'], chain?: ChainNamespace) {
+    ChainController.setAccountProp('username', username, chain)
+  },
```
tomiir, line 155: Tricky case, cause they should be updated on all namespaces but not all namespaces might support auth wallet

packages/appkit/src/client.ts
magiziz: For some reason it doesn't expose username. It only exposes username when connecting to a social

apps/laboratory/src/components/AppKitInfo.tsx
magiziz: good idea

packages/core/exports/vue.ts
magiziz: yeap let me do that

tomiir (approved):

PR #3575, opened 2025-01-06
feat: expose getPublicKey method for Bitcoin adapter and connectors

enesozturk:
# Description

Exposes PublicKey to developers for the ones need on Bitcoin implementations. Returns `allAccounts` array in `useAppKitAccount` hooks and subscribe methods for devs to get active account and reach to `publicKey` field.

Example usage;
```js
const { allAccounts, address } = useAppKitAccount()
const bip122Account = allAccounts?.find(a => a.address === address)
const publicKey = bip122Account?.publicKey
```

## Type of change

# Associated Issues

For GH issues: closes #...

# Showcase (Optional)

# Checklist

### DIFF
--- a/packages/adapters/bitcoin/package.json
+++ b/packages/adapters/bitcoin/package.json
@@ -21,26 +21,6 @@
       "types": "./dist/types/exports/index.d.ts",
       "import": "./dist/esm/exports/index.js",
       "default": "./dist/esm/exports/index.js"
-    },
-    "./react": {
-      "types": "./dist/types/exports/react.d.ts",
-      "import": "./dist/esm/exports/react.js",
-      "default": "./dist/esm/exports/react.js"
-    },
-    "./vue": {
-      "types": "./dist/types/exports/vue.d.ts",
-      "import": "./dist/esm/exports/vue.js",
-      "default": "./dist/esm/exports/vue.js"
-    }
-  },
-  "typesVersions": {
-    "*": {
-      "react": [
-        "./dist/types/exports/react.d.ts"
-      ],
-      "vue": [
-        "./dist/types/exports/vue.d.ts"
-      ]
     }
   },
   "dependencies": {
@@ -55,17 +35,11 @@
     "sats-connect": "3.0.1"
   },
   "devDependencies": {
-    "@types/react": "18.3.1",
-    "@types/react-dom": "18.3.1",
     "@leather.io/rpc": "2.1.21",
     "@vitest/coverage-v8": "2.1.3",
-    "@vue/runtime-core": "3.4.3",
     "@wallet-standard/features": "1.0.3",
     "@walletconnect/types": "2.17.2",
-    "react": "18.3.1",
-    "react-dom": "18.3.1",
-    "vitest": "2.1.3",
-    "vue": "3.x"
+    "vitest": "2.1.3"
   },
   "author": "Reown <support@reown.com> (https://reown.com)",
   "license": "Apache-2.0",
--- a/packages/adapters/bitcoin/package.json
+++ b/packages/adapters/bitcoin/package.json
@@ -21,6 +21,26 @@
       "types": "./dist/types/exports/index.d.ts",
       "import": "./dist/esm/exports/index.js",
       "default": "./dist/esm/exports/index.js"
+    },
+    "./react": {
+      "types": "./dist/types/exports/react.d.ts",
+      "import": "./dist/esm/exports/react.js",
+      "default": "./dist/esm/exports/react.js"
+    },
+    "./vue": {
+      "types": "./dist/types/exports/vue.d.ts",
+      "import": "./dist/esm/exports/vue.js",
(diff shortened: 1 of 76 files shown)


packages/adapters/bitcoin/package.json
```diff
@@ -21,6 +21,26 @@
       "types": "./dist/types/exports/index.d.ts",
       "import": "./dist/esm/exports/index.js",
       "default": "./dist/esm/exports/index.js"
+    },
+    "./react": {
```
enesozturk, line 25: Revert these as we don't need to return react/vue related things from Bitcoin adapte r

packages/adapters/bitcoin/src/adapter.ts
```diff
@@ -139,6 +141,14 @@ export class BitcoinAdapter extends AdapterBlueprint<BitcoinConnector> {
     })
   }
 
+  public async getPublicKey(): Promise<string> {
```
enesozturk, line 144: I guess we don't need this any more

packages/adapters/bitcoin/src/connectors/LeatherConnector.ts
```diff
@@ -80,6 +80,11 @@ export class LeatherConnector extends SatsConnectConnector {
     }
   }
 
+  public override async getPublicKey(): Promise<string> {
```
enesozturk, line 83: Same for this and other connectors

packages/core/src/utils/TypeUtil.ts
```diff
@@ -885,6 +885,8 @@ export type AccountTypeMap = {
     namespace: K
     address: string
     type: NamespaceTypeMap[K]
+    publicKey?: string
+    path?: string
```
enesozturk, line 889: @zoruka Do you think we could do this better? Maybe like separating type definitions specificially for Bitcoin and others, with this way the other adapters will return `publicKey` and `path` as `undefined`. Or maybe we could add JSDoc for devs?

tomiir: You can actually do something similar to what I did for namespace type map and `&` it. Else you could use generics

packages/adapters/bitcoin/src/connectors/OKXConnector.ts
```diff
@@ -52,10 +52,15 @@ export class OKXConnector extends ProviderEventEmitter implements BitcoinConnect
...
 
-    return accounts.map(account => ({
-      address: account,
-      purpose: 'payment'
-    }))
+    const accountList = await Promise.all(
+      accounts.map(async account => ({
+        address: account,
+        purpose: 'payment' as const,
+        publicKey: await this.wallet.getPublicKey()
```
tomiir, line 59: no path here?

enesozturk: Let me validate that with Felipe - I'm not quite sure what would the devs need or if they need `path` tbh. I'll take a look

tomiir: `path` would be the derivaion path used to get this account, should be exposed by wallet (if exposed)

tomiir (approved):

packages/core/src/utils/TypeUtil.ts
zoruka: I think we could have specific types for different chains something like:
```typescript
type Account = { address: string } & (
   {
    namespace: 'bip122'
    type: 'payment' | 'ordinals'
    publicKey?: string
    path?: string
   } |
  {
    namespace: 'eip155'
    type: 'smartaccount' | 'eoa'
   }
   // ...
)
```

tomiir: @zoruka that would cause typing issue later down the line
properly typing this would be either doing something like 
```
type Account<T extends ChainNamespace> = { address: string, namespace: T } & T extends 'bip122' ? (
   {
    namespace: 'bip122'
    type: 'payment' | 'ordinals'
    publicKey?: string
    path?: string
   } : ...
   ```

Or something like this. I know there's a way to do this so that we always have a specific type depending on the namespace

File: packages/adapters/bitcoin/src/connectors/OKXConnector.ts (hunk 5 above)
Comment (zoruka) [inline packages/adapters/bitcoin/src/connectors/OKXConnector.ts:59]: I guess this may be wrong if `getAccounts` brings more than one address then the publicKey will be different. `getPublicKey` i[n this case seems to get only the public key of the active account](https://www.okx.com/web3/build/docs/sdks/chains/bitcoin/provider#getpublickey). If we let it this way we probably don't need to call it for each address in the array.

File: packages/core/src/utils/TypeUtil.ts (hunk 4 above)
Comment (tomiir) [reply packages/core/src/utils/TypeUtil.ts:889]: I'd add a further level by saying like 
```
T extends 'bip122' ? { publicKey: string, path: string }
```

I know something like this is possible

tomiir (approved):

PR #3576, opened 2025-01-06
chore: add loading indicator to `<w3m-account-button>` component before balance is fetched

magiziz:
# Description

When the page has loaded for the first time we see `0.00` balance on the account button component which is not correct. We should have a loading indicator to let user know that we're currently fetching the balance.

## Type of change

# Associated Issues

For Linear issues: Closes APKT-1151

# Showcase (Optional)
Screenshot 2025-01-06 at 12 48 24

Screenshot 2025-01-06 at 12 47 58

# Checklist


packages/scaffold-ui/src/modal/w3m-account-button/index.ts
```diff
@@ -84,6 +84,7 @@ class W3mAccountButtonBase extends LitElement {
     }
 
     const shouldShowBalance = this.balance === 'show'
+    const shouldShowLoading = typeof this.balanceVal !== 'string'
```
magiziz, line 87: I tried having a separate loading state in `AccountController.state`, but when opening the modal and switching between different views like swap page view you'll always see loading indicator on the button which is not quite nice. This is happening because we're fetching the balance in different components which causes this behaviour.

For now during page reload or first time connection this would work fine for now 🙏

enesozturk: > you'll always see loading indicator on the button which is not quite nice

I had this issue many months ago and needed to reverted back - so confirming that this is better solution for now

svenvoskamp (approved):

tomiir (approved):

PR #3583, opened 2025-01-07
fix: error handling swap button

svenvoskamp:
# Description

In some cases when swapping with a high amount (most of time, more exotic tokens), the swap button could end up with an infinite spinner. This happens when `fetchSwapQuote` returns a `400`. This happens because the filled in amount exceeds the balance of the user. In this case `state.loadingQuote` will never be false and swap button will show an infinite spinner. 

The fix in this PR is to catch this error and show a correct error message.

## Type of change

# Associated Issues

For Linear issues: Closes APKT-850

# Checklist


packages/core/src/controllers/SwapController.ts
```diff
@@ -501,39 +501,44 @@ export const SwapController = {
...
 
-    const quoteToAmount = quoteResponse?.quotes?.[0]?.toAmount
+      const quoteToAmount = quoteResponse?.quotes?.[0]?.toAmount
 
-    if (!quoteToAmount) {
-      return
-    }
+      if (!quoteToAmount) {
+        return
+      }
```
magiziz, line 520: Do you think it'd be useful to show alert error using `AlertController.open` ?

packages/core/src/controllers/SwapController.ts
```diff
@@ -501,39 +501,44 @@ export const SwapController = {
...
+        userAddress: address,
+        projectId: OptionsController.state.projectId,
+        from: sourceToken.address,
+        to: toToken.address,
+        gasPrice: state.gasFee,
+        amount: amountDecimal.toString()
+      })
 
-    state.loadingQuote = false
+      state.loadingQuote = false
```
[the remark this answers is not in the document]
tomiir: it's an indent caused by adding the `try` block

packages/core/src/controllers/SwapController.ts
tomiir: won't this cause a no-response on the ui ?

tomiir (approved):

packages/core/src/controllers/SwapController.ts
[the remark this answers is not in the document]
svenvoskamp: Setting this to false stops the button to be in a loading state

packages/core/src/controllers/SwapController.ts
svenvoskamp: I will add an alert

tomiir (approved):

PR #3598, opened 2025-01-09
fix: expose `fetchBalance` method

svenvoskamp:
# Description

The balance in our account button doesn't update after a transaction has happened. This is because we don't have control when a transaction outside AppKit is happening. We tried to find a solution with querying for pending transactions. This resulted in clients exceeding their quota. 

This PR exposes the `fetchBalance` method. Apps can call this after a transaction happened in their app. This will update the correct balance in AppKit. 

## Type of change

# Associated Issues

For Linear issues: Closes APKT-1777

# Showcase (Optional)

# Checklist


packages/appkit/exports/vue.ts
```diff
@@ -61,4 +61,8 @@ export function useAppKitNetwork(): Ref<UseAppKitNetworkReturn> {
   return state
 }
 
+export async function useAppKitBalance() {
+  await modal?.fetchBalance()
+}
```
magiziz, line 66: Do you think it'd be better to do something like this ?

```ts
const { data, error, isSuccess, isError } = useAppKitBalance()
```

packages/appkit/src/client.ts
```diff
@@ -1646,6 +1656,8 @@ export class AppKit {
     chainId: string | number
     chainNamespace: ChainNamespace
   }) {
+    console.log('syncBalance', params)
```
magiziz, line 1659: Remove ?

lukaisailovic:
Logic in `react.ts` and `vue.ts` is exactly the same. Can we reuse it as in this case we're not using any framework specific features?

enesozturk:
@svenvoskamp What is the status with this? Can we merge soon? Ready to review?

packages/appkit/exports/react.ts
```diff
@@ -60,3 +61,11 @@ export function useAppKitNetwork(): UseAppKitNetworkReturn {
 }
 
 export { useAppKitAccount } from '@reown/appkit-core/react'
+
+export async function useAppKitBalance() {
+  if (!modal) {
+    throw new Error('AppKit not initialized. Please call createAppKit first.')
+  }
+
+  return await fetchBalance(modal)
```
magiziz, line 70: Do we need await ?

```ts
return fetchBalance(modal)
```

magiziz (approved):

arein (commented):
Do we need to add some rate limit for the internal logic here? i.e. it only calls this max every 10 seconds now matter how often it's triggered?

packages/appkit/exports/react.ts
enesozturk: Imo we should return this function in an object. Later there might be other properties or functions we might want to expose from this hook and we wouldn't want it to be breaking change. 

Also to follow the same convention with other hooks

```suggestion
 function fetchBalance() {
  return fetchBalance(modal)
 }
 
  return {
   fetchBalance
  }
```

enesozturk: Also if I'm not wrong they will send request for fetchBalance as soon as the they call this hook. They cannot control the fetch time with this way right?

```jsx

function MyComp () {
   const balance = useAppKitBalance() // this will immediately call the balance, which would be better to let responsiblity to the developers imo
...

}
```

tomiir: +1 to Enes' point. All of our hooks usually return handlers, not responses

packages/appkit/src/client/appkit-base-client.ts
```diff
@@ -1871,6 +1875,20 @@ export abstract class AppKitBaseClient {
...
+    const chainId = this.getCaipNetwork()?.id
+
+    if (address && chainNamespace && chainId) {
+      const balance = await this.updateNativeBalance(address, chainId, chainNamespace)
+
+      return balance
+    }
+
+    return undefined
+  }
```
tomiir, line 1890: do we need this ? why not rename `updateNativeBalance`?

tomiir (approved):

magiziz (approved):

PR #3607, opened 2025-01-09
feat: default account type option

zoruka:
# Description

Add `defaultAccountTypes` option for AppKit initialization.

## Type of change

# Associated Issues

For Linear issues: Closes APKT-1870
For GH issues: closes #...

# Showcase (Optional)

# Checklist


packages/appkit/src/utils/TypesUtil.ts
```diff
@@ -79,4 +79,9 @@ export type AppKitOptions = {
    * @default undefined
    */
   universalProvider?: UniversalProvider
-} & OptionsControllerState
+  /**
+   * The default account type used for each chain namespace.
+   * @default "{ bip122: 'payment', eip155: 'smartAccount', polkadot: 'eoa', solana: 'eoa' }"
+   */
+  defaultAccountTypes?: Partial<OptionsControllerState['defaultAccountTypes']>
+} & Omit<OptionsControllerState, 'defaultAccountTypes'>
```
tomiir, line 87: does this make sense given SAs are only on EVM?

zoruka: but we got account types for all the chain namespaces, this is mainly to avoid rework in the future thinking about a multichain way

packages/adapters/ethers/src/client.ts
```diff
@@ -346,7 +347,8 @@ export class EthersAdapter extends AdapterBlueprint {
 
     if (type === 'AUTH') {
       const { address } = await (selectedProvider as unknown as W3mFrameProvider).connect({
-        chainId
+        chainId,
+        preferredAccountType: OptionsController.state.defaultAccountTypes.eip155
```
enesozturk, line 351: Do we already setting the `OptionsController` before this line is getting called? Might there be race condition that we are initializing adapters before initializing the controllers? Lets just double check, otherwise if possible we could directly use this value from props rather than state?

packages/appkit/src/client.ts
```diff
@@ -1242,14 +1243,15 @@ export class AppKit {
...
+      const preferredAccountType = (user.preferredAccountType ||
+        OptionsController.state.defaultAccountTypes[namespace]) as W3mFrameTypes.AccountType
       this.setPreferredAccountType(preferredAccountType, namespace)
 
       const userAccounts = user.accounts?.map(account =>
         CoreHelperUtil.createAccount(
           namespace,
           account.address,
-          namespace === ConstantsUtil.CHAIN.EVM ? account.type : 'eoa'
+          account.type || OptionsController.state.defaultAccountTypes[namespace]
```
enesozturk, line 1254: Why not using preferredAccountType variable here?

```suggestion
          account.type || preferredAccountType

tomiir (approved):

enesozturk (approved):

packages/appkit/src/client.ts
zoruka: That's this way because here it is mapping the `accounts` from `user`, which seems to may be different from `user.preferredAccountType`

PR #3619, opened 2025-01-13
refactor: clean up and reuse wallet connect connector

zoruka:
# Description

## Type of change

# Associated Issues

For GH issues: closes #...

# Showcase (Optional)

# Checklist

### DIFF
--- a/packages/adapters/wagmi/src/client.ts
+++ b/packages/adapters/wagmi/src/client.ts
@@ -58,6 +58,7 @@ import type { W3mFrameProvider } from '@reown/appkit-wallet'
 import { normalize } from 'viem/ens'
 import { parseWalletCapabilities } from './utils/helpers.js'
 import { LimitterUtil } from './utils/LimitterUtil.js'
+import { WalletConnectConnector } from '@reown/appkit/connectors'
 
 interface PendingTransactionsFilter {
   enable: boolean
@@ -471,7 +472,19 @@ export class WagmiAdapter extends AdapterBlueprint {
     }
   }
 
-  public async connectWalletConnect(onUri: (uri: string) => void, chainId?: number | string) {
+  public override async connectWalletConnect(
+    onUri: (uri: string) => void,
+    chainId?: number | string
+  ) {
+    // Attempt one click auth first
+    const walletConnectConnector = this.getWalletConnectConnector()
+    const isAuthenticated = await walletConnectConnector.authenticate()
+
+    if (isAuthenticated) {
+      return { clientId: await walletConnectConnector.provider.client.core.crypto.getClientId() }
+    }
+
+    // Attempt to connect using wagmi connector
     const connector = this.getWagmiConnector('walletConnect')
 
     if (!connector) {
@@ -486,11 +499,11 @@ export class WagmiAdapter extends AdapterBlueprint {
       )
     }
 
-    provider.on('display_uri', (uri: string) => {
-      onUri(uri)
-    })
+    provider.on('display_uri', onUri)
 
     await connect(this.wagmiConfig, { connector, chainId: chainId ? Number(chainId) : undefined })
+
+    return { clientId: await provider.client.core.crypto.getClientId() }
   }
 
   public async connect(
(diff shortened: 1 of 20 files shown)


packages/appkit/src/adapters/ChainAdapterBlueprint.ts
```diff
@@ -90,17 +92,7 @@ export abstract class AdapterBlueprint<
...
-      id: CommonConstantsUtil.CONNECTOR_ID.WALLET_CONNECT,
-      type: 'WALLET_CONNECT',
-      name: PresetsUtil.ConnectorNamesMap[CommonConstantsUtil.CONNECTOR_ID.WALLET_CONNECT],
-      provider: universalProvider,
-      imageId: PresetsUtil.ConnectorImageIds[CommonConstantsUtil.CONNECTOR_ID.WALLET_CONNECT],
-      chain: this.namespace,
-      chains: []
-    } as unknown as Connector)
-  }
+  public abstract setUniversalProvider(universalProvider: UniversalProvider): void
```
tomiir, line 95: why do we need this as abstract ? From what I've seen all implementations do the same thing?

packages/adapters/wagmi/src/client.ts
```diff
@@ -471,7 +472,19 @@ export class WagmiAdapter extends AdapterBlueprint {
...
+  ) {
+    // Attempt one click auth first
+    const walletConnectConnector = this.getWalletConnectConnector()
+    const isAuthenticated = await walletConnectConnector.authenticate()
+
+    if (isAuthenticated) {
+      return { clientId: await walletConnectConnector.provider.client.core.crypto.getClientId() }
+    }
+
+    // Attempt to connect using wagmi connector
```
tomiir, line 487: for a diff pr: 
should we attempt to remove this method? shouldn't the WC connector know how to handle this and not each adapter?

packages/appkit/src/adapters/ChainAdapterBlueprint.ts
```diff
@@ -374,6 +372,18 @@ export abstract class AdapterBlueprint<
   public abstract revokePermissions(
     params: AdapterBlueprint.RevokePermissionsParams
   ): Promise<`0x${string}`>
+
+  protected getWalletConnectConnector(): WalletConnectConnector {
+    const connector = this.connectors.find(c => c instanceof WalletConnectConnector) as
+      | WalletConnectConnector
+      | undefined
```
tomiir, line 379: I don't think you ever return undefined (you throw first)

packages/appkit/src/adapters/ChainAdapterBlueprint.ts
zoruka: It is abstract because `Connector` is generic and the adapter must add the connector that extends the correct `Connector` interface, take a look on how Solana and Bitcoin have this method implemented

packages/adapters/wagmi/src/client.ts
zoruka: That's already the attempt, but some of the adapters may have specific requirement, like for wagmi it goes through this other specific flow calling `connect` from wagmi, so we can override.

packages/appkit/src/adapters/ChainAdapterBlueprint.ts
zoruka: but `connector` object may be undefined as the result of `find` method

packages/appkit/src/connectors/WalletConnectConnector.ts
```diff
@@ -0,0 +1,108 @@
...
+  public readonly type = 'WALLET_CONNECT'
+  public readonly imageId = PresetsUtil.ConnectorImageIds[ConstantsUtil.CONNECTOR_ID.WALLET_CONNECT]
+  public readonly chain: Namespace
+
+  public provider: UniversalProvider
+
+  protected caipNetworks: CaipNetwork[]
+
+  constructor({ provider, caipNetworks, namespace }: WalletConnectConnector.Options<Namespace>) {
+    this.caipNetworks = caipNetworks
```
tomiir, line 25: I've seen you populate this with namespace scoped chains. Won't that have issues? WC session has all of the namespaces combined

zoruka: `namespace` is a mandatory key of `Connector` interface. I don't think it will have a problem because still each adapter have a `WalletConnectConnect` with its own namespace. If this would be a problem than we would need to set a kind of `Connector` that combines namespaces.

tomiir: I mean we'd need to triple check how chain switching is working. I'd assume that my WC connector has all the namespaces available not jsut the ones scoped to the current adapter (since I can switch to a diff namespace=

enesozturk (approved):

PR #3681, opened 2025-01-17
fix: redundant goBack call when switching to another namespace

enesozturk:
# Description

- Refactors `onNewNetwork` and `onNewAddress` methods to handle unnecessary `RouterController.goBack()` calls
- Refactors `onSwitchNetwork` method and few other parts to show `SwitchActiveChain` page when users trying to switch to a network from different namespace, and filter connectors with that namespace

### Steps to repro

- Go to multichain example
- Connect to a namespace
- Switch network
- Connect to another namespace

or

- Go to multichain example
- Connect to a namespace
- Switch network but don't connect, close modal
- Switch to network that previously connected

Try several flows like this

## Type of change

# Associated Issues

For Linear issues: Closes APKT-2040
For GH issues: closes #...

# Showcase (Optional)

# Checklist

### DIFF
--- a/packages/appkit/src/client.ts
+++ b/packages/appkit/src/client.ts
@@ -1538,7 +1538,8 @@ export class AppKit {
 
     const { chainId: activeChainId } = StorageUtil.getActiveNetworkProps()
     const chainIdToUse = chainId || activeChainId
-    const isUnsupportedNetwork = ChainController.state.activeCaipNetwork?.name === 'Unknown Network'
+    const isUnsupportedNetwork =
+      ChainController.state.activeCaipNetwork?.name === ConstantsUtil.UNSUPPORTED_NETWORK_NAME
     const shouldSupportAllNetworks = ChainController.getNetworkProp(
       'supportsAllNetworks',
       chainNamespace
@@ -2018,7 +2019,7 @@ export class AppKit {
     return {
       id: caipNetworkId.split(':')[1],
       caipNetworkId,
-      name: 'Unknown Network',
+      name: ConstantsUtil.UNSUPPORTED_NETWORK_NAME,
       chainNamespace: caipNetworkId.split(':')[0],
       nativeCurrency: {
         name: '',
--- a/packages/appkit/src/client.ts
+++ b/packages/appkit/src/client.ts
@@ -787,22 +787,8 @@ export class AppKit {
     const namespace = this.getActiveChainNamespace()
 
     if (namespace) {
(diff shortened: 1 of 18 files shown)


packages/appkit/src/client.ts
```diff
@@ -1552,7 +1538,11 @@ export class AppKit {
 
     const { chainId: activeChainId } = StorageUtil.getActiveNetworkProps()
     const chainIdToUse = chainId || activeChainId
-
+    const isUnsupportedNetwork = ChainController.state.activeCaipNetwork?.name === 'Unknown Network'
```
magiziz, line 1541: Can we put `Unknown Network` in a constants file ?

packages/appkit/src/client.ts
```diff
@@ -2022,16 +2015,39 @@ export class AppKit {
...
+        symbol: ''
+      },
+      rpcUrls: {
+        default: {
+          http: []
+        }
+      }
+    } as CaipNetwork
+  }
+
```
tomiir, line 2036: this does not require internal AppKit state, can we move to utils?

packages/scaffold-ui/src/views/w3m-switch-active-chain-view/index.ts
```diff
@@ -70,28 +70,32 @@ export class W3mSwitchActiveChainView extends LitElement {
...
     }
 
+    ChainController.setIsSwitchingNamespace(true)
+
     if (this.caipNetwork) {
-      await ChainController.switchActiveNetwork(this.caipNetwork)
+      ChainController.setActiveCaipNetwork(this.caipNetwork)
     } else {
       ChainController.setActiveNamespace(this.switchToChain)
     }
```
tomiir, line 96: should we encapsulate this in ChainController directly? I think `setActiveCaipNetwork`  should set the isSwitchingNamespace flag directly

magiziz (approved):

packages/scaffold-ui/src/views/w3m-switch-active-chain-view/index.ts
enesozturk: Totally, and it already has that call inside, removing here

packages/appkit/src/client.ts
enesozturk: We need to set some values dynamically like `id`, `caipNetworkId`, `name` and `chainNamespace`, so I'd prefer to keep rest here as well

PR #3724, opened 2025-01-22
feat: network configuration on demo

enesozturk:
# Description

Introduces network configuration feature on Demo app

## Type of change

# Associated Issues

For GH issues: closes #...

# Showcase (Optional)

# Checklist


apps/builder/components/network-list.tsx
```diff
@@ -0,0 +1,86 @@
...
+                if (enabledNetworks.includes(n.network.id)) {
+                  if (enabledNetworks.length > 1) {
+                    removeNetwork(n)
+                  }
+                } else {
+                  addNetwork(n)
+                }
+              }}
+              name={n.network.name}
+            />
```
tomiir, line 61: can we remove inline def?

packages/core/src/controllers/ChainController.ts
```diff
@@ -187,6 +187,36 @@ export const ChainController = {
     )
   },
 
+  addNetwork(network: CaipNetwork) {
+    const chainAdapter = state.chains.get(network.chainNamespace)
+    if (chainAdapter) {
+      chainAdapter.caipNetworks = [...(chainAdapter.caipNetworks || []), network]
```
tomiir, line 193: you should do a Set here to prevent duplications

packages/core/src/controllers/ChainController.ts
```diff
@@ -187,6 +187,36 @@ export const ChainController = {
...
+
+      // Filter out the network being removed
+      chainAdapter.caipNetworks = chainAdapter.caipNetworks?.filter(
+        network => network.id !== networkId
+      )
+
+      // If active network was removed and there are other networks available, switch to first one
+      if (isActiveNetwork && chainAdapter?.caipNetworks?.[0]) {
+        this.setActiveCaipNetwork(chainAdapter.caipNetworks[0])
+      }
```
tomiir, line 213: what happens when there are no other networks? We should prevent removing the only network avail .

apps/builder/components/configuration-sections/section-connect-options.tsx
```diff
@@ -26,9 +27,30 @@ const SortableConnectMethodList = dynamic(
...
 
 const ChainList = dynamic(() => import('@/components/chain-list').then(mod => mod.ChainList), {
-  ssr: false
+  ssr: false,
+  loading: () => (
+    <div className="flex gap-2 flex-wrap">
+      <RoundOptionItemLoading />
+      <RoundOptionItemLoading />
+      <RoundOptionItemLoading />
+    </div>
```
magiziz, line 36: wdyt about if we add a component `<ItemLoaders />` where you can pass in a `total` prop instead of duplicating ?

apps/builder/lib/config.ts
```diff
@@ -85,29 +95,43 @@ const metadata = {
...
-const initialNetworks: AppKitNetwork[] = []
+// Enabled network object list
+let initialNetworks: AppKitNetwork[] = []
 
 initialEnabledChains.forEach(chain => {
   if (chain === 'eip155') {
-    initialNetworks.push(...evmNetworks)
+    const enabledNetworks = evmNetworks.filter(network =>
+      initialEnabledNetworks.includes(network.id)
+    )
```
magiziz, line 111: should we have some type of helper function that checks this for us ?

apps/builder/components/configuration-sections/section-connect-options.tsx
enesozturk: Better to not have additional component since they are three liners imo

apps/builder/components/ui/round-option-item.tsx
```diff
@@ -10,35 +10,47 @@ interface RoundOptionItemProps {
...
   disabled,
-  name
+  name,
+  size = 'md',
+  message
 }: RoundOptionItemProps) {
   return (
     <TooltipProvider>
-      <Tooltip delayDuration={500}>
+      <Tooltip delayDuration={0}>
```
magiziz, line 28: Can `delayDuration` be removed or does it have default delay duration already ?

apps/builder/lib/constants.ts
```diff
@@ -19,3 +36,45 @@ export const FONT_OPTIONS = [
...
+  network: AppKitNetwork
+}
+
+export const NETWORK_OPTIONS = [
+  { namespace: 'eip155', network: mainnet },
+  { namespace: 'eip155', network: optimism },
+  { namespace: 'eip155', network: bsc },
+  { namespace: 'eip155', network: polygon },
+  { namespace: 'eip155', network: avalanche },
+  { namespace: 'eip155', network: arbitrum },
```
magiziz, line 51: Can we do something like this ?

```ts
const NETWORK_OPTIONS = {
  eip155: [mainnet, optimism, polygon],
  solana: [solana, solanaDevnet]
}
```

enesozturk: When rendering the items this approach will require more complexity to JS operations:

```js
{NETWORK_OPTIONS.map(n => { render(..) }}
```

When using map:

```js
return Object.entries(NETWORK_OPTIONS_2).map(([namespace, networks]) => ( ... ));
```

enesozturk: Since it's safed namespace strings with TS, it's okay to go

apps/builder/components/ui/round-option-item.tsx
enesozturk: Default is `700`: https://www.radix-ui.com/primitives/docs/components/tooltip#root, it might feel be a bit slower. And `0` would be annoying they they try to hover over the networks

svenvoskamp (approved):

apps/builder/components/ui/round-option-item.tsx
enesozturk: Oh sorry I've made it 0 intentionally lol

PR #3752, opened 2025-01-28
refactor: lazy load coinbase package

enesozturk:
# Description

We have third party connector dependencies installed with AppKit which is increasing our package install size. Also it's been requested by some users that they wouldn't want that.

This PR introduces changes to may dynamic import for Coinbase Wallet SDK to let this choice to the users. Wagmi already doing dynamic import for Coinbase Wallet SDK internally. We are changing our business logics for Ethers/Ethers5 adapters.

## Type of change

# Associated Issues

For Linear issues: Closes APKT-2061
For GH issues: closes #...

# Showcase (Optional)

# Checklist


packages/appkit/src/client.ts
```diff
@@ -2027,7 +2027,7 @@ export class AppKit {
   private async initChainAdapter(namespace: ChainNamespace) {
     this.onConnectors(namespace)
     this.listenAdapter(namespace)
-    this.chainAdapters?.[namespace].syncConnectors(this.options, this)
+    await this.chainAdapters?.[namespace].syncConnectors(this.options, this)
```
magiziz, line 2030: What happens if coinbase sdk import fails ? Will it block other initialisation steps ?

packages/adapters/ethers/package.json
```diff
@@ -31,10 +31,10 @@
   },
   "peerDependencies": {
     "@ethersproject/sha2": "5.7.0",
-    "@coinbase/wallet-sdk": "4.2.4",
     "ethers": ">=6"
   },
   "devDependencies": {
+    "@coinbase/wallet-sdk": "4.2.4",
```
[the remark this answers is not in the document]
enesozturk: Yeah, we don't want to push people to install since that's the main idea of this PR (updating descs shortly) - but we sill need it as dev dep to make TS happy for dynamic imports

packages/appkit/src/client.ts
enesozturk: Good point, will wrap it with try catch - better

packages/adapters/ethers/package.json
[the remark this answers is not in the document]
tomiir: shouldn't it be marked as a peerDep then ?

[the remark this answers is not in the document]
shoom3301: p.s. It's not a problem in fact to keep the dependency in "dependencies".
The most important is to make sure that the library is loaded only on demand in browsers.

packages/adapters/ethers/src/client.ts
```diff
@@ -64,26 +60,27 @@ export class EthersAdapter extends AdapterBlueprint {
       return injectedProvider
     }
 
-    function getCoinbaseProvider() {
-      if (coinbaseProvider) {
-        return coinbaseProvider
-      }
+    async function getCoinbaseProvider() {
+      if (options.enableCoinbase) {
+        const { createCoinbaseWalletSDK } = await import('@coinbase/wallet-sdk')
```
shoom3301, line 65: Perfect!

tomiir (approved):

packages/adapters/bitcoin/src/adapter.ts
```diff
@@ -82,7 +82,8 @@ export class BitcoinAdapter extends AdapterBlueprint<BitcoinConnector> {
     }
   }
 
-  override syncConnectors(_options?: AppKitOptions, appKit?: AppKit): void {
+  // eslint-disable-next-line @typescript-eslint/require-await
```
magiziz, line 85: needed ?

packages/adapters/ethers/package.json
```diff
@@ -29,9 +29,11 @@
     "@walletconnect/universal-provider": "2.18.0",
     "valtio": "1.11.2"
   },
+  "optionalDependencies": {
+    "@coinbase/wallet-sdk": "4.2.4"
+  },
```
magiziz, line 34: Can we include it in `dependencies`, but lazy load them in adapter ?

packages/adapters/ethers/src/client.ts
```diff
@@ -64,26 +60,30 @@ export class EthersAdapter extends AdapterBlueprint {
...
+    async function getCoinbaseProvider() {
+      try {
+        const { createCoinbaseWalletSDK } = await import('@coinbase/wallet-sdk')
 
-      if (typeof window === 'undefined') {
-        return undefined
-      }
+        if (typeof window === 'undefined') {
+          return undefined
+        }
```
magiziz, line 69: Is it possible to use `CoreHelperUtil.isClient()` ?

packages/adapters/ethers/package.json
enesozturk: `optionalDependencies` will be installed by the package managers as well, the only difference is that if somehow there is an issue installing these packages, the project will proceed to installing others, not going to fail and this will not be used. 

So for third party 'optional' things, we should use `optionalDependencies`

packages/adapters/ethers/src/client.ts
enesozturk: Totally

PR #3760, opened 2025-01-29
chore: fetch wallet images only when modal is open

magiziz:
# Description

We now fetch wallet images when modal is open instead of fetching them during page load.

## Type of change

# Associated Issues

For GH issues: closes #...

# Showcase (Optional)

https://github.com/user-attachments/assets/3eec3102-d35a-49d9-a6e5-c35cc972c10b
 
# Checklist


packages/scaffold-ui/src/modal/w3m-modal/index.ts
```diff
@@ -44,7 +46,7 @@ export class W3mModal extends LitElement {
   public constructor() {
     super()
     this.initializeTheming()
-    ApiController.prefetch()
+    ApiController.prefetchAnalyticsConfig()
```
tomiir, line 49: do we need this here? @Cali93 can we remove ?

tomiir (approved):

packages/core/src/controllers/ApiController.ts
```diff
@@ -272,17 +272,21 @@ export const ApiController = {
...
-  prefetch() {
+  prefetchWalletImages() {
     const promises = [
       ApiController.fetchFeaturedWallets(),
       ApiController.fetchRecommendedWallets(),
       ApiController.fetchNetworkImages(),
       ApiController.fetchConnectorImages()
     ]
+
+    state.prefetchPromise = Promise.race([Promise.allSettled(promises)])
```
Cali93, line 283: Why do we need `Promise.race` ?

Cali93 (approved):
LGTM, not sure if we need to prefetch the analytics config but if I recall correctly there's was an edge case where we wouldn't sent the first event when analytics isn't specified in the options but analytics is enabled through Cloud.

packages/core/src/controllers/ApiController.ts
magiziz: oh seems like it was alwasy there, will remove it 🙏

enesozturk:
@magiziz There is this issue happening that when you don't open the model, the network button doesn't render the network image. 

Screenshot 2025-01-30 at 10 37 25

And when I open the networks list, this time it's fetching and updating ETH logo but the other networks are still empty.

Screenshot 2025-01-30 at 10 39 33

I'd propose going with an approach that fetching the network image when the item is visible on the screen. We can achieve this by using the `IntersectionOberserver` API and fetch the item inside of each network item / image component. This will also help to not fetch images if you don't scroll (if item is not visible on the screen). 

I've implemented this on "All Wallets" screen that you can take as a ref. We also have tech doc about it: https://www.notion.so/walletconnect/Image-rendering-optimization-f1f292ccecfe43deb62e30860ceef3f3?pvs=4

packages/core/src/controllers/ApiController.ts
```diff
@@ -272,17 +272,21 @@ export const ApiController = {
     await ApiController.fetchRecommendedWallets()
   },
 
-  prefetch() {
+  prefetchWalletImages() {
     const promises = [
       ApiController.fetchFeaturedWallets(),
       ApiController.fetchRecommendedWallets(),
       ApiController.fetchNetworkImages(),
```
tomiir, line 279: This needs to be called on init as well else network btton doesn't properly render. 
Maybe a cool optimizaton is loading this on the button render/init instead of here

packages/core/src/controllers/ApiController.ts
```diff
@@ -265,24 +270,20 @@ export const ApiController = {
...
       ApiController.fetchConnectorImages()
     ]
+
+    state.prefetchPromise = Promise.allSettled(promises)
+  },
+
+  async prefetchAnalyticsConfig() {
     if (OptionsController.state.features?.analytics) {
-      promises.push(ApiController.fetchAnalyticsConfig())
+      await ApiController.fetchAnalyticsConfig()
```
tomiir, line 285: do we need to await this?

tomiir (approved):

packages/core/src/controllers/ApiController.ts
magiziz: yeah we don't do await it

PR #3765, opened 2025-01-29
chore: reduced bundle size for `@reown/appkit-wallet`, added missing ui components and removed `process.env`

magiziz:
# Description

- We now have import `W3mFrameRpcConstants` from `@reown/appkit-wallet/utils`. This is because controllers in the core package was importing the whole `@reown/appkit-wallet` package which then imported zod and other unnecessary deps. We can now store simple constants and utils in `@reown/appkit-wallet/utils` to simplify things when it comes to bundle size.
- Added some missing ui components to composites
- Removed `process.env` from the code since it causes bundling conflicts with other frameworks / libraries like vite

## Type of change

# Associated Issues

For GH issues: closes #...

# Showcase (Optional)

**Before:**

before

**After:**

after

# Checklist


packages/appkit/exports/basic.ts
```diff
@@ -5,10 +5,8 @@ import type { AppKitOptions } from '../src/utils/TypesUtil.js'
 import { PACKAGE_VERSION } from './constants.js'
 
 // -- Utils & Other -----------------------------------------------------
-export * from '../src/utils/index.js'
 export type * from '@reown/appkit-core'
 export type { CaipNetwork, CaipAddress, CaipNetworkId } from '@reown/appkit-common'
-export { CoreHelperUtil, AccountController } from '@reown/appkit-core'
```
magiziz, line 11: I think we don't need these so removed them

packages/appkit/src/adapters/ChainAdapterBlueprint.ts
```diff
@@ -16,7 +16,7 @@ import {
   type WriteContractArgs
 } from '@reown/appkit-core'
 import { PresetsUtil } from '@reown/appkit-utils'
-import { W3mFrameProvider } from '@reown/appkit-wallet'
+import type { W3mFrameProvider } from '@reown/appkit-wallet'
```
magiziz, line 19: Btw if you do this it'll be included in the bundle

```ts
import { type W3mFrameProvider } from '@reown/appkit-wallet'
```

So we'd need to do this 

```ts
import type { W3mFrameProvider } from '@reown/appkit-wallet'
```

packages/core/src/controllers/OnRampController.ts
```diff
@@ -104,8 +100,7 @@ export const OnRampController = {
 
   setSelectedProvider(provider: OnRampProvider | null) {
     if (provider && provider.name === 'meld') {
-      const pubKey =
-        process.env['NODE_ENV'] === 'production' ? MELD_PROD_PUBLIC_KEY : MELD_DEV_PUBLIC_KEY
+      const pubKey = MELD_PROD_PUBLIC_KEY
```
magiziz, line 103: Removed `process.env` since it causes issues if appkit is being lazy loaded. We can always change values when doing dev stuff.

packages/appkit/exports/basic.ts
tomiir: yea we should only export what's really necessary,

packages/appkit/src/client/appkit.ts
```diff
@@ -21,12 +21,8 @@ import {
   ThemeController
 } from '@reown/appkit-core'
 import { ErrorUtil, HelpersUtil, ConstantsUtil as UtilConstantsUtil } from '@reown/appkit-utils'
-import {
-  W3mFrameHelpers,
-  W3mFrameProvider,
-  W3mFrameRpcConstants,
-  type W3mFrameTypes
-} from '@reown/appkit-wallet'
+import { W3mFrameHelpers, W3mFrameProvider, type W3mFrameTypes } from '@reown/appkit-wallet'
```
tomiir, line 24: import type correctly 🙏

packages/core/src/controllers/OnRampController.ts
tomiir: hmmmm not sure I like this to be honest. What's the issue with lazy loading and envs?¿

magiziz: When you do lazy import with vite which we're currently doing in ethereum provider, it throws a bundle error where it says "can't read properties of `process....` because it expects `import.meta.env`.

Can be tricky to resolve, but can be a broader issue for other bundlers. I'd say we can remove those for now, but lmk if you want to revert the change.

tomiir: lets merge but I think there's a workaround here

tomiir (approved):

PR #3767, opened 2025-01-30
chore: lazy load svg icons

magiziz:
# Description

Our `<wui-icon>` is bundling all of the svg icons which can increase the bundle size. Made a simple refactor where we lazy load the icons and also include caching so we don't re-fetch them when not needed.

## Type of change

# Associated Issues

For GH issues: closes #...

# Checklist

### DIFF
--- a/packages/ui/src/components/wui-icon/index.ts
+++ b/packages/ui/src/components/wui-icon/index.ts
@@ -1,167 +1,118 @@
 import type { TemplateResult } from 'lit'
 import { LitElement, html } from 'lit'
 import { property } from 'lit/decorators.js'
+import { until } from 'lit/directives/until.js'
 
-// -- Svg's-------------------------------- //
-import { addSvg } from '../../assets/svg/add.js'
-import { allWalletsSvg } from '../../assets/svg/all-wallets.js'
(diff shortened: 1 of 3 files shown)


packages/ui/src/components/wui-icon/index.ts
```diff
@@ -182,7 +133,7 @@ export class WuiIcon extends LitElement {
       --local-width: ${`var(--wui-icon-size-${this.size});`}
     `
 
-    return html`${svgOptions[this.name]}`
+    return html`${until(getSvg(this.name), html`<div></div>`)}`
```
magiziz, line 136: I'm just returning an empty div for now while the svg load

tomiir: does it have the same class / respect size so that the screen does not shift after load=

magiziz: Mhmm good point, it doesn't. We'd somehow need to know the svg height and width so we can mock the styles

magiziz: Good catch! Just pushed a change where we set the correct size while svg's load 🙏

tomiir (approved):
smart man does smart things

PR #3774, opened 2025-01-31
fix: appkit basic screens

magiziz:
# Description

When using basic we need to show the correct screens. For instance we don't show all wallets screen on mobile and instead we show connect screen which has a weird UI.


packages/core/src/controllers/ChainController.ts
```diff
@@ -289,7 +289,11 @@ export const ChainController = {
...
     const isSupported = this.checkIfSupportedNetwork(caipNetwork.chainNamespace)
 
-    if (!isSupported && !OptionsController.state.allowUnsupportedChain) {
+    if (
+      !isSupported &&
+      !OptionsController.state.allowUnsupportedChain &&
+      !ConnectionController.state.wcBasic
+    ) {
       this.showUnsupportedChainUI()
     }
```
magiziz, line 298: Try connecting to MetaMask and you'll see `UnsupportedChain` modal view.

magiziz: Screenshot 2025-01-31 at 11 33 07

packages/core/src/controllers/ModalController.ts
```diff
@@ -54,8 +54,12 @@ export const ModalController = {
...
     } else if (caipAddress) {
       RouterController.reset('Account')
-    } else if (noAdapters && !CoreHelperUtil.isMobile()) {
-      RouterController.reset('ConnectingWalletConnectBasic')
+    } else if (noAdapters) {
+      if (CoreHelperUtil.isMobile()) {
+        RouterController.reset('AllWallets')
+      } else {
+        RouterController.reset('ConnectingWalletConnectBasic')
+      }
```
magiziz, line 62: Before if you were on mobile we showed this screen

Screenshot 2025-01-31 at 11 33 03

I know we can set custom connectors, but maybe we can show all wallets on mobile instead ?

packages/appkit/src/utils/TypesUtil.ts
```diff
@@ -10,6 +10,11 @@ export type AppKitOptions = {
    * @default []
    */
   adapters?: ChainAdapter[]
+  /**
+   * Used for ethereum provider
+   * @default false
+   */
+  basic?: boolean
```
tomiir, line 17: why?

packages/core/src/controllers/ModalController.ts
tomiir: wym by all walletS? I mean the all wallets button should be there

tomiir: and main wallets should surface to the main modal. I think we are missing some ui components on the basic side and that's why they are not rendering? 
What does DOM say

packages/core/src/controllers/ChainController.ts
tomiir: nice catch 💪

packages/appkit/src/utils/TypesUtil.ts
magiziz: I needed this so we can check internally to not show "Switch Network" modal if user is not in the supported network. This is only used internally and won't be exposed to users.

tomiir: I'd write a bit more context on the comment else ok

tomiir (approved):

PR #3786, opened 2025-02-04
fix: correctly switch chains for MM in ak basic

svenvoskamp:
# Description

UniversalProvider returns a session with the approved chains from the specific wallet. Most of these wallets return the whole array of networks. This isn't the case for MM, they only return the active network. https://docs.reown.com/advanced/providers/universal#chain-switching doesn't work for MM. This PR adds logic to switch networks, so that this active network will change accordingly. 

## Type of change

- [x]  Bug fix (non-breaking change which fixes an issue)

# Associated Issues

For Linear issues: Closes APKT-2175
For GH issues: closes #...

# Showcase (Optional)

# Checklist


packages/appkit/src/universal-adapter/client.ts
```diff
@@ -179,6 +181,44 @@ export class UniversalAdapter extends AdapterBlueprint {
...
+                  blockExplorerUrls: [caipNetwork.blockExplorers?.default.url]
+                }
+              ]
+            })
+          } catch (error) {
+            throw new Error('Chain is not supported')
+          }
+        }
+      }
+    }
```
tomiir, line 221: for the future, this looks repeated in the adapters no? maybe there's a way to abstract these. food for thought

packages/appkit/src/universal-adapter/client.ts
```diff
@@ -179,6 +181,44 @@ export class UniversalAdapter extends AdapterBlueprint {
...
+              params: [
+                {
+                  chainId: toHex(caipNetwork.id),
+                  rpcUrls: [caipNetwork?.rpcUrls['chainDefault']?.http],
+                  chainName: caipNetwork.name,
+                  nativeCurrency: {
+                    name: caipNetwork.nativeCurrency.name,
+                    decimals: caipNetwork.nativeCurrency.decimals,
+                    symbol: caipNetwork.nativeCurrency.symbol
+                  },
```
tomiir, line 211: doesn't `caipNetwork.nativeCurrency` work=?

tomiir (approved):

packages/appkit/src/universal-adapter/client.ts
svenvoskamp: This is reused in the ethers client. So agree to think about abstracting this to a util or smth

PR #3827, opened 2025-02-10
adding wallet get assets usage on appkit send

KannuSingh:
# Description

Enhance Send Flow with `wallet_getAssets` for CA Balance Display

This PR integrates the `wallet_getAssets` method into the Send Flow to display the user's Chain Abstraction (CA) balance. Since the embedded wallet supports CA for USDC and USDT and implements the wallet_getAssets RPC (indicated by the assetDiscovery capability), the AppKit Send Flow now:

Checks the connected chain's namespace and the wallet's capabilities.
If assetDiscovery is supported, it calls wallet_getAssets to fetch and display token balances.
Otherwise, it falls back to blockchainApi for retrieving account balances.

## Type of change

# Associated Issues

For Linear issues: Closes PAY-3
For GH issues: closes #...

# Showcase (Optional)

https://github.com/user-attachments/assets/e8abf331-978c-483d-8585-5f7e2108eeb5

# Checklist


lukaisailovic:
@KannuSingh Can we add the Showcase to the PR? Might be easier to understand what you did. Also the description is not very useful. It explains what you did but not why.

packages/adapters/wagmi/src/client.ts
```diff
@@ -707,6 +708,31 @@ export class WagmiAdapter extends AdapterBlueprint {
...
   }
 
+  public async walletGetAssets(
+    params: AdapterBlueprint.WalletGetAssetsParams
+  ): Promise<AdapterBlueprint.WalletGetAssetsResponse> {
+    if (!this.wagmiConfig) {
+      throw new Error('connectionControllerClient:walletGetAssets - wagmiConfig is undefined')
+    }
+
+    const connections = getConnections(this.wagmiConfig)
```
tomiir, line 718: don't you have the connector id in params?

packages/core/src/controllers/SendController.ts
```diff
@@ -162,14 +166,52 @@ export const SendController = {
...
+    return []
+  },
+
+  fetchNetworkBalance() {
+    if (state.tokenBalances.length === 0) {
+      return
+    }
 
-    if (!balances) {
+    const networkTokenBalances = SendApiUtil.mapBalancesToSwapTokens(state.tokenBalances)
```
tomiir, line 209: why do you need this?

packages/core/src/controllers/SendController.ts
```diff
@@ -162,14 +166,52 @@ export const SendController = {
...
+      state.loading = false
+    }
+
+    return []
+  },
+
+  fetchNetworkBalance() {
+    if (state.tokenBalances.length === 0) {
+      return
+    }
```
tomiir, line 207: don't we already have this ?

packages/scaffold-ui/src/views/w3m-wallet-send-select-token-view/index.ts
```diff
@@ -34,8 +33,8 @@ export class W3mSendSelectTokenView extends LitElement {
     super()
     this.unsubscribe.push(
       ...[
-        AccountController.subscribe(val => {
-          this.tokenBalance = val.tokenBalance
+        SendController.subscribe(val => {
+          this.tokenBalances = val.tokenBalances
```
tomiir, line 37: hm. why move the state to SendController? I mean the tokens would also be required on portfolio in some cases so I'd think it's better to keep them in the account state

KannuSingh: >why move the state to SendController?

added this state to SendController bc this hold the tokens balance retrieved via `wallet_getAssets` rpc (CA balances). The AccountController holds the token balances from connected chain only.

>I mean the tokens would also be required on portfolio in some cases

yes, but right now we don't have design to show both the balances nicely

packages/core/src/controllers/SendController.ts
KannuSingh: We have a similar method in AccountController, but currently, we're maintaining separate token balance lists—one for CA balances (`wallet_getAssets`) in SendController and another for active chain balances in AccountController.

This method is called only when Send UI is opened.

packages/adapters/wagmi/src/client.ts
KannuSingh: I could be wrong, but params should only contain the required parameters for the `wallet_getAssets` call, right?

packages/core/src/controllers/SendController.ts
KannuSingh: Previously, the implementation retrieved balances using:
```
const balances = await SwapApiUtil.getMyTokensWithBalance();
```
Inside SwapApiUtil.getMyTokensWithBalance(), it first fetched balances from BlockchainApi, then mapped them using mapBalancesToSwapTokens before returning the result.

In the current implementation, since SendController already maintains the token balance state, I’m now directly using it to call mapBalancesToSwapTokens, avoiding the API call.

packages/scaffold-ui/src/views/w3m-wallet-send-select-token-view/index.ts
tomiir: @lukaisailovic food for thought but aren't portolio api and the CA api related? are we using diff providers? 
feels kind of weird having to mantain 2 pieces of code / reconciliate, seems like they are doing basically the same?

lukaisailovic: I think we can utilize 7811 in portfolio if we wanted. I think in this case, we need to keep the portfolio call separate (inside the secure site) in order to have 1 place where we have the real user balance (no overrides)

packages/adapters/bitcoin/src/adapter.ts
```diff
@@ -256,6 +256,11 @@ export class BitcoinAdapter extends AdapterBlueprint<BitcoinConnector> {
     return Promise.resolve('0x')
   }
 
+  public override async walletGetAssets(
+    _params: AdapterBlueprint.WalletGetAssetsParams
```
enesozturk, line 260:
```suggestion
    params: AdapterBlueprint.WalletGetAssetsParams
```

Tiny but just to keep it same across adapters

packages/adapters/ethers/src/client.ts
```diff
@@ -642,4 +642,19 @@ export class EthersAdapter extends AdapterBlueprint {
 
     return await provider.request({ method: 'wallet_revokePermissions', params: [params] })
   }
+
+  public async walletGetAssets(
+    params: AdapterBlueprint.WalletGetAssetsParams
+  ): Promise<AdapterBlueprint.WalletGetAssetsResponse> {
+    const provider = ProviderUtil.getProvider(CommonConstantsUtil.CHAIN.EVM)
+
+    if (!provider) {
```
enesozturk, line 651: Do we handle the provider undefined state on the UI side somehow?

packages/adapters/solana/src/client.ts
```diff
@@ -108,6 +108,12 @@ export class SolanaAdapter extends AdapterBlueprint<SolanaProvider> {
     return Promise.resolve('0x')
   }
 
+  public override async walletGetAssets(
+    _params: AdapterBlueprint.WalletGetAssetsParams
```
enesozturk, line 112: Same here (if `params` not conflicting with another varialbe)

packages/adapters/wagmi/src/client.ts
```diff
@@ -364,7 +364,8 @@ export class WagmiAdapter extends AdapterBlueprint {
       account: data.fromAddress,
       abi: data.abi,
       functionName: data.method,
-      args: data.args
+      args: data.args,
+      __mode: 'prepared'
```
enesozturk, line 368: Why this has `__` prefix?

packages/appkit/src/client.ts
```diff
@@ -1106,6 +1106,11 @@ export class AppKit {
         }
 
         return '0x'
+      },
+      walletGetAssets: async (params: object | readonly unknown[]) => {
+        const adapter = this.getAdapter(ChainController.state.activeChain as ChainNamespace)
```
enesozturk, line 1111: We shouldn't cast the `activeChain` to `ChainNamespace` as it's bad practice (we are doing this also in many places, I'll add ticket for tech debt). The `ChainController.state.activeChain` can be undefined and this undefined should be handled in `getAdapter`

packages/core/src/controllers/SendController.ts
```diff
@@ -5,8 +5,9 @@ import { type Balance, type CaipAddress, NumberUtil } from '@reown/appkit-common
 import { ContractUtil } from '@reown/appkit-common'
 import { W3mFrameRpcConstants } from '@reown/appkit-wallet'
 
+import { ConstantsUtil } from '../utils/ConstantsUtil.js'
 import { CoreHelperUtil } from '../utils/CoreHelperUtil.js'
-import { SwapApiUtil } from '../utils/SwapApiUtil.js'
+import { SendApiUtil } from '../utils/SendApiUtil.js'
```
enesozturk, line 10: Are we sure this is not going to break anything on the send flows?

packages/core/src/utils/ERC7811Util.ts
```diff
@@ -0,0 +1,105 @@
...
+
+export type WalletGetAssetsResponse = Record<Hex, Asset[]>
+
+export const ERC7811Utils = {
+  /**
+   * Creates a Balance object from an ERC7811 Asset object
+   * @param asset - Asset object to convert
+   * @param chainId - Chain ID in CAIP-2 format
+   * @returns Balance object
+   */
```
enesozturk, line 37: 👌🏽

packages/adapters/bitcoin/src/adapter.ts
KannuSingh: @enesozturk  for all the methods in this adapter where the params are not used inside the method it is marked with `_params` to avoid eslint:error 

```params' is defined but never used. Allowed unused args must match /^_/u```

packages/adapters/solana/src/client.ts
KannuSingh: @enesozturk same as above comment

tomiir (approved):

PR #3829, opened 2025-02-10
fix: sign message not working in basic example

magiziz:
# Description

Fixed an issue where bifrost wallet signing wasn't working

### DIFF
--- a/examples/html-ak-basic/src/main.js
+++ b/examples/html-ak-basic/src/main.js
@@ -137,7 +137,7 @@ async function getPayload() {
     },
     eip155: {
       method: 'personal_sign',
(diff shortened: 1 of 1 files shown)


examples/html-ak-basic/src/main.js
```diff
@@ -194,7 +194,7 @@ async function signMessage() {
     if (!payload) {
       throw Error('Chain not supported by laboratory')
     }
-
+    console.log(payload, networkState?.caipNetwork?.caipNetworkId)
```
tomiir, line 197: remove
```suggestion
```

tomiir (approved):

PR #3835, opened 2025-02-10
fix: cannot reconnect after cancelling modal

magiziz:
# Description

Fixed an issue where if you open the modal and cancel right after you can't open again

## Type of change

# Associated Issues

For Linear issues: Closes APKT-2245

### DIFF
--- a/examples/html-ak-basic-up/src/main.js
+++ b/examples/html-ak-basic-up/src/main.js
@@ -117,13 +117,11 @@ async function initializeApp() {
     updateDom()
   })
 
-  provider.on('display_uri', uri => {
-    modal.open({ uri, view: 'ConnectingWalletConnectBasic' })
-  })
-
   // Button handlers
   document.getElementById('connect')?.addEventListener('click', async () => {
     setLoading(true)
+    await modal.open()
+    modal.subscribeEvents(({ data }) => data.event === 'MODAL_CLOSE' && setLoading(false))
     await provider.connect({ optionalNamespaces: OPTIONAL_NAMESPACES })
     updateDom()
     setLoading(false)
--- a/examples/html-ak-basic-up/src/main.js
+++ b/examples/html-ak-basic-up/src/main.js
@@ -121,7 +121,11 @@ async function initializeApp() {
   document.getElementById('connect')?.addEventListener('click', async () => {
     setLoading(true)
     await modal.open()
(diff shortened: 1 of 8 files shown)


packages/appkit/src/client/appkit-basic.ts
```diff
@@ -16,7 +16,7 @@ export { AccountController }
 
 // -- Types --------------------------------------------------------------------
 export interface OpenOptions {
-  view:
+  view?:
```
magiziz, line 19: Made this optional since we no longer need to open `ConnectingWalletConnectBasic` as it's already handled in `ModalContoller.open`

magiziz: Noticed this issue when i was on mobile and it didn't open `AllWallets` view

tomiir: these are internal options right? cause at a ModalController level we always require a view

tomiir (approved):

examples/html-ak-basic-up/src/main.js
```diff
@@ -117,13 +117,11 @@ async function initializeApp() {
...
 
-  provider.on('display_uri', uri => {
-    modal.open({ uri, view: 'ConnectingWalletConnectBasic' })
-  })
-
   // Button handlers
   document.getElementById('connect')?.addEventListener('click', async () => {
     setLoading(true)
+    await modal.open()
+    modal.subscribeEvents(({ data }) => data.event === 'MODAL_CLOSE' && setLoading(false))
```
enesozturk, line 124: Can we avoid using inline if's?

packages/appkit/src/client/appkit-basic.ts
enesozturk: What are the use cases users might want to call open without any params?

Also, the other params seems optional, does this requires passing empty object to the open function? Like

```js
modal.open({ }) // we shouldn't require object if all props are optional
```

This should be available and happy with TS:

```js
modal.open() ✅ 
```

magiziz: With WCM deprecation there are a few scenarios that users might pass either `view` or `uri`

If user wants to manually control the modal flow and they use ethereum provider, sign client or basic they'd call this
```ts
modal.open({ uri })
```

If user uses appkit directly and they want to open any available screen
```ts
modal.open({ view: 'Swap' )}
```

Both should be optional, but yes in most cases users just need to call `modal.open()`

PR #3841, opened 2025-02-11
fix: typos in documentation files

maximevtush:


enesozturk:
Thank you for your contribution. At this time, Reown does not generally accept these types of contributions for security reasons.

PR #3853, opened 2025-02-13
Implement erc7811 for usdc send test

KannuSingh:
# Description

Add ERC 7811 wallet_getAssets support on the Appkit USDC send test

## Type of change

# Associated Issues

For GH issues: closes #...

# Showcase (Optional)

https://github.com/user-attachments/assets/c90ba090-4231-404d-a950-fb677f58a3ba

# Checklist


apps/laboratory/src/components/Wagmi/WagmiSendUSDCTest.tsx
lukaisailovic, line 1: Will this approach load all balances on page load?

KannuSingh: Yes, it will. 
By either making `wallet_getAssets` rpc request to wallet directly if its support, or using wallet service. else fallback to fetching it via querying onchain data.

lukaisailovic: Why is that necessary? If I'm just using the lab to test some stuff and not care about USDC send (which is the most common use case) why would it fetch balances for me?

KannuSingh: Ok. I can add a "Fetch Balance" button in the WagmiSendUSDCTest component, so the balance is fetched only when the button is clicked.

KannuSingh: Makes sense. Since it's calling the Portfolio API, it would trigger unnecessary requests during testing

lukaisailovic: what about replacing send usdc inline form with a modal. When you click send it opens a modal and then it fetches the balance before it allows you to send. We can also then remove the cache and the provider and just fetch balance each time the modal is opened

lukaisailovic (approved):

PR #3860, opened 2025-02-13
chore: initialize solana connection on adapter construction

tomiir:
# Description
- Initializes solana connection on adapter construction instead of waiting for connection

Why: W3M solana users had the connection setup on init and with multichain migration this got changed.
 
## Type of change

# Checklist


packages/adapters/solana/src/client.ts
```diff
@@ -45,6 +45,12 @@ export class SolanaAdapter extends AdapterBlueprint<SolanaProvider> {
     this.wallets = options.wallets
   }
 
+  public override construct(params: AdapterBlueprint.Params): void {
+    super.construct(params)
+    const rpcUrl = params.networks?.[0]?.rpcUrls.default.http[0] as string
```
svenvoskamp, line 50: Check if a value exists in `localStorage` using `storageUtil.getActiveCaipNetworkId()`. Use this value in stead to find the correct network and `rpcUrl` value. We can fallback to the above implementation

svenvoskamp (changes requested):

packages/adapters/solana/src/client.ts
Sam-Newman: @tomiir Can we get this merged?

packages/adapters/solana/src/client.ts
```diff
@@ -47,6 +47,17 @@ export class SolanaAdapter extends AdapterBlueprint<SolanaProvider> {
     this.wallets = options.wallets
   }
 
+  public override construct(params: AdapterBlueprint.Params): void {
```
enesozturk, line 50: Why do we need to do it here instead of AppKit client? Can we add PR description "the why" also? 🙏🏽

enesozturk (approved):

packages/adapters/solana/src/client.ts
tomiir: This is a solana specific change. W3M solana users had the connection setup on init and with multichain migration this got changed. See no reason to not init when we can instead of waiting for an actual wallet connection.

PR #3879, opened 2025-02-19
fix: semaphore balance calls 

ganchoradkov:
# Description
This PR fixes bug with the blockchain api caching. When multiple calls are fired before the cache is set, all of them were hitting the endpoint. Now, the first request sets a promise and all subsequent requests wait to resolve 

Implemented tests as well

## Type of change

# Associated Issues
https://www.notion.so/Performance-Engineering-Sync-19f3a661771e805f9078f3f8dd01111b?d=19f3a661771e80b6b4d0001ca5026c14&pvs=4#19f3a661771e800b8b65c002afb7493b

# Checklist


tomiir (approved):
This looks ok but wondering if we can abstract this to something like a BalanceManager or something so you'd only do 
```
const manager = new BalanceManager({ getBalance: this.getBalance.bind(this), caipAddress })
const balance = balanceManager.getBalance()
``` 

We can implement later but wondering if you see anything wrong with this . It could also handle cache and such internally, think it would be really clean

ganchoradkov:
> This looks ok but wondering if we can abstract this to something like a BalanceManager or something so you'd only do
> 
> ```
> const manager = new BalanceManager({ getBalance: this.getBalance.bind(this), caipAddress })
> const balance = balanceManager.getBalance()
> ```
> 
> We can implement later but wondering if you see anything wrong with this . It could also handle cache and such internally, think it would be really clean

Agree, too many things around the adapters are copy/paste. They should be just proxies with a single invoking point e.g. adapterManager

PR #3883, opened 2025-02-19
feat: `enableNetworkSwitch` option

tomiir:
# Description

- Adds `showNetworkSwitcher` prop to allow disabling of network switching in the modal.

- TODO: Handle case where user switches network in the wallet

## Type of change

# Associated Issues

For Linear issues: Closes APKT-2134

# Checklist


packages/appkit/src/utils/TypesUtil.ts
```diff
@@ -95,4 +95,10 @@ export type AppKitOptions = {
    * @default "{ bip122: 'payment', eip155: 'smartAccount', polkadot: 'eoa', solana: 'eoa' }"
    */
   defaultAccountTypes?: Partial<OptionsControllerState['defaultAccountTypes']>
+
+  /**
+   * Enable or disable network switching in the modal.
+   * @default true
+   */
+  showNetworkSwitcher?: boolean
```
tomiir, line 103: I think we can better name this to `enableNetworkSwitch`. Cause if user switches from wallet it is also restricted and not related to network switcher

tomiir (commented):
Do you think we should prevent programatic network switching as well=?

packages/controllers/src/controllers/ChainController.ts
```diff
@@ -354,6 +354,7 @@ export const ChainController = {
 
     if (
       !isSupported &&
+      OptionsController.state.showNetworkSwitcher &&
```
tomiir, line 357: I don't think this would work as it would prompt the modal to switch to any other network configured, it should only allow to switch to the already connected network

packages/appkit/tests/client/public-methods.test.ts
```diff
@@ -1043,4 +1044,69 @@ describe('Base Public methods', () => {
       }
     })
   })
+
+  it('should handle network switcher UI when showNetworkSwitcher is true', async () => {
```
tomiir, line 1048: This should check it only contains the single already connected network

packages/appkit/src/client/appkit-base-client.ts
```diff
@@ -1001,6 +1003,7 @@ export abstract class AppKitBaseClient {
       if (network?.chainNamespace === ChainController.state.activeChain) {
         // If the network is unsupported and the user doesn't allow unsupported chains, we show the unsupported chain UI
         if (
+          OptionsController.state.showNetworkSwitcher &&
           !OptionsController.state.allowUnsupportedChain &&
           ChainController.state.activeCaipNetwork?.name === ConstantsUtil.UNSUPPORTED_NETWORK_NAME
         ) {
```
enesozturk, line 1009: Why do we show unsupported network only if network switcher enabled? Let's say user has Mainnet and Polygon enabled in AppKit. And via hooks or from wallet listeners they switched to OP, and if this prop is false, we're not going to show unsupported network UI right? Is this expected?

packages/controllers/src/controllers/ChainController.ts
magiziz: this shouldn't prompt the modal at all, if `showNetworkSwitcher` is false it'd ignore prompting the modal to switch to any chain

packages/appkit/tests/client/public-methods.test.ts
magiziz: which list ? the `caipNetworks` list ?

packages/appkit/src/client/appkit-base-client.ts
tomiir: If network switching is not allowed, we should still react to the network switch, but prompt the user to switch to the single network allowed (the connected one)

packages/controllers/src/controllers/ChainController.ts
tomiir: mb was meant for other piece of code

PR #3924, opened 2025-02-26
fix: ethers write contract using caipAddress

tomiir:
# Description

- Ethers adapters were using `caipAddress` instead of the regular `address` for `writeContract` interactions and tried to create a signer with it, causing Send and Swap to fail.

## Type of change

# Checklist

### DIFF
--- a/packages/core/src/controllers/SendController.ts
+++ b/packages/core/src/controllers/SendController.ts
@@ -291,6 +291,7 @@ export const SendController = {
       })
       this.resetSend()
     } catch (error) {
+      console.error('SendController:sendERC20Token - failed to send native token', error)
       const errorMessage = error instanceof Error ? error.message : 'Unknown error'
       EventsController.sendEvent({
         type: 'track',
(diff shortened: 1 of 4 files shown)


packages/core/src/controllers/SendController.ts
```diff
@@ -344,6 +345,21 @@ export const SendController = {
         this.resetSend()
       }
     } catch (error) {
+      console.error('SendController:sendERC20Token - failed to send erc20 token', error)
```
magiziz, line 348: Maybe we can use `AlertController` here ?

magiziz (approved):

enesozturk (approved):

packages/core/src/controllers/SendController.ts
enesozturk: Unless we dont properly handle the error on the UI (like let user know why its happened, what they can/should do etc) I disagree with the Alert comp

PR #3970, opened 2025-03-07
fix: `watchConnections` emits a disconnect even during a page refresh

magiziz:
# Description

A very unusual edge case, but when refreshing the page after connecting via email or social login a disconnection occurs before reconnecting. Instead of relying on `watchConnections` which causes this issue we should use `watchAccount` only for emitting disconnect event.

## Type of change

# Associated Issues

For Linear issues: Closes APKT-2280

# Checklist


packages/adapters/wagmi/src/client.ts
```diff
@@ -240,13 +240,6 @@ export class WagmiAdapter extends AdapterBlueprint {
         }
       }
     })
-    watchConnections(this.wagmiConfig, {
-      onChange: connections => {
-        if (connections.length === 0) {
-          this.emit('disconnect')
-        }
-      }
-    })
```
magiziz, line 249: We already have tests for `watchAccount`

enesozturk (approved):

enesozturk:
@magiziz There are some ts issues

tomiir (approved):
Can we add a test?

magiziz:
@tomiir sorry might have resolved a comment i put, but seems like we already have test covered for `watchAccount` https://github.com/reown-com/appkit/pull/3970#discussion_r1984846734

packages/adapters/wagmi/src/client.ts
magiziz: It only happens when you're logged in with social provider and refresh the page. The wagmi thinks there are no connections at first and then it emits disconnect event.

I think it's only possible to catch this in an e2e test, but it's an extreme edge case

magiziz: i was only able to reproduce this in web wallet

svenvoskamp: I remember I added this for a specific case, but can't remember anymore. Not sure if it save to delete

svenvoskamp: But this was before we did improvements on initialisation AppKit, so might be okay to remove it. We just need make sure this gets manually tested

PR #4032, opened 2025-03-17
feat: pay

lukaisailovic:
# Description

Added Pay with exchange as well as pay with wallet option.

Its interacted with the new `openPay` method:
```
import { openPay } from '@reown/appkit-pay'

await openPay({
    network: 'eip155:8453',
    recipient: '',
    asset: '0x833589fcd6edb6e08f4c7c32d4f71b54bda02913',
    amount: 20,
    metadata: {
      name: 'Ethereum',
      symbol: 'ETH',
      decimals: 18
    {
})
```

If pay with wallet is selected, the new connection listener will be set up and will automatically trigger the payment upon wallet connection.

If pay with exchange is set up it will open an exchange URL fetched from blockchain API

## Type of change

# Associated Issues

For GH issues: closes #...

# Showcase (Optional)

The method will open new Pay view: 
Screenshot 2025-04-08 at 15 27 27

Pay with exchange interaction recording: [Slack URL](https://reown-inc.slack.com/archives/C08AJ6DK9QV/p1743615617535239)

# Checklist

- [] I have tested my changes on the preview link (BLOCKCHAIN API NEEDS TO BE REDEPLOYED, PAY WITH EXCHANGE WILL NOT WORK!)


apps/laboratory/app/library/pay-default/page.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+import { AppKitButtonsMultiChain } from '@/src/components/AppKitButtonsMultiChain'
+import { AppKitInfo } from '@/src/components/AppKitInfo'
+import { AppKitInfoMultiChain } from '@/src/components/AppKitInfoMultiChain'
+import { AppKitPay } from '@/src/components/AppKitPay'
+import { ConstantsUtil } from '@/src/utils/ConstantsUtil'
+
+const queryClient = new QueryClient()
+
+const networks = ConstantsUtil.AllNetworks
```
tomiir, line 21: Is this actually enabled in all networks?

apps/laboratory/src/components/AppKitPay.tsx
```diff
@@ -0,0 +1,229 @@
...
+  amount: number
+  metadata: Metadata
+}
+
+export function AppKitPay() {
+  const { isOpen, onToggle } = useDisclosure()
+  const { open, isLoading, error, result } = usePay()
+  const toast = useChakraToast()
+  const [paymentDetails, setPaymentDetails] = useState<AppKitPaymentAsset>({
+    network: 'eip155:8453',
```
tomiir, line 47: Is this only enabled here? or should we use the id from `useAppKitNetwork`?

packages/pay/exports/react.ts
```diff
@@ -0,0 +1,120 @@
...
+  /**
+   * Function to open the payment modal with the specified options.
+   * @param options - The payment configuration options.
+   */
+  open: (options: PaymentOptions) => Promise<void>
+
+  /**
+   * Indicates whether a payment process is currently in progress (e.g., loading exchanges, waiting for transaction).
+   */
+  isLoading: boolean
```
tomiir, line 19: do we want to differentiate the different parts of state that are loading?

packages/pay/src/client.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export function getPayResult() {
+  return PayController.state.payResult
+}
+export function getPayError() {
+  return PayController.state.error
+}
+export function getIsPaymentInProgress() {
+  return PayController.state.isPaymentInProgress
+}
```
tomiir, line 20: do we need this? seems like a proxy to the controller entirely

packages/pay/src/controllers/PayController.ts
```diff
@@ -0,0 +1,281 @@
...
+  paymentAsset: {
+    network: 'eip155:1',
+    recipient: '0x0',
+    asset: '0x0',
+    amount: 0,
+    metadata: {
+      name: '0x0',
+      symbol: '0x0',
+      decimals: 0
+    }
```
tomiir, line 52: is it ok to have this default?

packages/pay/src/controllers/PayController.ts
```diff
@@ -0,0 +1,281 @@
...
+  },
+
+  async fetchExchanges() {
+    try {
+      state.isLoading = true
+      const response = await getExchanges({
+        page: DEFAULT_PAGE
+      })
+      // Putting this here in order to maintain backawrds compatibility with the UI when we introduce more exchanges
+      state.exchanges = response.exchanges.slice(0, 2)
```
tomiir, line 120: wouldn't this be handled by BE sorting?

packages/pay/src/controllers/PayController.ts
```diff
@@ -0,0 +1,281 @@
...
+      SnackController.showError(AppKitPayErrorMessages.UNABLE_TO_GET_EXCHANGES)
+      throw new AppKitPayError(AppKitPayErrorCodes.UNABLE_TO_GET_EXCHANGES)
+    } finally {
+      state.isLoading = false
+    }
+  },
+
+  async getPayUrl(exchangeId: string) {
+    try {
+      const amount = Number(state.paymentAsset.amount)
```
tomiir, line 131: is this an Int? If it's a bigint, this might cause issues

packages/pay/src/controllers/PayController.ts
```diff
@@ -0,0 +1,281 @@
...
+    if (state.isConfigured) {
+      return
+    }
+    ProviderUtil.subscribeProviders(async _ => {
+      const chainNamespace = ChainController.state.activeChain as ChainNamespace
+      const provider = ProviderUtil.getProvider(chainNamespace)
+      if (!provider) {
+        return
+      }
+      await this.handlePayment()
```
tomiir, line 158: do we need any guardrails to prevent this executing on other flows? 
how are we preventing this from firing when someone connects through regular flow or reconnection (where provider is set)

tomiir: Is this intended to fire once after the dev uses pay flow to trigger `subscribeEvents` ? 
If so, should we unsubscribe after execution?

packages/pay/src/types/errors.ts
```diff
@@ -0,0 +1,57 @@
...
+  [AppKitPayErrorCodes.INVALID_ASSET]: 'Invalid asset specified',
+  [AppKitPayErrorCodes.INVALID_AMOUNT]: 'Invalid payment amount',
+
+  [AppKitPayErrorCodes.UNKNOWN_ERROR]: 'Unknown payment error occurred',
+  [AppKitPayErrorCodes.UNABLE_TO_INITIATE_PAYMENT]: 'Unable to initiate payment',
+  [AppKitPayErrorCodes.INVALID_CHAIN_NAMESPACE]: 'Invalid chain namespace',
+  [AppKitPayErrorCodes.GENERIC_PAYMENT_ERROR]: 'Unable to process payment',
+  [AppKitPayErrorCodes.UNABLE_TO_GET_EXCHANGES]: 'Unable to get exchanges',
+  [AppKitPayErrorCodes.ASSET_NOT_SUPPORTED]: 'Asset not supported by the selected exchange'
+}
```
tomiir, line 34: should we set the `error` type on the hook to these instead of `string`?

packages/pay/src/ui/w3m-pay-view/index.ts
```diff
@@ -0,0 +1,250 @@
...
+   * Check if wallet is connected based on active address
+   */
+  private get isWalletConnected(): boolean {
+    return AccountController.state.status === 'connected'
+  }
+
+  // -- Lifecycle ----------------------------------------- //
+  public override connectedCallback() {
+    super.connectedCallback()
+    this.initializePaymentDetails()
```
tomiir, line 71: why here instead of constructor?

packages/scaffold-ui/src/types/router.d.ts
```diff
@@ -0,0 +1,61 @@
+import '@reown/appkit-controllers'
```
tomiir, line 1: is this needed?

packages/pay/exports/react.ts
lukaisailovic: For now this only means payment is loading, not appkit pay is loading. Might change the name to make it more explicit. Nothing else needed atm

packages/pay/src/client.ts
lukaisailovic: Yeah, thats what we need to expose to the public api for non-react clients. Wdyt is there an easier way to do it? I don't want to expose the controller directly

packages/pay/src/controllers/PayController.ts
lukaisailovic: Yeah, we're checking them after so seems good. I don't want the default asset set by us

packages/pay/src/controllers/PayController.ts
lukaisailovic: Do you think this is a good way to handle it? We can also do it on the BE

packages/pay/src/controllers/PayController.ts
lukaisailovic: Good point, we should unsub. And yes, if you open pay and then close it, and then open appkit and connect it would still be triggered. Acceptable IMO as there isn't a use case where pay should be used along appkit (modal) in the same dapp

packages/pay/src/ui/w3m-pay-view/index.ts
lukaisailovic: Leftover from before, moving!

apps/laboratory/src/components/AppKitInfo.tsx
```diff
@@ -21,6 +21,11 @@ import { EmbeddedWalletInfo } from './EmbeddedWalletInfo'
 export function AppKitInfo() {
   const { caipAddress, address } = useAppKitAccount()
   const { chainId } = useAppKitNetwork()
+  const [isMounted, setIsMounted] = React.useState(false)
+
+  React.useEffect(() => {
+    setIsMounted(true)
+  }, [])
```
magiziz, line 28: Don't we already have `<InitializeBoundary>` in place for `<AppKitInfo>` ?

package.json
```diff
@@ -17,6 +17,7 @@
     "build:example:vue-wagmi": "pnpm --filter @examples/vue-wagmi build",
     "build:example:parcel-react-wagmi": "pnpm --filter @examples/parcel-react-wagmi build",
     "build:example:wagmi-app-router": "pnpm --filter @examples/next-wagmi-app-router build",
+    "build:local": "turbo run build --filter={./packages/**/*} --filter=!{./packages/cdn} --concurrency=31",
```
magiziz, line 20: needed ?

packages/pay/src/client.ts
magiziz: Can we maybe have create a function called `createAppKitPay` which has all those functions ?

magiziz: Example:

```ts
const appKitPay = createAppKitPay()

const isPaymentInProgress = appkitPay.getIsPaymentInProgress
```

magiziz: also would be nice if we can subscribe events like `appKitPay.subscribeState` so users can subscribe to events realtime

packages/pay/src/controllers/PayController.ts
magiziz: is setting `decimals` to 0 fine ? What happens if we set it to 18 ?

packages/pay/src/controllers/PayController.ts
magiziz: can users choose how many exchanges they want to display ? Maybe they can define it in appkit config ?

packages/pay/src/controllers/PayController.ts
```diff
@@ -0,0 +1,348 @@
...
+        asset: formatCaip19Asset(state.paymentAsset.network, state.paymentAsset.asset),
+        amount: amount.toString(16),
+        recipient: `${state.paymentAsset.network}:${state.paymentAsset.recipient}`
+      })
+
+      return response.url
+    } catch (error) {
+      if ((error as Error).message.includes('is not supported')) {
+        throw new AppKitPayError(AppKitPayErrorCodes.ASSET_NOT_SUPPORTED)
+      }
```
magiziz, line 176:
```suggestion
      if (error instanceof Error && error.message.includes('is not supported')) {
        throw new AppKitPayError(AppKitPayErrorCodes.ASSET_NOT_SUPPORTED);
      }
```

packages/pay/src/controllers/PayController.ts
```diff
@@ -0,0 +1,348 @@
...
+      throw new Error((error as Error).message)
+    }
+  },
+
+  subscribeEvents() {
+    if (state.isConfigured) {
+      return
+    }
```
magiziz, line 184: Ah we already have subscribe events here, ignore the comment i left above 🙏

packages/pay/src/controllers/PayController.ts
```diff
@@ -0,0 +1,348 @@
...
+      state.paymentType = 'exchange'
+      state.isPaymentInProgress = false
+      const payUrl = await this.getPayUrl(exchangeId)
+      if (!payUrl) {
+        throw new AppKitPayError(AppKitPayErrorCodes.UNABLE_TO_INITIATE_PAYMENT)
+      }
+      RouterController.push('PayLoading')
+
+      if (state.openInNewTab) {
+        window.open(payUrl, '_blank')
```
magiziz, line 327: Should we also add `'noopener,noreferrer'` in the 3rd param when opening a new window ? I think we have `CoreHelperUtil.returnOpenHref` for that

packages/pay/src/ui/w3m-pay-view/index.ts
```diff
@@ -0,0 +1,238 @@
...
+  }
+
+  private async onDisconnect(e: Event) {
+    e.stopPropagation()
+    try {
+      await ConnectionController.disconnect()
+      ModalController.close()
+    } catch {
+      SnackController.showError('Failed to disconnect')
+    }
```
magiziz, line 226: Can we also add console log in case something goes wrong and we want to debug ?

apps/laboratory/src/components/AppKitPay.tsx
lukaisailovic: It could be different. We are checking now and doing the switch if necessary.

packages/pay/src/client.ts
lukaisailovic: @magiziz Instead of the whole state, I added a method to subscribe to specific keys. I think thats good enough for now. If we expose whole state, then its the public interface which isn't great.

packages/pay/src/controllers/PayController.ts
lukaisailovic: Nothing. It will be overridden by the config

packages/pay/src/controllers/PayController.ts
lukaisailovic: They can do `include` and `exclude` on the exchanges, but ultimately max of 2 can be displayed on the UI

apps/laboratory/src/components/AppKitInfo.tsx
lukaisailovic: We didn't have it at a time, removing mine

package.json
lukaisailovic: Build without the cdn takes 4 seconds. With cdn 20 seconds. Running watch is much more resource expensive so I added this another way to build the project without cdn

packages/pay/exports/react.ts
```diff
@@ -0,0 +1,121 @@
...
+  const open = useCallback(async (options: PaymentOptions) => {
+    await PayController.handleOpenPay(options)
+  }, [])
+
+  return {
+    open,
+    isLoading: isControllerLoading || isPaymentInProgress,
+    error,
+    result: result ?? null
+  }
```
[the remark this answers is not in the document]
lukaisailovic: Good point! Done

magiziz (approved):

packages/pay/exports/react.ts
```diff
@@ -160,3 +170,147 @@ export function usePay(parameters?: UsePayParameters): UsePayReturn {
...
+    } finally {
+      setIsLoading(false)
+    }
+  }, [])
+
+  useEffect(() => {
+    if (isFetchOnInit) {
+      fetchExchanges(initialPage).catch(() => {
+        // Error is already handled and set in state by fetchExchanges
+      })
```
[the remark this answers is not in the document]
lukaisailovic: There is a possibility for something to fail outside of that (setting the state etc) but its not really significant so thats why I don't action it. Wdyt should we just remove it?

packages/pay/package.json
```diff
@@ -0,0 +1,67 @@
...
+    ".": {
+      "types": "./dist/types/exports/index.d.ts",
+      "import": "./dist/esm/exports/index.js",
+      "default": "./dist/esm/exports/index.js"
+    },
+    "./react": {
+      "types": "./dist/types/exports/react.d.ts",
+      "import": "./dist/esm/exports/react.js",
+      "default": "./dist/esm/exports/react.js"
+    }
```
enesozturk, line 21: Prob better to support Vue too like other hooks we have

packages/pay/src/controllers/PayController.ts
```diff
@@ -0,0 +1,415 @@
...
+            )
+          }
+          if (state.paymentAsset.asset.startsWith('0x')) {
+            state.currentPayment.result = await processEvmErc20Payment(
+              state.paymentAsset,
+              address as `0x${string}`
+            )
+          }
+          break
+        case ConstantsUtil.CHAIN.SOLANA:
```
enesozturk, line 297: Why do we return nothing on Solana but throw error for others?

packages/pay/src/ui/w3m-pay-view/index.ts
```diff
@@ -0,0 +1,246 @@
...
+    try {
+      this.loadingExchangeId = exchangeId
+      const result = await PayController.handlePayWithExchange(exchangeId)
+      if (result) {
+        await ModalController.open({
+          view: 'PayLoading'
+        })
+        CoreHelperUtil.openHref(result.url, result.openInNewTab ? '_blank' : '_self')
+      }
+    } finally {
```
enesozturk, line 221: Should we handle if result is not as expected?

enesozturk (approved):

packages/pay/package.json
lukaisailovic: We don't need it atm. When the customer requests it we can just build it then

packages/common/src/utils/TypeUtil.ts
```diff
@@ -145,3 +145,37 @@ export type SdkFramework = 'html' | 'react' | 'vue' | 'cdn' | 'unity'
...
+  | 'erc20'
+  | 'erc721'
+  | 'erc1155'
+  | 'nft'
+  | (string & Record<never, never>)
+
+/**
+ * Format: [-.%a-zA-Z0-9]{1,128}
+ */
+export type AssetReference = string
```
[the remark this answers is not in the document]
lukaisailovic: Not worth it atm, too complex IMO

PR #4046, opened 2025-03-19
fix: account not syncing on some namespaces when using multichain

magiziz:
# Description

When using multichain if we don’t emit `chainId` the CAIP address will not be set for the namespace.

## Type of change

# Associated Issues

For Linear issues: Closes APKT-2519

# Checklist

### DIFF
--- a/packages/adapters/wagmi/src/client.ts
+++ b/packages/adapters/wagmi/src/client.ts
@@ -226,7 +226,8 @@ export class WagmiAdapter extends AdapterBlueprint {
           ) {
             this.setupWatchPendingTransactions()
             this.emit('accountChanged', {
-              address: accountData.address
+              address: accountData.address,
+              chainId: accountData.chainId
             })
           }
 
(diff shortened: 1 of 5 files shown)


packages/adapters/wagmi/src/client.ts
```diff
@@ -226,7 +226,8 @@ export class WagmiAdapter extends AdapterBlueprint {
           ) {
             this.setupWatchPendingTransactions()
             this.emit('accountChanged', {
-              address: accountData.address
+              address: accountData.address,
+              chainId: accountData.chainId
```
tomiir, line 230: isn't there a `chainChanged` event below?

magiziz: yes but we also need to include it in `accountChanged`

tomiir: but this emit does not prevent the other from firing, wouldn't appkit react to both? can it have conflicts?

magiziz: Yeah i thought the same. I tested it rn and other things aren’t getting fired because there’s a lot of active namespace logic in the code that runs before triggering actions like API calls so we should be good there

magiziz: Here's the logic

https://github.com/reown-com/appkit/blob/main/packages/appkit/src/client/appkit-base-client.ts#L738-L756

enesozturk (approved):

PR #4068, opened 2025-03-24
fix: coinbase multichain not working

magiziz:
# Description

When using multichain to connect to Coinbase with Ethereum, there’s a bug where the active namespace isn’t properly checked which causes it to connect to the Coinbase Solana wallet instead.

## Type of change

# Associated Issues

For Linear issues: Closes APKT-2547

# Checklist

### DIFF
--- a/packages/controllers/src/controllers/ChainController.ts
+++ b/packages/controllers/src/controllers/ChainController.ts
@@ -656,9 +656,12 @@ export const ChainController = {
       const disconnectResults = await Promise.allSettled(
         chainsToDisconnect.map(async ([ns, adapter]) => {
           try {
-            if (adapter.connectionControllerClient?.disconnect) {
+            const { caipAddress } = this.getAccountData(ns) || {}
+
+            if (caipAddress && adapter.connectionControllerClient?.disconnect) {
               await adapter.connectionControllerClient.disconnect(ns)
             }
+
             this.resetAccount(ns)
             this.resetNetwork(ns)
           } catch (error) {
--- a/packages/controllers/src/controllers/ConnectorController.ts
+++ b/packages/controllers/src/controllers/ConnectorController.ts
@@ -224,9 +224,12 @@ export const ConnectorController = {
   getConnectorById(id: string) {
     return state.allConnectors.find(c => c.id === id)
   },
-
   getConnector(id: string, rdns?: string | null) {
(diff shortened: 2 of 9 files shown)


packages/controllers/src/controllers/ChainController.ts
```diff
@@ -656,9 +656,12 @@ export const ChainController = {
       const disconnectResults = await Promise.allSettled(
         chainsToDisconnect.map(async ([ns, adapter]) => {
           try {
-            if (adapter.connectionControllerClient?.disconnect) {
+            const { caipAddress } = this.getAccountData(ns) || {}
+
+            if (caipAddress && adapter.connectionControllerClient?.disconnect) {
               await adapter.connectionControllerClient.disconnect(ns)
             }
```
magiziz, line 663: We need to check if a user is connected before calling `disconnect` for every adapter

tomiir: shouldn't this be done at blueprint level then?

tomiir (approved):

packages/controllers/src/controllers/ConnectorController.ts
```diff
@@ -224,9 +224,12 @@ export const ConnectorController = {
   getConnectorById(id: string) {
     return state.allConnectors.find(c => c.id === id)
   },
-
   getConnector(id: string, rdns?: string | null) {
-    return state.allConnectors.find(c => c.explorerId === id || c.info?.rdns === rdns)
+    const connectorsByNamespace = state.allConnectors.filter(
```
enesozturk, line 228:
```suggestion
    const activeNamespaceConnectors = state.allConnectors.filter(
```
maybe which indicates what does the filter do exactly - but not super imp

packages/controllers/tests/controllers/ConnectionController.test.ts
```diff
@@ -247,6 +247,11 @@ describe('ConnectionController', () => {
       polkadot: 'polkadot-connector',
       bip122: CommonConstantsUtil.CONNECTOR_ID.WALLET_CONNECT
     }
+    ChainController.state.chains.set('eip155', {
```
enesozturk, line 250: Maybe use again `ConstantsUtil.CHAIN.EVM` as you did on `ChainController.test.ts`?

enesozturk: Same for the belows

enesozturk (approved):

PR #4104, opened 2025-03-28
fix: log instead of throwing when universal provider fails to load

magiziz:
# Description

When UniversalProvider fails to load we are again throwing another error and it causes `createAppKit` to crash fully. Instead we need to log the errors.

## Type of change

# Associated Issues

For GH issues: closes #...

# Showcase (Optional)

# Checklist

### DIFF
--- a/packages/appkit/src/client/appkit-base-client.ts
+++ b/packages/appkit/src/client/appkit-base-client.ts
@@ -1196,8 +1196,9 @@ export abstract class AppKitBaseClient {
     if (!this.universalProvider) {
       try {
         await this.createUniversalProvider()
-      } catch (error) {
-        throw new Error('AppKit:getUniversalProvider - Cannot create provider')
+      } catch (err) {
+        // eslint-disable-next-line no-console
+        console.error('AppKit:getUniversalProvider - Cannot create provider', err)
       }
     }
 


ganchoradkov (approved):

packages/appkit/src/client/appkit-base-client.ts
```diff
@@ -1196,8 +1196,9 @@ export abstract class AppKitBaseClient {
     if (!this.universalProvider) {
       try {
         await this.createUniversalProvider()
-      } catch (error) {
-        throw new Error('AppKit:getUniversalProvider - Cannot create provider')
+      } catch (err) {
+        // eslint-disable-next-line no-console
+        console.error('AppKit:getUniversalProvider - Cannot create provider', err)
```
ganchoradkov, line 1201: you could still throw but at least log the actual error

ganchoradkov: if its not recoverable

magiziz: I realised if we throw the error the whole `createAppKit` would fail, maybe it's better to log so users can still use appkit without WC connector

enesozturk: Yes I think we should still throw right?

enesozturk: But is shouldn't block AppKit to be implemented i think this is what we have rn

magiziz: If we throw users won't be able to use appkit at all, this is the state they'll be in

Screenshot 2025-03-28 at 15 28 32

enesozturk: Yeah I got this point. Only concern is that what happens if there is an error but use try to connect with UPA? They should get the reasonable error on AppKit somehow, can you reproduce that case?

tomiir: Hm interesting, this is like a critical error cause you should have a provider at this point.
We can 
A) Fail completely with a good error 
B) Do not fail, recover unconnected state

PR #4110, opened 2025-03-31
chore: Disable Try Again button when client is not connected

arein:
# Description

`Try Again` currently can end in a scenario where there is no feedback. This is especially likely when the user's network condition is really bad. This improves this by introducing visual feedback through resetting the UI.

## Type of change

# Showcase (Optional)

https://github.com/user-attachments/assets/7e994c8e-7183-475c-b188-1bf5ac939172

# Checklist


packages/scaffold-ui/test/partials/w3m-connecting-wc-mobile.test.ts
```diff
@@ -23,7 +34,11 @@ describe('W3mConnectingWcMobile', () => {
       ...ConnectionController.state,
       wcUri: WC_URI
     })
-    vi.spyOn(CoreHelperUtil, 'isMobile').mockReturnValue(true)
+  })
+
+  afterEach(() => {
+    vi.resetAllMocks()
+    vi.useRealTimers()
```
lukaisailovic, line 41: Curious, why use real times instead of fake timers and advancing the time manually?

arein: there appeared to be some side-effects from other tests - this prevented that

magiziz (approved):

packages/scaffold-ui/src/partials/w3m-connecting-wc-mobile/index.ts
```diff
@@ -90,6 +94,24 @@ export class W3mConnectingWcMobile extends W3mConnectingWidget {
 
   protected override onTryAgain() {
     if (!this.buffering) {
+      // Clear existing timeouts
```
enesozturk, line 97: Seems like we have the same logics in the constructor, maybe we could move these business logics to a function and call it? Would increase the maintainability

arein: thank you - addressed

packages/scaffold-ui/test/partials/w3m-connecting-wc-mobile.test.ts
arein: Weirdly I was able to remove this now - appears like another merged code change fixed the side-effect

PR #4151, opened 2025-04-06
refactor: read caip networks state by via getter functions

enesozturk:
# Description

Refactors the `caipNetworks` state reading logics to make the read ChainController state via `getCaipNetworks` logic.

In the Demo app when we add/remove new networks, AppKit should dynamically re-generate the QR with the new set of available networks. But it's currently instead using the first state of the CAIP networks when we initialize the AppKit.

This is causing a problem where after playing around with the networks on Demo and try to connect, the Demo app is not opening an expected connection with the wallet.

## Type of change

# Associated Issues

For GH issues: closes #...

# Showcase (Optional)

# Checklist


magiziz (approved):

apps/demo/lib/config.ts
```diff
@@ -40,24 +33,12 @@ if (!projectId) {
...
-  aurora,
-  mantle
-] as [AppKitNetwork, ...AppKitNetwork[]]
+export const evmNetworks = [mainnet, optimism, bsc, polygon, avalanche, arbitrum, zksync, base] as [
+  AppKitNetwork,
+  ...AppKitNetwork[]
+]
 
-export const solanaNetworks = [solana, solanaTestnet, solanaDevnet] as AppKitNetworksType
+export const solanaNetworks = [solana, solanaDevnet] as AppKitNetworksType
```
tomiir, line 41: why?

packages/adapters/ethers/src/client.ts
```diff
@@ -465,7 +465,7 @@ export class EthersAdapter extends AdapterBlueprint {
     params: AdapterBlueprint.GetBalanceParams
   ): Promise<AdapterBlueprint.GetBalanceResult> {
     const address = params.address
-    const caipNetwork = this.caipNetworks?.find(network => network.id === params.chainId)
+    const caipNetwork = this.getCaipNetworks().find(network => network.id === params.chainId)
```
tomiir, line 468: should we remoev the `caipNetworks` property alltogether? else it would be confusing to know which to use

packages/scaffold-ui/src/views/w3m-connecting-wc-view/index.ts
```diff
@@ -65,7 +65,12 @@ export class W3mConnectingWcView extends LitElement {
     try {
       const { wcPairingExpiry, status } = ConnectionController.state
 
-      if (retry || CoreHelperUtil.isPairingExpired(wcPairingExpiry) || status === 'connecting') {
+      const shouldRetry =
+        retry || OptionsController.state.enableEmbedded
+          ? true
+          : CoreHelperUtil.isPairingExpired(wcPairingExpiry)
```
tomiir, line 71: isnt 
```retry || OptionsController.state.enableEmbedded || CoreHelperUtil.isPairingExpired(wcPairingExpiry)``` 
the same?

apps/demo/lib/config.ts
enesozturk: It was previously like that, since we are showing limited networks to the users, it's better to keep the same array on background. Just reverting

packages/adapters/ethers/src/client.ts
enesozturk: Exactly

PR #4152, opened 2025-04-07
chore: filter recent wallets by chain namespace

magiziz:
# Description

Updated recent wallet list to only show wallets that match the current chain namespace

## Type of change

# Associated Issues

For Linear issues: Closes APKT-2509

# Showcase (Optional)

https://github.com/user-attachments/assets/bdcf930d-165c-4473-bca8-240e651eba56

# Checklist


packages/scaffold-ui/src/partials/w3m-connect-recent-widget/index.ts
```diff
@@ -42,13 +44,29 @@ export class W3mConnectRecentWidget extends LitElement {
...
+            const chainNamespace = c.split(':')[0] as ChainNamespace
+
+            return currentNamespace === chainNamespace
+          })
+
+          return hasMatchingNamespace
+        }
+
+        return true
+      })
```
enesozturk, line 69: What about doing a one iteration? That'd be more performant. Maybe we could move the logic to helper util like `filterRecentWallets` and pass the function to filter, like :
```suggestion
      .filter(filterRecentWallets)
```

packages/scaffold-ui/src/partials/w3m-connect-recent-widget/index.ts
```diff
@@ -84,6 +84,26 @@ export class W3mConnectRecentWidget extends LitElement {
 
     ConnectorController.selectWalletConnector(wallet)
   }
+
+  private hasNoWalletConnector(wallet: WcWallet) {
+    return !this.connectors.some(
+      connector => connector.id === wallet.id || connector.name === wallet.name
+    )
+  }
```
tomiir, line 92: it is usually a good practice to NOT use negated naming of functions/methods. 
So in this case it's preferred to do 

```
  private hasWalletConnector(wallet: WcWallet) {
    return this.connectors.some(
      connector => connector.id === wallet.id || connector.name === wallet.name
    )
  }
  ```
  
  and do
  
  ```
  .filter(wallet => !this.hasWalletConnector(wallet))
  ```

File: packages/scaffold-ui/src/partials/w3m-connect-recent-widget/index.ts [hunk 3]
```diff
@@ -84,6 +84,26 @@ export class W3mConnectRecentWidget extends LitElement {
...
+
+  private isWalletCompatibleWithCurrentChain(wallet: WcWallet) {
+    const currentNamespace = ChainController.state.activeChain
+
+    if (currentNamespace && wallet.chains) {
+      return wallet.chains.some(c => {
+        const chainNamespace = c.split(':')[0] as ChainNamespace
```
tomiir, line 99: don't we have a util for this?

magiziz: doesn't seem like it, we do `c.split(':')[0] as ChainNamespace` everywhere in the code

tomiir: we should add it imo I think we have `ParseUtil`

tomiir: I think you can use `parseCaipAddress` which should give you the 3 fields

enesozturk (approved):

PR #4158, opened 2025-04-07
fix: enable disable network logics on demo app

enesozturk:
# Description

Due to programmatic network and chain add/remove mechanism, the AppKit instance on Demo app is not generating the WC connection as expected. This is causing issues while connecting with deeplinks.

The issue is relying under the logics of adding/removing chains and networks for the Demo app programmatically. When we initializing the AppKit on Demo, we are reading the values from URL state and creating list of adapters and passing the configs object. And this is problematic.

In this PR, it's enabling all the adapters by default, user can still toggle networks, but AppKit won't remove chain adapters from the instance.

## Type of change

# Associated Issues

For GH issues: closes #...

# Showcase (Optional)

# Checklist

### DIFF
--- a/packages/appkit/src/client/appkit-base-client.ts
+++ b/packages/appkit/src/client/appkit-base-client.ts
@@ -369,16 +369,21 @@ export abstract class AppKitBaseClient {
 
         const fallbackCaipNetwork = this.getCaipNetwork(chainToUse)
 
-        const res = await adapter.connect({
-          id,
-          info,
-          type,
-          provider,
-          chainId: caipNetwork?.id || fallbackCaipNetwork?.id,
-          rpcUrl:
-            caipNetwork?.rpcUrls?.default?.http?.[0] ||
-            fallbackCaipNetwork?.rpcUrls?.default?.http?.[0]
-        })
+        const res = await adapter
+          .connect({
+            id,
+            info,
+            type,
+            provider,
+            chainId: caipNetwork?.id || fallbackCaipNetwork?.id,
+            rpcUrl:
+              caipNetwork?.rpcUrls?.default?.http?.[0] ||
+              fallbackCaipNetwork?.rpcUrls?.default?.http?.[0]
+          })
+          .catch(error => {
+            // eslint-disable-next-line no-console
+            console.error('@appkit-base-client: connectExternal: error', error)
+          })
 
         if (!res) {
           return
--- a/.github/workflows/pr_checks.yml
+++ b/.github/workflows/pr_checks.yml
@@ -5,6 +5,14 @@ on:
     types: [opened, synchronize, reopened, ready_for_review]
     branches:
       - main
+  merge_group:
+    # The merge queue provides the same benefits
+    # as the Require branches to be up to date before merging branch protection,
+    # but does not require a pull request author to update their pull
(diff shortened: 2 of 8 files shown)


packages/appkit/src/client/appkit-base-client.ts
```diff
@@ -115,15 +115,30 @@ export abstract class AppKitBaseClient {
...
     this.defaultCaipNetwork = this.extendDefaultCaipNetwork(options)
     this.chainAdapters = this.createAdapters(options.adapters as AdapterBlueprint[])
     this.initialize(options)
     this.sendInitializeEvent(options)
   }
 
+  private getChainNamespacesSet(adapters: AdapterBlueprint[], caipNetworks: CaipNetwork[]) {
+    const adapterNamespaces = adapters
+      ?.map(adapter => adapter.namespace)
+      .filter((namespace): namespace is ChainNamespace => Boolean(namespace))
```
tomiir, line 131: why use the adapters and not the namepsaces from the networks?

magiziz: maybe that'd be easier yeah

packages/appkit/src/client/appkit-base-client.ts
```diff
@@ -340,7 +355,7 @@ export abstract class AppKitBaseClient {
         const adapter = this.getAdapter(chainToUse)
 
         if (chain && chain !== activeChain && !caipNetwork) {
-          const toConnectNetwork = this.caipNetworks?.find(
+          const toConnectNetwork = this.getCaipNetworks()?.find(
```
magiziz, line 358: At what point does `this.getCaipNetworks` become `undefined` ?

packages/appkit/src/client/appkit-base-client.ts
```diff
@@ -1856,18 +1817,16 @@ export abstract class AppKitBaseClient {
       throw new Error(`Adapter for namespace ${namespace} doesn't exist`)
     }
 
-    const networkToRemove = this.caipNetworks?.find(n => n.id === networkId)
+    const networkToRemove = this.getCaipNetworks()?.find(n => n.id === networkId)
+
     if (!networkToRemove) {
-      throw new Error(`Network with ID ${networkId} not found`)
```
magiziz, line 1861: Should we not throw ?

magiziz (approved):

packages/appkit/src/client/appkit-base-client.ts
enesozturk: I've refactored the Demo app's use case a bit. We are not adding removing adapters any more as it's creating problems. Instead the adapter always exist, but we are enable/disable networks. So there is a case where adapter exist but no networks available. For this case, we should still use count that adapter's namespace to initialize adapters

packages/appkit/src/client/appkit-base-client.ts
enesozturk: No actually yeah - updating

packages/appkit/src/client/appkit-base-client.ts
enesozturk: If there is no network it's ok to do nothing

PR #4167, opened 2025-04-08
fix: redundant flows - first batch

ganchoradkov:
# Description
Reworked several connection/reconnection flows to fix several race conditions causing wasted work and unreability.
Refactored the following adapters
-  `ethers`
-  `ethers5`
- `wagmi`
- `solana`

## Type of change

# Associated Issues

For GH issues: closes #...

# Showcase (Optional)

# Checklist

### DIFF
--- a/packages/adapters/ethers/src/client.ts
+++ b/packages/adapters/ethers/src/client.ts
@@ -4,6 +4,7 @@ import { InfuraProvider, JsonRpcProvider, formatEther } from 'ethers'
 import { type AppKitOptions, WcConstantsUtil } from '@reown/appkit'
 import { ConstantsUtil as CommonConstantsUtil, ParseUtil } from '@reown/appkit-common'
 import {
+  AccountController,
   type CombinedProvider,
   type Connector,
   type ConnectorType,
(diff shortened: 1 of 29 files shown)


packages/adapters/ethers/src/client.ts
```diff
@@ -696,4 +703,11 @@ export class EthersAdapter extends AdapterBlueprint {
       params: [params]
     })
   }
+
+  private getPreferredAccountType() {
+    return (
+      AccountController.state.preferredAccountType ||
+      OptionsController.state.defaultAccountTypes.eip155
+    )
+  }
```
enesozturk, line 712: I think we should move this function to adapter blueprint and eacy adapter should get the preferred account type like `ChainController.getAccountProp(this.namespace).preferredAccountType` this will be more generic approach

```suggestion
  private getPreferredAccountType() {
    return (
      ChainController.getAccountProp(this.namespace).preferredAccountType ||
     OptionsController.state.defaultAccountTypes[this.namespace]
    )
  }
```

enesozturk: Even more we should save the latest account type to the local storage for each namespace and get the value from there - I've did this in one of my local branches, you can skip and I could implement after we merge this

packages/controllers/src/controllers/ConnectionController.ts
```diff
@@ -154,7 +155,10 @@ export const ConnectionController = {
     if (!authConnector) {
       return
     }
-    await authConnector?.provider.setPreferredAccount(accountType)
+    AccountController.setPreferredAccountType(
```
enesozturk, line 158: Shouldn't we call provider.setPreferredAccount to make the changes on secure site? 🤔

PR #4192, opened 2025-04-10
chore: improve race condition in `w3m-modal` index

svenvoskamp:
# Description

There are cases that the listeners are getting fired at the same time, we get into a race condition. In this PR they handle separated logic and they are not dependent on each other anymore.. Previously `setIsSwitchingNamespace` influenced both listeners, now is only used in 1 listener.

## Type of change

# Associated Issues

For GH issues: closes #...

# Showcase (Optional)

# Checklist


packages/scaffold-ui/src/modal/w3m-modal/index.ts
```diff
@@ -250,37 +250,64 @@ export class W3mModal extends LitElement {
...
+    const wasUnsupportedNetwork =
+      prevCaipNetwork?.name === CommonConstantsUtil.UNSUPPORTED_NETWORK_NAME
     /**
      * If user is on connecting external, there is a case that they might select a connector which is in another adapter.
      * In this case, we are switching both network and namespace. And this logic will be triggered.
      * But we don't want to go back because we are already on the connecting external view.
      */
     const isConnectingExternal = RouterController.state.view === 'ConnectingExternal'
-    // If user is not connected, we should go back
+    // Check connection status based on the address state *before* this update cycle potentially finishes
```
enesozturk, line 273: Oh you went into rabbit hole

packages/scaffold-ui/src/views/w3m-network-switch-view/index.ts
```diff
@@ -141,6 +141,9 @@ export class W3mNetworkSwitchView extends LitElement {
   private async onSwitchNetwork() {
     try {
       this.error = false
+      if (ChainController.state.activeChain !== this.network?.chainNamespace) {
```
enesozturk, line 144: Is it okay to do this in the view side? What if network has changed on the wallet side and appkit started changing network? Prob it's not going to enter here if this view is not open at that moment

svenvoskamp: Good point, will do some more investigation

svenvoskamp: We also change this on `setActiveCaipNetwork` in the chainController. This takes care of this case. But for the case where the network-switch-view is opened, we need the state before, otherwise logic in the listeners is not up to date. 

But for the case where network is switched from the wallet, this is not needed, so it is safe to proceed with this change

apps/laboratory/tests/shared/validators/ModalValidator.ts
```diff
@@ -390,6 +390,9 @@ export class ModalValidator {
   }
 
   async expectToBeConnectedInstantly() {
+    // Wait for the page to be loaded
+    const initializeBoundary = this.page.getByTestId('w3m-page-loading')
+    await expect(initializeBoundary).toBeHidden()
```
ganchoradkov, line 395: do we need a timeout? what if it doesn't get hidden?

svenvoskamp: AFAIK playwright has a default timeout of 5s on expect assertions

enesozturk (approved):

PR #4195, opened 2025-04-10
fix: viem transport fallback for non Blockchain API requests.

tomiir:
# Description

- Fixes issue where transports would not be properly set to default fallbacks when not supported by blockchain api
- Fixes unhandled throw on `getBalance`  in wagmi client
- Fixes issue where initialize flag would be set after syncing connections.

## Type of change

# Associated Issues

APKT-2624

# Checklist

### DIFF
--- a/.changeset/seven-wombats-glow.md
+++ b/.changeset/seven-wombats-glow.md
@@ -1,0 +1,24 @@
+---
+'@reown/appkit-adapter-wagmi': patch
+'@reown/appkit-utils': patch
+'@reown/appkit': patch
+'@reown/appkit-adapter-bitcoin': patch
+'@reown/appkit-adapter-ethers': patch
+'@reown/appkit-adapter-ethers5': patch
+'@reown/appkit-adapter-solana': patch
+'@reown/appkit-cdn': patch
+'@reown/appkit-cli': patch
+'@reown/appkit-common': patch
+'@reown/appkit-controllers': patch
+'@reown/appkit-core': patch
+'@reown/appkit-experimental': patch
+'@reown/appkit-polyfills': patch
+'@reown/appkit-scaffold-ui': patch
+'@reown/appkit-siwe': patch
+'@reown/appkit-siwx': patch
+'@reown/appkit-ui': patch
+'@reown/appkit-wallet': patch
+'@reown/appkit-wallet-button': patch
+---
+
+Fixes issue wherenetworks not supported by Blockchain API would fail to execute requests in wagmi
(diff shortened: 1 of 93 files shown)


.changeset/seven-wombats-glow.md
```diff
@@ -0,0 +1,24 @@
...
+'@reown/appkit-polyfills': patch
+'@reown/appkit-scaffold-ui': patch
+'@reown/appkit-siwe': patch
+'@reown/appkit-siwx': patch
+'@reown/appkit-ui': patch
+'@reown/appkit-wallet': patch
+'@reown/appkit-wallet-button': patch
+---
+
+Fixes issue wherenetworks not supported by Blockchain API would fail to execute requests in wagmi
```
enesozturk, line 24:
```suggestion
Fixes issue where networks not supported by Blockchain API would fail to execute requests in wagmi
```
^

packages/adapters/wagmi/src/client.ts
```diff
@@ -593,20 +593,26 @@ export class WagmiAdapter extends AdapterBlueprint {
...
+            StorageUtil.updateNativeBalanceCache({
+              caipAddress,
+              balance: balance.formatted,
+              symbol: balance.symbol,
+              timestamp: Date.now()
+            })
+            resolve({ balance: balance.formatted, symbol: balance.symbol })
+          } catch (error) {
+            // eslint-disable-next-line no-console
+            console.error('>> Error getting balance', error)
```
enesozturk, line 613: Should we keep this error? If yes maybe write a better format that indicates it's coming from AppKit?
Maybe smth like 
```suggestion
            console.error('AppKit: Error getting balance', error)
```

packages/appkit-utils/src/CaipNetworkUtil.ts
```diff
@@ -209,14 +209,15 @@ export const CaipNetworksUtil = {
   getViemTransport(caipNetwork: CaipNetwork, projectId: string, customRpcUrls?: CustomRpcUrl[]) {
     const transports: HttpTransport[] = []
 
+    // Add custom RPC URLs
     customRpcUrls?.forEach(rpcUrl => {
       transports.push(http(rpcUrl.url, rpcUrl.config))
     })
 
+    // Add Reown RPC URL
     if (WC_HTTP_RPC_SUPPORTED_CHAINS.includes(caipNetwork.caipNetworkId)) {
```
enesozturk, line 218: Also this is a static list prob we'd better to check out if there is an update to it (if new networks has been added or not)

enesozturk (approved):

packages/adapters/wagmi/src/client.ts
tomiir: Unsure to be hones. Like we shouldn't really log but it's also not good to have an error and have no clue where it came from. 
Will reduce to warn and add some text to facilitate recognition

packages/appkit-utils/src/CaipNetworkUtil.ts
tomiir: Tried to use blockchainapicontroller but it made all methods async and broke a bunch of flows :/. keeping as is for now

svenvoskamp (approved):

PR #4281, opened 2025-04-28
chore: add cosmos namespace

magiziz:
# Description

Added cosmos namespaces

## Type of change

# Associated Issues

For Linear issues: Closes APKT-2593

# Checklist


enesozturk (approved):

packages/controllers/src/utils/AssetUtil.ts
```diff
@@ -19,7 +19,9 @@ const namespaceImageIds: Record<ChainNamespace, string> = {
   // Polkadot
   polkadot: '',
   // Bitcoin
-  bip122: '0b4838db-0161-4ffe-022d-532bf03dba00'
+  bip122: '0b4838db-0161-4ffe-022d-532bf03dba00',
+  // Cosmos
+  cosmos: ''
```
tomiir, line 24: we should add these if possible

tomiir (approved):
This is OK for AK types but in Core we should be able to use whatever

packages/controllers/src/utils/AssetUtil.ts
magiziz: talked this with Enes about this and since we only use those images in multichain view we can skip them unless we'll support cosmos connectors

magiziz:
> This is OK for AK types but in Core we should be able to use whatever

+1

PR #4294, opened 2025-04-29
feat: add `experimental_preferUniversalLinks` property to switch to Universal Links by default for mobile linking

enesozturk:
# Description

Introduces `enableUniversalLinks` property to let users default to Universal Link as a first approach to trigger mobile wallet links.

- Updates connecting WC mobile UI to have similar experience with web wallets. 
   - There is an "Open" button instead of "Try again" and this button gets enabled when the WC connection established.
- Adds `experimental_preferUniversalLinks` property where uses `link_mode` property of the wallets if enabled and if that value exist.
   - If this enabled and the wallet doesn't have UL, it'll fallback to deep link as default method. 

## Type of change

# Associated Issues

For GH issues: closes #...

# Showcase (Optional)

In App Browser when the deep link is allow listed:

https://github.com/user-attachments/assets/2dd86b47-4c8c-497e-a343-82429adfc4f0

In App Browser when the deep link is not allow listed and Universal Link is **not** enabled:

https://github.com/user-attachments/assets/2b9c33b7-8301-4bd3-82e2-cc9ba9543b18

In App Browser when the deep link is not allow listed and Universal Link is enabled: 

https://github.com/user-attachments/assets/8c7bd7c6-ad14-4bef-aa47-659f82875ac9

Default OS browser, using "Open" function to re-trigger the deeplink (or UL if enabled):

https://github.com/user-attachments/assets/1e5d3817-ecfb-4836-aed9-1553261fb7f0

# Checklist


packages/controllers/src/utils/CoreHelperUtil.ts
```diff
@@ -117,18 +117,31 @@ export const CoreHelperUtil = {
...
     if (!safeAppUrl.includes('://')) {
       safeAppUrl = appUrl.replaceAll('/', '').replaceAll(':', '')
       safeAppUrl = `${safeAppUrl}://`
     }
+
     if (!safeAppUrl.endsWith('/')) {
       safeAppUrl = `${safeAppUrl}/`
     }
+
+    if (!safeUniversalLink?.endsWith('/')) {
```
tomiir, line 141: shouldn't you check that `universalLink != null` ?

arein (commented):
Universal Link shouldn't be the "default" most likely. Instead only the fallback. Universal Links only work proper if triggered immediately after user interaction. When using Wagmi this is unpredictable as often there are network calls before triggering the Universal Link. That's why a few years back we defaulted to Deep Links.

tomiir (approved):

arein:
https://github.com/user-attachments/assets/b0d01d09-025c-4b25-86da-8634a73ddb89


The `Open` button is clickable even when the websocket has not connected in the default mobile view (Not universal).

The Universal Link demo is quite smooth

packages/scaffold-ui/src/partials/w3m-connecting-wc-mobile/index.ts
```diff
@@ -32,47 +58,45 @@ export class W3mConnectingWcMobile extends W3mConnectingWidget {
...
   public override disconnectedCallback() {
     super.disconnectedCallback()
-    document.removeEventListener('visibilitychange', this.onBuffering.bind(this))
     clearTimeout(this.btnLabelTimeout)
-    clearTimeout(this.labelTimeout)
   }
 
   // -- Private ------------------------------------------- //
-
-  private initializeStateAndTimers() {
```
svenvoskamp, line 42: Is it ok to remove this? Could there be a case that we await the request from the wallet and the shows `active` state in stead of loading?

enesozturk: The WC URI still awaited in the constructor, we do here just suing "Open" button instead of "Try again" which is resolves no matter what after 5 secs, which is not really good UX - I'll still double check

svenvoskamp (approved):

PR #4296, opened 2025-04-30
fix: pre load iframe in PWA environments with social login enabled

svenvoskamp:
# Description

🐛  **Issue**

Social login doesn't work in PWA environments

🫚 **Root Cause**

The issue happens because we open a new tab immediately while the iframe is not yet loaded. When this happens, we lose the connection with the iframe. 

👨‍🔧 **Solution**

This PR fixes this issue by checking if the app is loaded in a PWA environment. If this is the case, we preload the iframe and disable the social buttons while loading the iframe. After the iframe has been loaded, the buttons become active and social login works as it should. 

## Type of change

# Associated Issues

For Linear issues: Closes APKT-1653
For GH issues: closes #...

# Showcase (Optional)

# Checklist


packages/controllers/src/utils/CoreHelperUtil.ts
```diff
@@ -197,6 +197,17 @@ export const CoreHelperUtil = {
     )
   },
 
+  isPWA() {
+    if (typeof window === 'undefined') {
+      return false
+    }
+
+    const isStandaloneDisplayMode = window.matchMedia?.('(display-mode: standalone)').matches
+    const isIOSStandalone = (window.navigator as unknown as { standalone: boolean }).standalone
```
tomiir, line 206: can we optional chain this? seems potentially breaking

packages/scaffold-ui/src/partials/w3m-social-login-widget/index.ts
```diff
@@ -202,6 +219,27 @@ export class W3mSocialLoginWidget extends LitElement {
...
+          await this.authConnector.provider.init()
+        }
+      } catch (error) {
+        AlertController.open(
+          {
+            shortMessage: 'Error loading embedded wallet in PWA',
+            longMessage: (error as Error).message
+          },
+          'error'
+        )
```
tomiir, line 237: will this alert collide with the one regarding misconfigurations / timeouts?

svenvoskamp: Misconfiguration will show immediately, timeout one will happen after 20 seconds (in new PR). This will show if we catch an error on it, so AFAIK this won't collide.

PR #4342, opened 2025-05-13
fix: connected wallet info sets connector id if connector or provider not found

enesozturk:
# Description

AppKit does set connector ID as connected wallet info if the connector or provider not found, which is misleading value bc if these are not found it should keep/set the connected value as undefined

## Type of change

# Associated Issues

For GH issues: closes #...

# Showcase (Optional)

# Checklist

### DIFF
--- a/.changeset/vast-goats-dream.md
+++ b/.changeset/vast-goats-dream.md
@@ -1,0 +1,25 @@
+---
+'@reown/appkit': patch
+'@reown/appkit-adapter-bitcoin': patch
+'@reown/appkit-adapter-ethers': patch
+'@reown/appkit-adapter-ethers5': patch
+'@reown/appkit-adapter-solana': patch
+'@reown/appkit-adapter-wagmi': patch
+'@reown/appkit-utils': patch
+'@reown/appkit-cdn': patch
+'@reown/appkit-cli': patch
+'@reown/appkit-common': patch
+'@reown/appkit-controllers': patch
+'@reown/appkit-core': patch
+'@reown/appkit-experimental': patch
+'@reown/appkit-pay': patch
+'@reown/appkit-polyfills': patch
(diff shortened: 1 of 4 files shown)


.changeset/vast-goats-dream.md
```diff
@@ -0,0 +1,25 @@
...
+'@reown/appkit-core': patch
+'@reown/appkit-experimental': patch
+'@reown/appkit-pay': patch
+'@reown/appkit-polyfills': patch
+'@reown/appkit-scaffold-ui': patch
+'@reown/appkit-siwe': patch
+'@reown/appkit-siwx': patch
+'@reown/appkit-ui': patch
+'@reown/appkit-wallet': patch
+'@reown/appkit-wallet-button': patch
```
arein, line 22: nitpick: this is `minor`? is there a chance apps' implementations relies on how this currently works? if so `minor` else `patch`

enesozturk: Good point. I think this is still patch. In terms of the values that `walletInfo` gets over time after page refresh:

- First render: `undefined`
- Second render: `walletConnect`
- Third render: `Safe{Wallet}` (assuming connected with safe)

After this PR:

- First render: `undefined`
- Second render: `Safe{Wallet}` (assuming connected with safe)

If somebody expecting `walletInfo.name` to be `walletConnect`, they already know that this immediately turns to another value. This means they probably will use this only for watching the value to trigger something, not to render something to the screen depending on this value.

But this is a good point and I'd like to ask to the @reown-com/appkit-web-team wdyt?

tomiir: I think this is still a patch as the end behavior is still the same, avoiding a glitchy behavior

tomiir (approved):

PR #4358, opened 2025-05-16
chore: upgrade valtio to v2

enesozturk:
- **chore: upgrade valtio**
- **chore: upgrade Valtio to v2**
- **chore: changeset file**

# Description

## Type of change

# Associated Issues

For GH issues: closes #...

# Showcase (Optional)

# Checklist


packages/appkit/tests/client/initialization.test.ts
```diff
@@ -105,6 +105,20 @@ describe('Base', () => {
       expect(setEIP6963Enabled).toHaveBeenCalledWith(false)
     })
 
+    it('should use default account types when no account types are set', async () => {
```
enesozturk, line 108: Somehow there is a race condition that running this test after the `should set default account types` one generates wrong result

lukaisailovic:
@enesozturk Are we exposing valtio state somewhere publicly? We shouldn't do it directly, but if we do, do the customers need to use `ref` then?

lukaisailovic (approved):

enesozturk:
> @enesozturk Are we exposing valtio state somewhere publicly? We shouldn't do it directly, but if we do, do the customers need to use `ref` then?

We are exposing state only via hooks and plain JS functions, but the using `ref()` only matter when you create state/proxy - so users won't get any issues using the exposed values

PR #4438, opened 2025-05-30
fix: add chain fallback to get correct embeddedWalletInfo on `getAccount`

svenvoskamp:
# Description

In the `getAccount` public method, we do a check if the `activeConnectorId` is valid for auth connector. But to get that id we needed to have a provided namespace. This is not great DX, so if no namespace is provided, we fallback to the current active chain id. 

## Type of change

# Associated Issues

For Linear issues: Closes APKT-2851

# Showcase (Optional)

# Checklist


packages/appkit/src/client/appkit-base-client.ts
```diff
@@ -1715,7 +1715,7 @@ export abstract class AppKitBaseClient {
     const authConnector = ConnectorController.getAuthConnector(namespace)
     const accountState = ChainController.getAccountData(namespace)
     const activeChain = ChainController.state.activeChain as ChainNamespace
-    const activeConnectorId = StorageUtil.getConnectedConnectorId(namespace)
+    const activeConnectorId = StorageUtil.getConnectedConnectorId(namespace || activeChain)
```
tomiir, line 1718: Maybe we should even remove the storage fetch here? storage should not be used besides initialization in general

A bit scared it will break something tho tbch

tomiir (approved):

PR #4444, opened 2025-05-30
fix: use workspace relative url when uploading artifacts

tomiir:
# Description

- `merge_reports` is failing when `ui_tests` is executed from external repository as no artifacts are uploaded for it to merge
- Artifacts were not being uploaded because the action looks for a local relative path that does not exist in external repository 

## Type of change


enesozturk (approved):

.github/workflows/ui_tests.yml
```diff
@@ -357,11 +365,14 @@ jobs:
 
   merge_reports:
     needs: ui_tests # Ensures this job runs only after all 'test' (sharded) jobs complete
-    if: always() && inputs.build-and-cache-only == false # Run this job if ui_tests ran (even if it failed), but not if ui_tests was skipped.
+    if: always() && inputs.build-and-cache-only == false' # Run this job if ui_tests ran (even if it failed), but not if ui_tests was skipped.
```
tomiir, line 368:
```suggestion
    if: always() && inputs.build-and-cache-only == false # Run this job if ui_tests ran (even if it failed), but not if ui_tests was skipped.
```

PR #4449, opened 2025-06-02
feat: cloud auth email capture flow

zoruka:
# Description

## Type of change

# Associated Issues

For GH issues: closes #...

# Showcase (Optional)

# Checklist


packages/wallet/src/W3mFrame.ts
```diff
@@ -16,28 +16,29 @@ interface W3mFrameConfig {
...
   chainId?: W3mFrameTypes.Network['chainId']
   enableLogger?: boolean
+  enableCloudAuthAccount?: boolean
   rpcUrl?: string
 }
 
 function createSecureSiteSdkUrl({
   projectId,
   chainId,
-  version,
```
tomiir, line 25: why did you remove the version from here?

packages/siwx/tests/ui/w3m-data-capture-otp-confirm-view.test.ts
```diff
@@ -0,0 +1,384 @@
...
+    getAuthConnector: vi.fn()
+  }
+}))
+
+describe('W3mDataCaptureOtpConfirmView', () => {
+  let element: W3mDataCaptureOtpConfirmView
+  let mockSiwx: ReownAuthentication
+
+  beforeEach(() => {
+    vi.clearAllMocks()
```
tomiir: we prefer using `restoreAllMocks` as `clear` will not wipe spy calls

packages/siwx/src/ui/w3m-data-capture-view/index.ts
```diff
@@ -0,0 +1,296 @@
...
+
+      if (otp.uuid === null) {
+        RouterController.replace('SIWXSignMessage')
+      } else {
+        RouterController.replace('DataCaptureOtpConfirm', { email: this.email })
+      }
+
+      await new Promise(resolve => {
+        // Wait for view transition
+        setTimeout(resolve, 300)
```
tomiir: do we need this?

packages/siwx/src/ui/w3m-data-capture-view/index.ts
```diff
@@ -0,0 +1,296 @@
...
+      this.loading = false
+    }
+  }
+
+  private onSkip() {
+    RouterController.replace('SIWXSignMessage')
+  }
+
+  private getRecentEmails() {
+    const recentEmails = SafeLocalStorage.getItem(SafeLocalStorageKeys.RECENT_EMAILS)
```
tomiir, line 271: is this only used here? we are trying not to read from storage in the views.

packages/siwx/src/ui/w3m-data-capture-view/index.ts
```diff
@@ -0,0 +1,296 @@
...
+  @state() private isRequired =
+    Array.isArray(OptionsController.state.remoteFeatures?.emailCapture) &&
+    OptionsController.state.remoteFeatures?.emailCapture.includes('required')
+
+  @state() private recentEmails = this.getRecentEmails()
+
+  public override connectedCallback() {
+    if (!this.siwx || !(this.siwx instanceof ReownAuthentication)) {
+      throw new Error('ReownAuthentication is not initialized')
+    }
```
tomiir: what happens when this throws?

packages/siwx/package.json
```diff
@@ -52,6 +59,9 @@
     "vitest": "3.1.3",
     "vue": "3.5.13"
   },
+  "peerDependencies": {
+    "lit": "3.3.0"
+  },
```
tomiir, line 64: why?

packages/controllers/src/utils/ModalUtil.ts
```diff
@@ -25,6 +26,13 @@ export const ModalUtil = {
       return
     }
 
+    if (
+      RouterController.state.view === 'DataCapture' ||
+      RouterController.state.view === 'DataCaptureOtpConfirm'
+    ) {
+      ConnectionController.disconnect()
```
tomiir, line 33: is this clearing the session already ?

packages/appkit/package.json
```diff
@@ -159,6 +159,9 @@
     "vitest": "3.1.3",
     "vue": "3.x"
   },
+  "optionalDependencies": {
+    "@reown/appkit-siwx": "workspace:*"
+  },
```
tomiir: is this needed?

packages/wallet/src/W3mFrame.ts
zoruka: This was only receiving a constant as a param, I've removed the param and used the constant directly. I may revert it if you prefer how it was.

packages/siwx/src/ui/w3m-data-capture-view/index.ts
zoruka: This was the only way I figured out to avoid blinking the view. As it finishes this promise, it goes for finally which removes the loading state before the view transition is over. Should we only remove the loading state on error? Would it be fine? I'm not sure exactly how to treat the view behavior for that scenario.

packages/siwx/src/ui/w3m-data-capture-view/index.ts
zoruka: It is only here. Should `recentEmails` come from component property? How would you suggest to do it if not reading localStorage directly?

packages/siwx/src/ui/w3m-data-capture-view/index.ts
zoruka: It's a blocking error, the UI won't work. My take is that if it reaches this state it means there is a miss configuration on AppKit, so we should not move forward, wdyt?

packages/siwx/package.json
zoruka: Did not get the question. Why is it peer? or why adding it?
It is peer because main packages will have it installed and we are adding it because there are new components being implemented in siwx package.

packages/controllers/src/utils/ModalUtil.ts
zoruka: At this point there is no session created yet.
`connection` => `data-capture` => `sign-message` (with captured data) => `session-created`

packages/appkit/package.json
zoruka: It's needed due to this https://github.com/reown-com/appkit/blob/cc47f0f6f88d9f53473a52bacfc46c56b46201ce/packages/appkit/src/client/appkit.ts#L660

packages/siwx/tests/ui/w3m-data-capture-otp-confirm-view.test.ts
zoruka: https://github.com/reown-com/appkit/pull/4449/commits/731c0612c54e33c4dc80bdda7d9d3ce5b37da0fb

packages/siwx/src/configs/ReownAuthenticationSIWX.ts
```diff
@@ -77,25 +84,35 @@ export class ReownAuthentication implements SIWXConfig {
         return []
       }
 
-      const siweSession = await this.request('me', undefined, 'authJwt')
+      const account = await this.request({
```
enesozturk, line 88: This isn ot a breaking change for old users by any chance right?

zoruka: its not a breaking change, I've renamed only to be more aligned with what the data is

packages/siwx/src/ui/w3m-data-capture-view/index.ts
enesozturk: Why don't we block the actions in this page (i.e block the buttons) and show AlertBar on the top of the page instead? @zoruka 

> It's a blocking error, the UI won't work

Is this mean that it'll show blank screen?

packages/siwx/src/ui/w3m-data-capture-otp-confirm-view/index.ts
```diff
@@ -0,0 +1,48 @@
...
+
+import { ReownAuthentication } from '../../configs/index.js'
+
+@customElement('w3m-data-capture-otp-confirm-view')
+export class W3mDataCaptureOtpConfirmView extends W3mEmailOtpWidget {
+  @state() private siwx = OptionsController.state.siwx as ReownAuthentication
+
+  public override connectedCallback() {
+    if (!this.siwx || !(this.siwx instanceof ReownAuthentication)) {
+      throw new Error('ReownAuthentication is not initialized')
```
enesozturk: Same here

enesozturk (approved):

packages/siwx/src/ui/w3m-data-capture-otp-confirm-view/index.ts
zoruka: I don't think this would be a breaking change since it's a new component that requires ReownAuthentication to be in use through siwx setting

packages/siwx/src/ui/w3m-data-capture-view/index.ts
zoruka: Throwing the error would not call directly the toast in the default error treatment of the modal? I can implement this local in the component too. Anyway this should be a state where the modal should never reach.

packages/siwx/src/ui/w3m-data-capture-view/index.ts
tomiir: Proper way would probably be to do it during intitialization but not sure you have a state where you persist siwx things

packages/siwx/src/ui/w3m-data-capture-view/index.ts
tomiir: No, you need to call `SnackController` to show the errror

packages/siwx/src/ui/w3m-data-capture-view/index.ts
tomiir: You can use `requestUpdate` for this!

packages/siwx/src/ui/w3m-data-capture-view/index.ts
zoruka: https://github.com/reown-com/appkit/pull/4449/commits/e6c71f60659d5df9d661311b6482822e22ef6b5e

packages/siwx/src/ui/w3m-data-capture-view/index.ts
zoruka: https://github.com/reown-com/appkit/pull/4449/commits/12f13f7e25ee023aac9039898782b306b97caf39

PR #4450, opened 2025-06-02
feat(multi-wallet): refactor controllers, storage and add hooks

magiziz:
# Description

- Refactored `ConnectionController.ts` and `StorageUtil.ts`
- Added `useAppKitConnection` and `useAppKitConnections` hook
- Added `getConnections`, `getStorageConnections`, `switchConnection`, `deleteConnection` and `subscribeConnections` public methods

## Type of change

# Checklist


packages/appkit/src/client/appkit-base-client.ts
```diff
@@ -1599,6 +1603,18 @@ export abstract class AppKitBaseClient {
...
+    ConnectionControllerUtil.getConnectionsData(namespace).connections
+
+  public getStorageConnections = (namespace: ChainNamespace) =>
+    ConnectionControllerUtil.getConnectionsData(namespace).storageConnections
+
+  public switchConnection: (typeof ConnectionController)['switchConnection'] = params =>
+    ConnectionController.switchConnection(params)
+
+  public deleteConnection: (typeof StorageUtil)['deleteAddressFromConnection'] = params =>
+    StorageUtil.deleteAddressFromConnection(params)
```
0xmkh: These are new methods that can be used if someone decides to implement multi-wallet in plain js

packages/common/src/utils/SafeLocalStorage.ts
```diff
@@ -50,7 +51,8 @@ export const SafeLocalStorageKeys = {
   ENS_CACHE: '@appkit/ens_cache',
   IDENTITY_CACHE: '@appkit/identity_cache',
   PREFERRED_ACCOUNT_TYPES: '@appkit/preferred_account_types',
-  CONNECTIONS: '@appkit/connections'
+  CONNECTIONS: '@appkit/connections',
+  DISCONNECTED_CONNECTOR_IDS: '@appkit/disconnected_connector_ids'
```
0xmkh, line 55: The reason why we have this new field is to remember what connectors were disconnected so we don't connect to them again during page refresh. This hack is called "shim disconnection" which is also used by wagmi

packages/controllers/exports/react.ts
```diff
@@ -72,9 +101,124 @@ export function useAppKitAccount(options?: { namespace?: ChainNamespace }): UseA
...
+
+  const { activeChain } = useSnapshot(ChainController.state)
+
+  const chainNamespace = namespace ?? activeChain
+
+  if (!chainNamespace) {
+    throw new Error('No namespace found')
+  }
+
+  const { connections, storageConnections } =
```
0xmkh: React hooks

packages/controllers/exports/vue.ts
```diff
@@ -83,9 +129,233 @@ export function useAppKitAccount(options?: {
...
+      name,
+      icon,
+      networkIcon: networkImage,
+      ...connection
+    }
+  }
+
+  function updateConnections() {
+    const chainNamespace = namespace ?? ChainController.state.activeChain
+
```
0xmkh: Vue hooks

packages/controllers/src/controllers/AccountController.ts
```diff
@@ -107,6 +108,10 @@ const controller = {
     return ChainController.getAccountProp('caipAddress', chain)
   },
 
+  getSmartAccountAddress(chain: ChainNamespace | undefined) {
+    return ChainController.getAccountProp('smartAccountAddress', chain)
+  },
```
0xmkh: This field is used to show a smart account UI icon badge if the wallet is a smart account

packages/controllers/src/controllers/ConnectionController.ts
```diff
@@ -27,36 +34,81 @@ import { TransactionsController } from './TransactionsController.js'
...
+
+export interface DisconnectAllParameters {
+  chainNamespace?: ChainNamespace
+}
+
 export interface ConnectionControllerClient {
   connectWalletConnect?: () => Promise<void>
-  disconnect: (chainNamespace?: ChainNamespace) => Promise<void>
+  disconnect: (params?: DisconnectParameters) => Promise<void>
+  disconnectAll?: (params?: DisconnectAllParameters) => Promise<void>
```
0xmkh: Adapter will also have a new method called `disconnectAll`. This will be used to disconnect all connectors at once, and `disconnect` will be responsible to disconnect connector in one by one order

packages/controllers/src/controllers/ConnectionController.ts
```diff
@@ -152,11 +210,13 @@ const controller = {
...
 
   async connectExternal(options: ConnectExternalOptions, chain: ChainNamespace, setChain = true) {
-    await ConnectionController._getClient()?.connectExternal?.(options)
+    const connectData = await ConnectionController._getClient()?.connectExternal?.(options)
 
     if (setChain) {
       ChainController.setActiveNamespace(chain)
     }
+
+    return connectData
```
0xmkh, line 224: This is needed so we can later show users a toast message, e.g `Connect with "0x...123"` after they're connected

packages/controllers/src/controllers/ConnectionController.ts
```diff
@@ -305,39 +365,216 @@ const controller = {
...
+    const status = ConnectionControllerUtil.getConnectionStatus(connection, namespace)
+
+    switch (status) {
+      case 'connected':
+      case 'active': {
+        const newAddress = await ConnectionController.handleActiveConnection({
+          connection,
+          namespace,
+          address
+        })
```
0xmkh, line 542: This is the most important place to look at if you want to see how we handle current, active and disconnected connections

packages/controllers/src/utils/ConnectionControllerUtil.ts
```diff
@@ -0,0 +1,126 @@
...
+    })
+  },
+  getConnectionsData(namespace: ChainNamespace) {
+    const connections = ConnectionController.state.connections.get(namespace) ?? []
+
+    const activeConnectorId = ConnectorController.state.activeConnectorIds[namespace]
+
+    const storageConnections = StorageUtil.getConnections()
+    const storageConnectionsByNamespace = storageConnections[namespace] ?? []
+    const storageConnectionsWithCurrentActiveConnectors = storageConnectionsByNamespace.filter(
```
0xmkh: These are helper utils for the `ConnectionController`

packages/controllers/src/utils/ConnectorControllerUtil.ts
```diff
@@ -1,13 +1,354 @@
...
+    onConnect
+  }: ConnectWalletConnectParameters): Promise<ParsedCaipAddress> {
+    return new Promise(async (resolve, reject) => {
+      if (walletConnect) {
+        ConnectorController.setActiveConnector(connector)
+      }
+
+      onOpen?.(CoreHelperUtil.isMobile() && walletConnect)
+
+      if (redirectViewOnModalClose) {
```
0xmkh: Ignore this code, i copied it from wallet button file (will replace it there too once this PR gets merged)

packages/controllers/src/utils/StorageUtil.ts
```diff
@@ -589,12 +597,34 @@ export const StorageUtil = {
...
+        const isAuth = conn.connectorId === CommonConstantsUtil.CONNECTOR_ID.AUTH
+
+        if (existingConn && !isAuth) {
+          const existingAddrs = new Set(existingConn.accounts.map(a => a.address.toLowerCase()))
+          const newAccounts = conn.accounts.filter(a => !existingAddrs.has(a.address.toLowerCase()))
+          existingConn.accounts.push(...newAccounts)
+        } else {
+          mergedMap.set(conn.connectorId, { ...conn })
+        }
       }
```
0xmkh: Storage connections are complicated 🥱. To keep it short and simple, once we connect with a connector we need to store all current and new accounts in storage so later users can reconnect to them

packages/appkit/src/client/appkit-base-client.ts
```diff
@@ -471,7 +473,7 @@ export abstract class AppKitBaseClient {
         })
 
         if (!res) {
-          return
+          return undefined
```
tomiir, line 477: isnt this redundant?

packages/common/src/utils/SafeLocalStorage.ts
tomiir: curious: why do we need this? shouldn't we be checking which connectors were connected instead?

packages/controllers/exports/react.ts
```diff
@@ -1,13 +1,42 @@
...
+export type { Connection } from '../src/controllers/ConnectionController.js'
+
+interface UseAppKitConnectionProps {
+  namespace?: ChainNamespace
+  onSuccess?: (params: {
+    address: string
+    namespace: ChainNamespace
+    isAccountSwitched: boolean
+    isWalletSwitched: boolean
+    isWalletDeleted: boolean
```
tomiir: nit: I'd use `has` instead of `is`

packages/controllers/exports/react.ts
```diff
@@ -72,9 +101,124 @@ export function useAppKitAccount(options?: { namespace?: ChainNamespace }): UseA
 }
 
 export function useDisconnect() {
-  async function disconnect(props?: { namespace?: ChainNamespace }) {
-    await ConnectionController.disconnect(props?.namespace)
+  async function disconnect(props?: { connectorId?: string; namespace?: ChainNamespace }) {
```
tomiir: can we do a type for this?

packages/controllers/exports/react.ts
```diff
@@ -72,9 +101,124 @@ export function useAppKitAccount(options?: { namespace?: ChainNamespace }): UseA
...
   }
 
   return { disconnect }
 }
+
+export function useAppKitConnections(namespace?: ChainNamespace) {
+  // Snapshots to trigger re-renders on state changes
+  useSnapshot(ConnectionController.state)
+  useSnapshot(ConnectorController.state)
+  useSnapshot(AssetController.state)
```
tomiir, line 120: why exactly is this needed?

packages/controllers/exports/react.ts
```diff
@@ -72,9 +101,124 @@ export function useAppKitAccount(options?: { namespace?: ChainNamespace }): UseA
...
+  const chainNamespace = namespace ?? activeChain
+
+  if (!chainNamespace) {
+    throw new Error('No namespace found')
+  }
+
+  const { connections, storageConnections } =
+    ConnectionControllerUtil.getConnectionsData(chainNamespace)
+
+  const formatConnection = useCallback((connection: Connection) => {
```
tomiir, line 133: why do you need a useCallback here? it has no dependencies

packages/controllers/exports/react.ts
```diff
@@ -72,9 +101,124 @@ export function useAppKitAccount(options?: { namespace?: ChainNamespace }): UseA
...
+  if (!chainNamespace) {
+    throw new Error('No namespace found')
+  }
+
+  const connection = useMemo(() => {
+    const connectorId = activeConnectorIds[chainNamespace]
+    const connList = connections.get(chainNamespace)
+
+    return connList?.find(c => c.connectorId.toLowerCase() === connectorId?.toLowerCase())
+  }, [connections, activeConnectorIds, chainNamespace])
```
enesozturk: I'm not sure useMemo would work here as expected. Since `connections` and `activeConnectorIds` are not primitive values (they are arrays), even tho they don't have same value, `useMemo` will still be recalculated

packages/controllers/exports/react.ts
```diff
@@ -72,9 +101,124 @@ export function useAppKitAccount(options?: { namespace?: ChainNamespace }): UseA
...
+
+  const switchConnection = useCallback(
+    async ({ connection: _connection, address }: SwitchConnectionParams) => {
```
enesozturk, line 172: Why renaming this as _connection?

packages/controllers/exports/react.ts
0xmkh: mhmm i think they're still stored in a single context even if they're array since its context doesn't change it won't cause useMemo to recalculate, but will remove it 🙏

packages/controllers/exports/react.ts
0xmkh: `connection` is already defined on top

packages/controllers/exports/vue.ts
```diff
@@ -83,9 +129,233 @@ export function useAppKitAccount(options?: {
 }
 
 export function useDisconnect() {
-  async function disconnect(props?: { namespace?: ChainNamespace }) {
-    await ConnectionController.disconnect(props?.namespace)
+  async function disconnect(props?: { id?: string; namespace?: ChainNamespace }) {
+    await ConnectionController.disconnect({ id: props?.id, namespace: props?.namespace })
```
tomiir: same can we type this ? seems like same typing across all the hooks

packages/controllers/exports/vue.ts
tomiir: wouldn't it be good to receive an optional namespace parameter here?

packages/controllers/exports/vue.ts
```diff
@@ -83,9 +129,233 @@ export function useAppKitAccount(options?: {
...
+  const state = ref({
+    connection: undefined,
+    isPending: false,
+    switchConnection: () => Promise.resolve(undefined),
+    deleteConnection: () => ({})
+  } as UseAppKitConnectionReturn)
+
+  const forceUpdateCounter = ref(0)
+
+  function updateConnection() {
```
tomiir, line 249: same on the parameter

packages/controllers/src/controllers/AccountController.ts
tomiir: why is smartAccountAddress a prop in account state? 
I don't think that's cool. It assumes that all connections potentially have a smart account, seems like a weird model. 

I'd rather have a utility looking for a 'smartAccount` in the connection accounts.

packages/controllers/src/controllers/ConnectionController.ts
```diff
@@ -27,36 +34,81 @@ import { TransactionsController } from './TransactionsController.js'
...
+    name: string | undefined
+    username: string | undefined
+  }
+}
+
+interface SwitchConnectionParams {
+  connection: Connection
+  address?: string
+  namespace: ChainNamespace
+  closeModalOnConnect?: boolean
```
tomiir, line 42: what is this?

packages/controllers/src/controllers/ConnectionController.ts
tomiir: wouldn't it be better to have it receive an id ?
if id => disconnect specific
no id => disconnect all

packages/controllers/src/controllers/ConnectionController.ts
```diff
@@ -305,39 +365,216 @@ const controller = {
...
 
   setConnections(connections: Connection[], chainNamespace: ChainNamespace) {
-    state.connections.set(chainNamespace, connections)
+    state.connections = new Map(state.connections.set(chainNamespace, connections))
   },
 
-  switchAccount({ connection, address, namespace }: SwitchAccountParams) {
-    const connectedConnectorId = ConnectorController.state.activeConnectorIds[namespace]
-    const isConnectorConnected = connectedConnectorId === connection.connectorId
+  async handleAuthAccountSwitch(address: string, namespace: ChainNamespace) {
```
tomiir: why multiple params instead of object param? it's worse for maintainability

tomiir: also why a method specific for auth?

packages/controllers/exports/vue.ts
```diff
@@ -3,9 +3,11 @@ import { type Ref, onMounted, onUnmounted, ref } from 'vue'
 import { type ChainNamespace, ConstantsUtil } from '@reown/appkit-common'
```
enesozturk, line 3: I think we should have unit and integration tests for the Vue an JS examples too. Unit tests in the packages itself not really enough. Multi-wallet related features should be added to our example repos too

packages/controllers/src/controllers/ConnectionController.ts
```diff
@@ -305,39 +365,216 @@ const controller = {
...
-        console.warn(`No current network found for namespace "${namespace}"`)
+        const { address: emailAddress } = await ConnectorControllerUtil.connectEmail({
+          closeModalOnConnect,
+          onOpen() {
+            ModalController.open({ view: 'EmailLogin' })
+          },
+          onConnect() {
+            RouterController.replace('ProfileWallets')
+          }
+        })
```
tomiir, line 472: Shouldn't connectorcontroller util be generic? these seem very auth specific

packages/controllers/src/controllers/ConnectionController.ts
```diff
@@ -305,39 +365,216 @@ const controller = {
...
+      isAddressSmartAccount && ConnectorController.canSwitchToSmartAccount(namespace)
+        ? 'smartAccount'
+        : 'eoa'
 
-    if (isConnectorConnected) {
-      const currentNetwork = ChainController.state.activeCaipNetwork
+    await ConnectionController.setPreferredAccountType(accountType, namespace)
+  },
+
+  async handleActiveConnection({ connection, namespace, address }: HandleActiveConnectionParams) {
```
tomiir, line 405: this looks like an extremely lengthy complex method. Let's discuss on the call today

packages/controllers/exports/vue.ts
0xmkh: Yeah was thinking about this, maybe we can do similar to what react-testing-library does. Will look into that when adding vue examples 🙏

packages/controllers/src/controllers/ConnectionController.ts
```diff
@@ -305,39 +365,216 @@ const controller = {
...
+            address: newAddress,
+            namespace,
+            isAccountSwitched: true,
+            isWalletSwitched: true
+          })
+        }
+        break
       }
+      default:
+        throw new Error(`Invalid connection status: ${status}`)
```
tomiir, line 576: do we really want to throw? how is this error handled in higher level contexts? anyone catching and disconnecting? will this bubble up to a user error?

packages/controllers/src/utils/ConnectionControllerUtil.ts
```diff
@@ -0,0 +1,126 @@
...
+      }
+
+      const { type } =
+        connection.accounts.find(
+          account => account.address.toLowerCase() === address.toLowerCase()
+        ) ?? {}
+
+      if (typeof type === 'string' && type !== 'payment') {
+        throw new Error(`Switching to non-payment accounts is not allowed for ${namespace}`)
+      }
```
tomiir, line 59: why?

packages/controllers/src/utils/ConnectionControllerUtil.ts
```diff
@@ -0,0 +1,126 @@
...
+
+      return { ...c, accounts: filteredAccounts }
+    })
+  },
+  getConnectionsData(namespace: ChainNamespace) {
+    const connections = ConnectionController.state.connections.get(namespace) ?? []
+
+    const activeConnectorId = ConnectorController.state.activeConnectorIds[namespace]
+
+    const storageConnections = StorageUtil.getConnections()
```
tomiir: why are we querying storage here? we should not query storage outside initialization flows

packages/controllers/src/utils/ConnectorControllerUtil.ts
```diff
@@ -1,13 +1,354 @@
...
-/**
- * Checks if the given namespace is associated with the specified connector id.
- * @param namespace - The namespace to check.
- * @param connectorId - The connector id to compare against.
- * @returns True if the namespace is associated with the connector id, false otherwise.
- */
-export function checkNamespaceConnectorId(namespace: ChainNamespace, connectorId: string): boolean {
-  return ConnectorController.getConnectorId(namespace) === connectorId
+// -- Constants ------------------------------------------ //
+const UPDATE_EMAIL_INTERVAL = 1_000
```
tomiir:
```suggestion
const UPDATE_EMAIL_INTERVAL_MS = 1_000
```

packages/controllers/src/utils/ConnectorControllerUtil.ts
```diff
@@ -1,13 +1,354 @@
...
+        }
+        unsubscribeChainController()
+      }
+    })
+
+    return new Promise((resolve, reject) => {
+      async function handleSocialConnection(event: MessageEvent) {
+        if (event.data?.resultUri) {
+          if (event.origin === CommonConstantsUtil.SECURE_SITE_SDK_ORIGIN) {
+            window.removeEventListener('message', handleSocialConnection, false)
```
tomiir, line 132: why? this might bring huge race conditions

packages/controllers/src/utils/ConnectorControllerUtil.ts
```diff
@@ -1,13 +1,354 @@
...
+              }
+            } catch (err) {
+              reject(new Error('Failed to connect'))
+              if (socialProvider) {
+                EventsController.sendEvent({
+                  type: 'track',
+                  event: 'SOCIAL_LOGIN_ERROR',
+                  properties: { provider: socialProvider }
+                })
+              }
```
tomiir: I'd reject after doing whatever you need to do as cleanup.

packages/controllers/src/utils/ConnectorControllerUtil.ts
```diff
@@ -1,13 +1,354 @@
...
+
+                const interval = setInterval(() => {
+                  if (socialWindow?.closed && !connectingSocial) {
+                    reject(new Error('Popup closed'))
+                    clearInterval(interval)
+                  }
+                }, 1000)
+
+                window.addEventListener('message', handleSocialConnection, false)
+              } else {
```
tomiir, line 264: I don't really like this, was this done this way before? adding listener to `message`  events is really tricky, this event gets fired ad-hoc by A LOT of things, more so in an SDK that is used by other dapps. Scared this might turn into an uncontrolled event storm

packages/controllers/src/utils/ConnectorControllerUtil.ts
```diff
@@ -1,13 +1,354 @@
...
+      const interval = setInterval(() => {
+        const newEmail = authConnector.provider.getEmail() ?? ''
+
+        if (newEmail !== initialEmail) {
+          ModalController.close()
+          clearInterval(interval)
+          unsubscribeModalController()
+          resolve({ email: newEmail })
+        }
+      }, UPDATE_EMAIL_INTERVAL)
```
tomiir: why do we need this ?

0xmkh: This code was taken from wallet button. Mentioned above that i'll take these functions and replace them in wallet button lib as well

packages/controllers/src/utils/ConnectorControllerUtil.ts
0xmkh: this code was taken from wallet button wayy before, and we're also doing it here in one of our views https://github.com/reown-com/appkit/blob/dd3a7ddb87eaf295f453c13ccba753897cccfbce/packages/scaffold-ui/src/views/w3m-connecting-social-view/index.ts#L214

packages/controllers/src/controllers/ConnectionController.ts
enesozturk: But there might be cases where user would want to disconnect:
- All connectors
- Specific connector with ID
- Specific Namespace (all connectors belongs to a namespace)

packages/controllers/src/utils/ConnectorControllerUtil.ts
0xmkh: commented below

packages/controllers/exports/react.ts
0xmkh: Put a comment above it as well, it's because we need to trigger re-renders for our controllers

packages/controllers/exports/react.ts
0xmkh: yeah so we can avoid doing new function instance everytime and memoize it, some apps might re-render a lot and can cause an impact

packages/controllers/src/utils/ConnectionControllerUtil.ts
0xmkh: These are local storage values, we store connections in local storage and get them via hooks

packages/controllers/src/utils/ConnectionControllerUtil.ts
0xmkh: We're working on fixing this with the design team to allow users to switch to ordinals and payment address in bitcoin

packages/controllers/src/controllers/ConnectionController.ts
0xmkh: I think it's cleaner to throw an error, later if there's an issue we can debug it way easier. Technically `ConnectionControllerUtil.getConnectionStatus` should always return the right status, but error is thrown just in case

packages/controllers/src/controllers/ConnectorController.ts
```diff
@@ -327,6 +329,16 @@ const controller = {
     return ConnectorController.mergeMultiChainConnectors(namespaceConnectors)
   },
 
+  canSwitchToSmartAccount(namespace: ChainNamespace) {
```
enesozturk: If a method doesn't mutate the controller state, I think it should be a util rather then living inside the controller file

packages/controllers/src/controllers/ConnectionController.ts
0xmkh: Yeah...

I needed this because of one reason. We have "Recently Connected" section in profile view, and we somehow need to know which wallet we shoudl connect to. For example connecting with email should open email login view, connecting with WC should connect to wc view

packages/controllers/src/utils/ChainControllerUtil.ts
```diff
@@ -18,20 +18,36 @@ export function getChainsToDisconnect(namespace?: ChainNamespace) {
   if (namespace) {
     chains.push([namespace, ChainController.state.chains.get(namespace) as ChainAdapter])
 
-    if (checkNamespaceConnectorId(namespace, CommonConstantsUtil.CONNECTOR_ID.WALLET_CONNECT)) {
+    if (
+      ConnectorControllerUtil.checkNamespaceConnectorId(
```
enesozturk, line 22: What does `checkNamespaceConnectorId` checking exactly? To see exist? Can there be a better name for the method that explains what it does?

packages/controllers/src/controllers/ConnectionController.ts
0xmkh: 1. Seems ok if we have two args no ? Since our eslint rule says if it's more than 3, we then introduce params
2. It's auth specific because we have functionality to switch between EOA and SA in profile view

packages/controllers/src/utils/ConnectionControllerUtil.ts
enesozturk: +

packages/controllers/src/controllers/ConnectionController.ts
0xmkh: These are used from wallet buttons, i think i added `closeModalOnConnect` property so we can keep backward compatibility with wallet buttons when connecting via WC and want to close modal right after

packages/controllers/exports/vue.ts
0xmkh: this fn is only responsible for updating reference state in vue so it can be reactive

packages/appkit/src/client/appkit-base-client.ts
0xmkh: i got a eslint error on this after modifying disconnect logic in appkit basic

packages/controllers/src/utils/ChainControllerUtil.ts
0xmkh: The logic for this code hasn't been changed at all. I just put `checkNamespaceConnectorId` to `ConnectorControllerUtil`

0xmkh: not sure what this code does tbh

packages/appkit/src/client/appkit-base-client.ts
```diff
@@ -490,6 +492,8 @@ export abstract class AppKitBaseClient {
         this.setAllAccounts(accounts, chainToUse)
         this.setStatus('connected', chainToUse)
         this.syncConnectedWalletInfo(chainToUse)
+
+        return undefined
```
svenvoskamp, line 497: Why this return undefined here?

packages/controllers/src/utils/ConnectorControllerUtil.ts
```diff
@@ -1,13 +1,354 @@
...
+    onConnect
+  }: ConnectWalletConnectParameters): Promise<ParsedCaipAddress> {
+    return new Promise((resolve, reject) => {
+      if (walletConnect) {
+        ConnectorController.setActiveConnector(connector)
+      }
+
+      onOpen?.(CoreHelperUtil.isMobile() && walletConnect)
+
+      if (redirectViewOnModalClose) {
```
enesozturk, line 68: Why we need `redirectViewOnModalClose` case? Is there a specific case? Can you add a JSDoc about it?

enesozturk: Bc the page redirections are always painful since we have many different use cases there

packages/appkit/src/client/appkit-base-client.ts
```diff
@@ -1637,7 +1653,7 @@ export abstract class AppKitBaseClient {
   }
 
   public async disconnect(chainNamespace?: ChainNamespace) {
```
svenvoskamp, line 1661: Is it not possible to disconnect all chainNamespaces?

packages/appkit/src/client/appkit-base-client.ts
0xmkh: Had an eslint error when i was modifying disconnect part so left it undefined

packages/controllers/src/utils/ConnectorControllerUtil.ts
0xmkh: Yeah if you connect to WC you should get back to profile wallets view

packages/common/src/utils/SafeLocalStorage.ts
svenvoskamp: I think the naming is a bit weird. But we remember these ID's to show them in the UI and give the user the chance to reconnect to them right?

packages/appkit/src/client/appkit-base-client.ts
0xmkh: Not for this feature, you can only disconnect one at a time for each namespace or disconnect all for one namespace

packages/controllers/exports/react.ts
```diff
@@ -1,13 +1,42 @@
...
+  }) => void
+  onError?: (error: Error) => void
+}
+
+interface SwitchConnectionParams {
+  connection: Connection
+  address?: string
+}
+
+interface UseAppKitConnectionDeleteRecentConnectionProps {
```
svenvoskamp: Long confusing name IMO

packages/controllers/exports/react.ts
```diff
@@ -72,9 +101,121 @@ export function useAppKitAccount(options?: { namespace?: ChainNamespace }): UseA
...
+  const formatConnection = useCallback((connection: Connection) => {
+    const connector = ConnectorController.getConnectorById(connection.connectorId)
+
+    const name = ConnectorController.getConnectorName(connector?.name)
+    const icon = AssetUtil.getConnectorImage(connector)
+    const networkImage = AssetUtil.getNetworkImage(connection.caipNetwork)
+
+    return {
+      name,
+      icon,
```
svenvoskamp, line 142: Can we rename icon to connectorIcon?

0xmkh: I think many devs will be used to `icon` since wagmi use it also https://wagmi.sh/dev/creating-connectors#properties

packages/controllers/exports/react.ts
```diff
@@ -72,9 +101,121 @@ export function useAppKitAccount(options?: { namespace?: ChainNamespace }): UseA
...
+}
+
+export function useAppKitConnection({ namespace, onSuccess, onError }: UseAppKitConnectionProps) {
+  const [, forceUpdate] = useState(0)
+
+  const { connections, isSwitchingConnection } = useSnapshot(ConnectionController.state)
+  const { activeConnectorIds } = useSnapshot(ConnectorController.state)
+  const { activeChain } = useSnapshot(ChainController.state)
+
+  const chainNamespace = namespace ?? activeChain
```
svenvoskamp, line 161: I think we should change the name of `activeChain` in `ChainController`. It feels a bit weird to have namings like namespace, chainNamespace, chain and activeChain. 

A lot of these namings are being used across the codebase. We should unify or distinct them

0xmkh: +1

packages/controllers/exports/react.ts
0xmkh: wdyt ?

packages/controllers/exports/vue.ts
```diff
@@ -14,6 +16,50 @@ import type {
...
+}
+
+interface UseAppKitConnectionDeleteRecentConnectionProps {
+  address: string
+  connectorId: string
+}
+
+interface FormattedConnection extends Connection {
+  name: string | undefined
+  icon: string | undefined
```
svenvoskamp, line 54: replace icon with connectorIcon

0xmkh: wdyt about this ? https://github.com/reown-com/appkit/pull/4450#discussion_r2123299650

packages/controllers/src/utils/ConnectorControllerUtil.ts
```diff
@@ -1,13 +1,354 @@
...
+                  EventsController.sendEvent({
+                    type: 'track',
+                    event: 'SOCIAL_LOGIN_REQUEST_USER_DATA',
+                    properties: { provider: socialProvider }
+                  })
+                }
+                await authConnector.provider.connectSocial(uri)
+
+                if (socialProvider) {
+                  StorageUtil.setConnectedSocialProvider(socialProvider)
```
enesozturk, line 155: I think we should make the Storage write after connection is been established - after `connectExternal`, if there is wrong happens on the connection, the storage will show wrong value I guess?

packages/controllers/src/controllers/AccountController.ts
svenvoskamp: How does the relation between a connection and an account work? Do we even need the accountController when we have the connectionController?

packages/controllers/src/utils/ConnectorControllerUtil.ts
enesozturk: Also maybe in the `catch` we should clear the storage for related connector?

0xmkh: There were some edge cases where i was forced to set it before, can't remember on top of my head

0xmkh: maybe we can get social provider directly from secure site ?

packages/controllers/src/controllers/ConnectionController.ts
```diff
@@ -305,39 +365,216 @@ const controller = {
...
+    isSwitchingConnection: ConnectionControllerState['isSwitchingConnection']
+  ) {
+    state.isSwitchingConnection = isSwitchingConnection
+  },
+
+  async disconnect({ id, namespace, disconnectAll }: DisconnectParams = {}) {
     try {
-      await ConnectionController._getClient()?.disconnect(namespace)
+      if (disconnectAll) {
+        await ConnectionController._getClient()?.disconnectAll?.({
```
svenvoskamp: Can we remove the getClient() in connectionController. I think it is redundant as we only use 1 client

packages/controllers/src/utils/ConnectorControllerUtil.ts
```diff
@@ -1,13 +1,354 @@
...
+
+        if (socialProvider === 'farcaster') {
+          onOpenFarcaster?.()
+          const unsubscribeModalController = ModalController.subscribeKey('open', val => {
+            if (!val && social === 'farcaster') {
+              reject(new Error('Popup closed'))
+              onConnect?.()
+              unsubscribeModalController()
+            }
+          })
```
enesozturk, line 221: This and the other `subscribe/subscribeKey` calls in this utility, I'm skeptical how they are working. They are opening a listener to controllers and will be working async in this function block. Wouldn't these things cause race conditions? This smells to me a bit

enesozturk: @tomiir @svenvoskamp wdyt?

0xmkh: Mentioned above that this code was taken from wallet button package, we can change this but wouldn't prefer doing it in this PR since it was just a copy paste

packages/controllers/src/utils/ConnectorControllerUtil.ts
svenvoskamp: @tomiir We do a specific check in the handleSocialConnection to see if the event has a `resultUri` and comes from the `secure_site_sdk_origin`

packages/controllers/src/controllers/ConnectionController.ts
0xmkh: how would we call appkit basic's disconnect method if we don't use `_getClient()` ?

svenvoskamp: I think we can directly call it somehow. But it needs a refactor. This code comes from our old implementation where we would get the specific controllerClient based on the connector library:

old:

connectionController -> wagmi.connectExternal
connectionController -> ethers.connectExternal

new: 

connectionController -> appkit.ConnectExternal > adapter.connectExternal

Every `connectExternal` will always go through AppKit base client, so there is no need to have this `getClient()` anymore. 

Don't know if we should change that in this PR, but something to remember for TDW

packages/controllers/src/utils/StorageUtil.ts
```diff
@@ -589,12 +597,34 @@ export const StorageUtil = {
   },
   setConnections(connections: Connection[], chainNamespace: ChainNamespace) {
     try {
-      const newConnections = {
-        ...StorageUtil.getConnections(),
-        [chainNamespace]: connections
+      const existingConnections = StorageUtil.getConnections()
+      const existing = existingConnections[chainNamespace] ?? []
+
+      const mergedMap = new Map<string, Connection>()
```
enesozturk:
```suggestion
      const connectorConnectionMap = new Map<string, Connection>()
```

packages/controllers/src/utils/ConnectorControllerUtil.ts
0xmkh: also we've added unit, integration and e2e test for this so far works ok

svenvoskamp: Agree with both of you. This is not a good pattern IMO, but let's not change it in this PR. We will do the architecture design changes in TDW

packages/controllers/exports/react.ts
svenvoskamp: Yea makes sense then, it confused me at first, but if this is a known pattern, then it is ok I guess

packages/controllers/src/controllers/ConnectionController.ts
0xmkh: ahh i see, yeah that'd be cleaner approach

packages/common/src/utils/SafeLocalStorage.ts
0xmkh: yeah so we have connections which are recent connections, and disconnected connector ids are used to check if we should reconnect during page refresh if they were already connected

packages/controllers/src/controllers/ConnectionController.ts
```diff
@@ -305,39 +370,210 @@ const controller = {
...
+    const smartAccountAddress = connection.accounts.find(c => c.type === 'smartAccount')?.address
+    const isAddressSmartAccount = smartAccountAddress?.toLowerCase() === address.toLowerCase()
+    const accountType =
+      isAddressSmartAccount && ConnectorControllerUtil.canSwitchToSmartAccount(namespace)
+        ? 'smartAccount'
+        : 'eoa'
 
-    if (isConnectorConnected) {
-      const currentNetwork = ChainController.state.activeCaipNetwork
+    await ConnectionController.setPreferredAccountType(accountType, namespace)
```
0xmkh, line 402: Just remembered that this was needed for "Recently Connected" section in profile wallets view

packages/controllers/src/controllers/ConnectionController.ts
```diff
@@ -123,6 +170,22 @@ const controller = {
     state._client = ref(client)
   },
 
+  initialize(adapters: ChainAdapter[]) {
+    const storageConnections = StorageUtil.getConnections()
+
+    for (const adapter of adapters) {
```
enesozturk, line 176: Will this feature work with adapterless (Basic) version? Can we validate that? It shouldn't be I guess since on Basic there is only WC available, but if yes we should make sure this initialization work fine with Basic too

packages/controllers/src/utils/ConnectionControllerUtil.ts
```diff
@@ -0,0 +1,126 @@
...
+    if (isConnectorConnected) {
+      return 'connected'
+    }
+
+    const isConnectionConnected = connections.some(
+      c => c.connectorId.toLowerCase() === connection.connectorId.toLowerCase()
+    )
+
+    if (isConnectionConnected) {
+      return 'active'
```
enesozturk, line 39: Can we instead of hard coded strings let's have a tuple on some common place?

enesozturk (approved):
Left two non blocking comment 🙏  LFG

packages/controllers/src/controllers/ConnectionController.ts
0xmkh: oh yeah good point, multi-wallet won't be available in basic

packages/controllers/src/utils/ConnectionControllerUtil.ts
0xmkh: Good idea, something like `CommonConstantsUtil.CONNECTION_STATUS.ACTIVE`

PR #4482, opened 2025-06-08
feat: multi-wallet

magiziz:
# Description

### AppKit Client Updates

- Wallet connection flow has been redesigned.  
  - Connecting, switching, and disconnecting is now handled through `connections`, `accountChanged`, and `disconnect` events ONLY!
- Added `syncConnections()`:
  - Syncs all active connections for each adapter.
  - Helps to reconnect all previously connected wallets during page refresh
- `accountChanged()` now accepts a `connector` object, this will be used to remember last switched connector
- WalletConnect event handling has been cleaned up

---

### Controller Changes

- Removed `allAccounts` from `AccountController`. Now using `accounts` from `ConnectionController`

---

### Adapter Updates

Each adapter now manages its own `connections` state and emits updates to the AppKit Client. This improves connection handling across multiple chains.

**Adapters with changes:**
- WagmiAdapter
- EthersAdapter
- Ethers5Adapter
- SolanaAdapter
- BitcoinAdapter

---

### UI Changes

- Removed `Profile` and `SwitchAddress` views.
- Added new `ProfileWallets` view:
  - If you're already connected to wc or auth wallet, it shows a disabled state in connect view.
  - If you try to connect the same extension wallet again from connect view, it asks you to switch accounts.

---

### AppKit Lab

- New "Connections" section in AppKit Lab.
- Lets you view and manage current and recent wallet connections which uses hooks.

---

## Type of change

# Associated Issues

For Linear issues: Closes APKT-3017

# Showcase (Optional)

https://github.com/user-attachments/assets/a08d12bb-7b73-4315-b2ef-4f386cdf9caa

<div align="center">
  
  
</div>

# Checklist


apps/laboratory/package.json
```diff
@@ -51,6 +51,9 @@
     "playwright:test:default-account-types": "playwright test --grep email-default-account-types.spec.ts",
     "playwright:test:core-sign-client": "playwright test --grep sign-client.spec.ts",
     "playwright:test:cloud-auth": "playwright test --grep cloud-auth.spec.ts",
+    "playwright:test:multi-wallet": "HEADLESS=true playwright test --grep multi-wallet.spec.ts",
+    "playwright:test:multi-wallet-siwx": "HEADLESS=true playwright test --grep multi-wallet-siwx.spec.ts",
+    "playwright:test:multi-wallet-multichain": "HEADLESS=true playwright test --grep multi-wallet-multichain.spec.ts",
```
tomiir, line 57: why do we need the node env here?

apps/laboratory/src/components/Connection/ConnectionAccount.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+  const isCurrentActiveAccount = account.address === address
+  const isCurrentActiveConnection = connection.connectorId === activeConnection?.connectorId
+
+  const isSwitching =
+    isLoading &&
+    lastSelectedConnection?.connectorId === connection.connectorId &&
+    lastSelectedConnection?.address === account.address
+
+  const shortenedAddress = `${account.address.slice(0, 6)}...${account.address.slice(-4)}`
```
enesozturk, line 53: If somehow `account.address` is undefined this `slice` will throw error - is `account.address` has value on any case?

packages/adapters/bitcoin/src/adapter.ts
```diff
@@ -79,6 +127,25 @@ export class BitcoinAdapter extends AdapterBlueprint<BitcoinConnector> {
...
+        },
+        {
+          namespace: ConstantsUtil.CHAIN.BITCOIN,
+          publicKey: accounts[1]?.publicKey ?? '',
+          path: accounts[1]?.path ?? '',
+          address: accounts[1]?.address ?? '',
+          type: 'ordinal'
+        }
+      ]
+    }
```
tomiir: Are these accoutns guaranteed to be in this order always? 
If so, I'd add some constants for the index like `PAYMENT_ACCOUNT = 0` etc

packages/adapters/bitcoin/src/adapter.ts
```diff
@@ -298,45 +478,194 @@ export class BitcoinAdapter extends AdapterBlueprint<BitcoinConnector> {
...
-    this.unbindEvents()
+  private providerHandlers: Record<
+    string,
+    {
+      disconnect: () => void
+      accountsChanged: (accounts: string[]) => void
+      chainChanged: (chainId: string) => void
+      connector: BitcoinConnector
+    } | null
+  > = {}
```
tomiir: why do you need this?

packages/adapters/bitcoin/src/connectors/WalletStandardConnector.ts
```diff
@@ -197,38 +192,6 @@ export class WalletStandardConnector extends ProviderEventEmitter implements Bit
...
-            if (data.accounts.length === 0) {
-              this.emit('disconnect')
-            } else {
-              this.emit(
-                'accountsChanged',
-                data.accounts.map(acc => acc.address)
-              )
-            }
-          }
-        })
```
tomiir, line 218: is it ok to remove this?

packages/adapters/bitcoin/src/adapter.ts
```diff
@@ -130,6 +197,83 @@ export class BitcoinAdapter extends AdapterBlueprint<BitcoinConnector> {
...
+        .filter(c => {
+          const { hasDisconnected, hasConnected } = getConnectorStorageInfo(c.id)
+
+          return !hasDisconnected && hasConnected
+        })
+        .map(async connector => {
+          if (connector.id === CommonConstantsUtil.CONNECTOR_ID.WALLET_CONNECT) {
+            const accounts = WcHelpersUtil.getWalletConnectAccounts(
+              this.universalProvider as UniversalProvider,
+              'bip122'
```
enesozturk:
```suggestion
              ConstantUtil.CHAIN.BITCOIN
```

packages/adapters/bitcoin/src/adapter.ts
```diff
@@ -130,6 +197,83 @@ export class BitcoinAdapter extends AdapterBlueprint<BitcoinConnector> {
     })
   }
 
+  public async syncConnections({
+    connectToFirstConnector,
+    caipNetwork,
+    getConnectorStorageInfo
+  }: AdapterBlueprint.SyncConnectionsParams) {
+    await Promise.allSettled(
```
enesozturk: From my experience while searching for the issue on connection syncing logics - these `Promise.allSetteled` and `Promise.all` are critical to make sure we are handling error management right, otherwise they might be really hard to identify the issue. 

We should have proper convention on how to handle catches in one of the iterated async functions. Better to be unit tested too.

How does the flow affected if an of these failed?

Same for the `disconnectAll` method below.

packages/adapters/bitcoin/src/adapter.ts
```diff
@@ -41,17 +48,58 @@ export class BitcoinAdapter extends AdapterBlueprint<BitcoinConnector> {
...
-
-    const address = await connector.connect()
-
     const chain = connector.chains.find(c => c.id === params.chainId) || connector.chains[0]
 
     if (!chain) {
       throw new Error('The connector does not support any of the requested chains')
     }
 
+    const connection = this.connections.find(c => c.connectorId === connector.id)
```
enesozturk: Pretty much implementation went into the adapter clients - this feels like went away from our approach abstracting things with the AdapterBlueprint and keep the clients simple as possible. Like to ask if it's possible to simplify these logics in Blueprint and not repeat on each adapter? From the point of the review it's a bit hard to see so like to discuss.

Also ping for @tomiir @svenvoskamp

packages/adapters/bitcoin/src/connectors/WalletStandardConnector.ts
svenvoskamp, line 1: What is the reason to remove the bindevents?

packages/adapters/ethers/src/client.ts
svenvoskamp, line 1: All clients got significantly bigger, also see some calls with `addConnection` etc. Is this for updating AppKit state? Because it goes against our design to update state in Adapters

packages/adapters/bitcoin/src/adapter.ts
0xmkh: yeah it's supposed to be in that order, we do it in the account screen as well 

https://github.com/reown-com/appkit/blob/main/packages/scaffold-ui/src/partials/w3m-account-default-widget/index.ts#L388

packages/adapters/bitcoin/src/connectors/WalletStandardConnector.ts
0xmkh: yeah since we have `accountsChanged` in bitcoin adapter that takes care of it

packages/adapters/bitcoin/src/adapter.ts
0xmkh: Right i agree, made it to be `Promise.all` instead

packages/adapters/bitcoin/src/adapter.ts
0xmkh: Created a `Connection` class to handle all connection logic 🙏
This should hopefully make the adapter code much cleaner and easier to maintain

packages/adapters/bitcoin/src/connectors/WalletStandardConnector.ts
0xmkh: Noticed we already had an `accountChanged` event handling this in bitcoin adapter, since i just wanted to keep the events and connections code simpler

packages/adapters/solana/src/client.ts
```diff
@@ -319,52 +371,176 @@ export class SolanaAdapter extends AdapterBlueprint<SolanaProvider> {
...
+
+      this.providerHandlers[connectorId] = null
     }
   }
 
   public override setUniversalProvider(universalProvider: UniversalProvider): void {
+    this.universalProvider = universalProvider
+    const wcConnectorId = CommonConstantsUtil.CONNECTOR_ID.WALLET_CONNECT
+
+    WcHelpersUtil.listenWcProvider({
```
svenvoskamp, line 417: Do we need this `listenWcProvider` in every client? Can't we unify this or do this in the AppKit Client?

apps/laboratory/tests/multi-wallet-multichain.spec.ts
```diff
@@ -0,0 +1,197 @@
...
+})
+
+// -- Tests --------------------------------------------------------------------
+test('should enable multi-wallet feature with a whitelisted project id', async ({ library }) => {
+  if (library === 'bitcoin') {
+    return
+  }
+
+  await page.getByTestId('project-id-button').click()
+  await page.getByTestId('project-id-input').fill('1b0841d0acfe3e32dcb0d53dbf505bdd')
```
enesozturk, line 56: Is this project ID limited to lab (for safety)?

apps/laboratory/tests/multi-wallet.spec.ts
```diff
@@ -0,0 +1,156 @@
+/* eslint-disable no-await-in-loop */
```
enesozturk, line 1: Maybe name of this file could be `multi-wallet-email.spec.ts` since it's using auth provider

packages/appkit/src/client/appkit-base-client.ts
```diff
@@ -164,6 +165,7 @@ export abstract class AppKitBaseClient {
     await this.initChainAdapters()
     this.sendInitializeEvent(options)
     await this.syncExistingConnection()
+    await this.syncAdapterConnections()
```
svenvoskamp, line 168: What is the difference between `syncExistingConnection` and `syncAdapterConnections`. Can we remove `syncExistingConnection`?

packages/appkit/src/connections/Connection.ts
```diff
@@ -0,0 +1,321 @@
...
+   * Otherwise, returns the first available valid connection.
+   *
+   * @param params - Connection parameters
+   * @param params.address - Optional address to filter by
+   * @param params.connectorId - Optional connector ID to filter by
+   * @param params.connections - List of available connections
+   * @param params.connectors - List of available connectors
+   * @returns Connection or null if none found
+   */
+  getConnection({ address, connectorId, connections, connectors }: GetConnectionParams) {
```
enesozturk: Cannot this function get connections (and maybe other props) from controllers? Do we really have to pass connections, connectors etc to it every call?

packages/appkit/src/client/appkit-base-client.ts
```diff
@@ -1356,47 +1405,73 @@ export abstract class AppKitBaseClient {
...
-
-      this.universalProvider.on('chainChanged', (chainId: number | string) => {
-        // eslint-disable-next-line eqeqeq
-        const caipNetwork = this.getCaipNetworks().find(c => c.id == chainId)
-        const currentCaipNetwork = this.getCaipNetwork()
-
-        if (!caipNetwork) {
-          this.setUnsupportedNetwork(chainId)
+      this.chainNamespaces.forEach(namespace => {
+        WcHelpersUtil.listenWcProvider({
```
svenvoskamp, line 1402: Can we remove as it is in the adapters?

packages/appkit/src/connections/Connection.ts
enesozturk: Same for the other props, if possible, it'd be simplify more to make these functions easier to use; get what you want using data from controllers

packages/appkit/src/universal-adapter/client.ts
```diff
@@ -41,9 +41,16 @@ export class UniversalAdapter extends AdapterBlueprint {
...
+      this.emit('disconnect')
     } catch (error) {
       console.warn('UniversalAdapter:disconnect - error', error)
     }
+
+    return { connections: [] }
+  }
+
+  public override syncConnections() {
+    return Promise.resolve()
```
enesozturk, line 55: Why is this just resolving without doing anything? Do we override it on all adapters?

packages/appkit/tests/client/disconnect.test.ts
```diff
@@ -689,11 +578,11 @@ describe('AppKit - disconnect - error handling scenarios', () => {
     expect(setLoadingSpy).toHaveBeenCalledWith(false, solanaNamespace) // Solana end (from catch)
 
     // --- Assertions for global cleanup (called after loop, before final throw if failures exist) ---
-    expect(ccResetWcConnectionSpy).toHaveBeenCalledOnce()
-    expect(siwxClearSessionsSpy).toHaveBeenCalledOnce()
+    expect(ccResetWcConnectionSpy).toHaveBeenCalledTimes(2)
```
enesozturk, line 581: Why this function getting called twice? Is the second call required or redundant? If so we should make sure it's called once

apps/laboratory/tests/multi-wallet.spec.ts
0xmkh: mhmm it's also using extension and wc, so calling it `multi-wallet-email.spec.ts` would be little confusing wdyt ?

packages/controllers/src/utils/ConnectionControllerUtil.ts
```diff
@@ -118,6 +83,15 @@ export const ConnectionControllerUtil = {
       recentConnectionsWithCurrentActiveConnectors
     )
 
+    if (!isMultiWalletEnabled) {
+      return {
+        connections: connections.filter(
+          c => c.connectorId.toLowerCase() === activeConnectorId?.toLowerCase()
```
enesozturk, line 89: What are the cases where connector id's would be same but requires to be lower cased for comparision? I mean do we have connector id's where we use upper case at some places and using lower case at some places? We should be safe to not use `toLowerCase()`, if you find any case better to fix the root cause? 

Or maybe it's breaking backwards compatibility?

packages/scaffold-ui/src/partials/w3m-account-default-widget/index.ts
```diff
@@ -437,8 +337,19 @@ export class W3mAccountDefaultWidget extends LitElement {
...
+        ? (ConnectionController.state.connections.get(this.namespace) ?? [])
+        : []
+      const hasConnections = connectionsByNamespace.length > 0
+      const connectorId =
+        this.namespace && ConnectorController.state.activeConnectorIds[this.namespace]
+      const isMultiWalletEnabled = this.remoteFeatures?.multiWallet
+      await ConnectionController.disconnect(isMultiWalletEnabled ? { id: connectorId } : {})
+      if (hasConnections && isMultiWalletEnabled) {
+        RouterController.reset('Account')
+        RouterController.push('ProfileWallets')
```
enesozturk: Could be
```suggestion
        RouterController.replace('ProfileWallets')
```
?

packages/appkit/src/client/appkit-base-client.ts
0xmkh: I was thinking of removing it, but later in the process it made sense to keep it:
- `syncExistingConnection` is responsible for syncing the last connected connector.
- `syncAdapterConnections` handles syncing all previously connected connectors.

I didn't want to make too many changes to the AppKit base client, so I went with this approach for now.

packages/scaffold-ui/src/partials/w3m-account-wallet-features-widget/index.ts
```diff
@@ -354,10 +354,32 @@ export class W3mAccountWalletFeaturesWidget extends LitElement {
     }
   }
 
+  private getAuthData() {
+    const socialProvider = StorageUtil.getConnectedSocialProvider() as SocialProvider | null
```
enesozturk, line 358: Fetching these kind of values from local storage feels wrong to me, we should get them from controllers where we have ability to listen changes, local storage should be read/write only when data is updated, not to show in UI imo

packages/appkit/src/connections/Connection.ts
0xmkh: ah yeah good idea

packages/appkit/src/client/appkit-base-client.ts
0xmkh: i kept it there in case users don't use adapters

packages/appkit/src/universal-adapter/client.ts
0xmkh: universal adapter doesn't handle multiple connections as it's only responsible for connecting to wc and that's it

packages/appkit/tests/client/disconnect.test.ts
0xmkh: ah yeah in the test i call disconnect twice so it's expected that it's called twice
```ts
await (appKit as any).connectionControllerClient.disconnect({ chainNamespace: eip155Namespace })
await (appKit as any).connectionControllerClient.disconnect({ chainNamespace: solanaNamespace })
```

packages/controllers/src/utils/ConnectionControllerUtil.ts
0xmkh: I mean yeah i could remove it i actually used it everywhere including the views and in adapters as well. I honestly just kept it there for safety reason

packages/scaffold-ui/src/partials/w3m-account-wallet-features-widget/index.ts
0xmkh: yeah i agree, we currently don't have a controller state that takes care of that and we use the same approach in all other views / utils.

I can create a followup PR that doesn't read the value from local storage if fine ?

packages/appkit/src/universal-adapter/client.ts
0xmkh: if we support auth or extension wallets in future for universal adapter then it'd make sense to add something there

packages/appkit/src/connections/Connection.ts
0xmkh: oh wait noticed an edge case, during initialization of appkit the connections and connector controllers sometimes weren't fully synced, so when i call `getConnection` the controller states is not up to date

packages/scaffold-ui/src/partials/w3m-social-login-widget/index.ts
```diff
@@ -241,6 +247,12 @@ export class W3mSocialLoginWidget extends LitElement {
       }
     }
   }
+
+  private hasConnection() {
```
enesozturk, line 246: Seems like this logic has been used in different views, maybe it could be a util function where it gets connector ID as param?

packages/scaffold-ui/src/views/w3m-account-settings-view/index.ts
```diff
@@ -310,8 +315,17 @@ export class W3mAccountSettingsView extends LitElement {
...
-      ModalController.close()
+      const namespace = this.network?.chainNamespace as ChainNamespace
+      const connectionsByNamespace = ConnectionController.state.connections.get(namespace) ?? []
+      const hasConnections = connectionsByNamespace.length > 0
+      const connectorId = namespace && ConnectorController.state.activeConnectorIds[namespace]
+      const isMultiWalletEnabled = this.remoteFeatures?.multiWallet
+      await ConnectionController.disconnect(isMultiWalletEnabled ? { id: connectorId } : {})
+      if (hasConnections && isMultiWalletEnabled) {
+        RouterController.reset('Account')
+        RouterController.push('ProfileWallets')
```
enesozturk: Same here, could it be?
```suggestion
        RouterController.replace('ProfileWallets')
```

packages/scaffold-ui/src/views/w3m-connecting-external-view/index.ts
```diff
@@ -1,26 +1,50 @@
...
 @customElement('w3m-connecting-external-view')
 export class W3mConnectingExternalView extends W3mConnectingWidget {
   // -- Members ------------------------------------------- //
   private externalViewUnsubscribe: (() => void)[] = []
+  private connections = this.connector
+    ? (ConnectionController.state.connections.get(this.connector?.chain) ?? [])
+    : []
+  private connectionsByNamespace = this.connector
+    ? (ConnectionController.state.connections.get(this.connector?.chain) ?? [])
+    : []
```
enesozturk: Seems these two points to exact same value?

packages/scaffold-ui/src/views/w3m-connecting-external-view/index.ts
```diff
@@ -74,6 +113,57 @@ export class W3mConnectingExternalView extends W3mConnectingWidget {
...
+
+        const newAccounts = ConnectionControllerUtil.getConnectionsByConnectorId(
+          newConnections,
+          this.connector.id
+        ).flatMap(c => c.accounts)
+
+        if (newAccounts.length === 0) {
+          if (this.hasMultipleConnections && isMultiWalletEnabled) {
+            RouterController.replace('Account')
+            RouterController.push('ProfileWallets')
```
enesozturk: Same here

packages/adapters/bitcoin/src/adapter.ts
```diff
@@ -130,6 +198,48 @@ export class BitcoinAdapter extends AdapterBlueprint<BitcoinConnector> {
     })
   }
 
+  public async syncConnections({
+    connectToFirstConnector,
+    caipNetwork,
+    getConnectorStorageInfo
+  }: AdapterBlueprint.SyncConnectionsParams) {
+    await this.connection.syncConnections({
```
tomiir: semantics question: how come a connection has `syncConnections` ?

packages/scaffold-ui/src/views/w3m-connecting-farcaster-view/index.ts
```diff
@@ -211,7 +221,13 @@ export class W3mConnectingFarcasterView extends LitElement {
           })
         }
         this.loading = false
-        ModalController.close()
+        if (hasConnections && isMultiWalletEnabled) {
+          RouterController.reset('Account')
+          RouterController.push('ProfileWallets')
```
enesozturk: Same here - you can ignore these comments if it's not the same solution

packages/adapters/bitcoin/src/adapter.ts
tomiir: maybe it should be `ConnectionManager` instead of `Connection` ?

packages/scaffold-ui/src/views/w3m-connecting-social-view/index.ts
```diff
@@ -66,8 +76,19 @@ export class W3mConnectingSocialView extends LitElement {
...
+        OptionsController.subscribeKey('remoteFeatures', val => {
+          this.remoteFeatures = val
+        }),
+        AccountController.subscribeKey('address', val => {
+          const isMultiWalletEnabled = this.remoteFeatures?.multiWallet
+
+          if (val && val !== this.address) {
+            if (this.hasMultipleConnections && isMultiWalletEnabled) {
+              RouterController.reset('Account')
+              RouterController.push('ProfileWallets')
```
enesozturk: +

packages/scaffold-ui/src/views/w3m-connecting-wc-view/index.ts
```diff
@@ -91,9 +92,21 @@ export class W3mConnectingWcView extends LitElement {
...
+            ChainController.state.activeChain as ChainNamespace
+          ) ?? []
+        const isMultiWalletEnabled = this.remoteFeatures?.multiWallet
+        const hasConnections = connectionsByNamespace.length > 0
         await ConnectionController.connectWalletConnect()
         if (!this.isSiwxEnabled) {
-          ModalController.close()
+          if (hasConnections && isMultiWalletEnabled) {
+            RouterController.reset('Account')
+            RouterController.push('ProfileWallets')
```
enesozturk: ++

packages/scaffold-ui/src/views/w3m-connecting-wc-view/index.ts
```diff
@@ -91,9 +92,21 @@ export class W3mConnectingWcView extends LitElement {
         CoreHelperUtil.isPairingExpired(wcPairingExpiry) ||
         status === 'connecting'
       ) {
+        const connectionsByNamespace =
+          ConnectionController.state.connections.get(
+            ChainController.state.activeChain as ChainNamespace
+          ) ?? []
```
enesozturk: There are many business logics like this to get certain values; connections, connections by namespace, connection by connector id etc. Imo better to create a util which has getter functions and use in all views and other places

packages/scaffold-ui/src/views/w3m-email-verify-otp-view/index.ts
```diff
@@ -20,6 +23,12 @@ export class W3mEmailVerifyOtpView extends W3mEmailOtpWidget {
         await this.authConnector.provider.connectOtp({ otp })
         EventsController.sendEvent({ type: 'track', event: 'EMAIL_VERIFICATION_CODE_PASS' })
 
+        const connectionsByNamespace =
+          ConnectionController.state.connections.get(
+            ChainController.state.activeChain as ChainNamespace
+          ) ?? []
```
enesozturk: Same here, repeating

packages/adapters/bitcoin/src/adapter.ts
```diff
@@ -298,45 +430,220 @@ export class BitcoinAdapter extends AdapterBlueprint<BitcoinConnector> {
...
+      connector.on('accountsChanged', accountsChanged)
+      connector.on('chainChanged', chainChanged)
+      connector.on('disconnect', disconnect)
+
+      this.providerHandlers[connector.id] = {
+        connector,
+        disconnect,
+        accountsChanged,
+        chainChanged
+      }
```
tomiir: is there any way we can abstract this somehow? this function has like 200 lines and defines all of the callbacks inside of it

packages/adapters/bitcoin/src/adapter.ts
```diff
@@ -298,45 +430,220 @@ export class BitcoinAdapter extends AdapterBlueprint<BitcoinConnector> {
...
+          )
+        ) {
+          this.emitFirstAvailableConnection()
+        }
+
+        if (this.connections.length === 0) {
+          this.emit('disconnect')
+        }
+      },
+      onAccountsChanged: accounts => {
```
tomiir: are all of these inline definitions needed? seems like quite the generic definitions, can they be abstracted?

packages/scaffold-ui/src/partials/w3m-social-login-widget/index.ts
0xmkh: good idea

packages/appkit/src/connections/Connection.ts
enesozturk: I got it

packages/adapters/ethers/src/client.ts
```diff
@@ -295,12 +301,156 @@ export class EthersAdapter extends AdapterBlueprint {
     })
   }
 
+  private async disconnectAll() {
+    const connections = await Promise.all(
+      this.connections.map(async connection => {
+        const connector = this.connectors.find(c =>
+          HelpersUtil.isLowerCaseMatch(c.id, connection.connectorId)
+        )
```
tomiir, line 304: nit improvement : `connections` could instead be an object { connectorId: connection } and this would be 0(1) instead of 0(n^2)

packages/adapters/ethers/src/client.ts
```diff
@@ -295,12 +301,156 @@ export class EthersAdapter extends AdapterBlueprint {
...
+          )
+        ) {
+          this.emitFirstAvailableConnection()
+        }
+
+        if (this.connections.length === 0) {
+          this.emit('disconnect')
+        }
+      },
+      onAccountsChanged: accounts => {
```
tomiir: this does indeed seem generic / duplicated (see my comment on btc client) we should probably abstract it

packages/adapters/ethers/src/client.ts
```diff
@@ -356,17 +543,34 @@ export class EthersAdapter extends AdapterBlueprint {
...
-      accounts = [address]
+      const caipNetwork = this.getCaipNetworks().find(n => n.id.toString() === chainId?.toString())
+
+      accounts = [_address]
+
+      this.addConnection({
+        connectorId: id,
+        accounts: authAccounts
+          ? authAccounts.map(account => ({ address: account.address }))
+          : [{ address: _address }],
```
tomiir: how come `[_address]` ,  `authAccounts.map(account => ({ address: account.address }))` and `[{ address: _address }]` are all different?

packages/adapters/ethers/src/client.ts
```diff
@@ -566,26 +883,58 @@ export class EthersAdapter extends AdapterBlueprint {
...
-    this.providerHandlers = {
-      disconnect,
-      accountsChanged: accountsChangedHandler,
-      chainChanged: chainChangedHandler
+      this.providerHandlers[connectorId] = {
+        provider,
+        disconnect,
+        accountsChanged: accountsChangedHandler,
+        chainChanged: chainChangedHandler
+      }
```
tomiir: same as before, this looks really generic and abstractable, also huge 200 line inline definition

packages/appkit/src/connections/Connection.ts
enesozturk: I got it - then maybe when we initialize the Connection instance we could pass reference of connections() getter of ChainAdapterBlueprint and use that reference to get original up to date list? Would that make sense?

packages/adapters/ethers5/src/client.ts
```diff
@@ -298,12 +304,156 @@ export class Ethers5Adapter extends AdapterBlueprint {
...
+  }
+
   public override setUniversalProvider(universalProvider: UniversalProvider): void {
+    this.universalProvider = universalProvider
+    const wcConnectorId = CommonConstantsUtil.CONNECTOR_ID.WALLET_CONNECT
+
+    WcHelpersUtil.listenWcProvider({
+      universalProvider,
+      namespace: CommonConstantsUtil.CHAIN.EVM,
+      onConnect: accounts => {
```
tomiir: same as before

apps/laboratory/tests/multi-wallet-multichain.spec.ts
0xmkh: yeah only to lab

packages/adapters/ethers5/src/client.ts
```diff
@@ -343,12 +493,50 @@ export class Ethers5Adapter extends AdapterBlueprint {
...
+        })
+
+        return {
+          address: connection.account.address,
+          chainId: caipNetwork.id,
+          provider: connector.provider,
+          type: connector.type,
+          id
+        }
+      }
```
tomiir, line 443: how can the connection already exist when calling connect?

packages/adapters/ethers5/src/client.ts
```diff
@@ -358,15 +546,34 @@ export class Ethers5Adapter extends AdapterBlueprint {
     let requestChainId: string | undefined = undefined
 
     if (type === 'AUTH') {
-      const { address } = await (selectedProvider as unknown as W3mFrameProvider).connect({
+      const { address: _address, accounts: authAccounts } = await (
+        selectedProvider as unknown as W3mFrameProvider
```
tomiir, line 458: is the unknown needed here?

packages/adapters/ethers5/src/client.ts
```diff
@@ -358,15 +546,34 @@ export class Ethers5Adapter extends AdapterBlueprint {
...
+
+      this.addConnection({
+        connectorId: id,
+        accounts: authAccounts
+          ? authAccounts.map(account => ({ address: account.address }))
+          : [{ address: _address }],
+        caipNetwork,
+        auth: {
+          name: StorageUtil.getConnectedSocialProvider(),
+          username: StorageUtil.getConnectedSocialUsername()
```
tomiir, line 477: we shouldn't be using storage here if possible

packages/adapters/ethers5/src/client.ts
```diff
@@ -544,22 +802,85 @@ export class Ethers5Adapter extends AdapterBlueprint {
     return { balance: '0.00', symbol: 'ETH' }
   }
 
-  private providerHandlers: {
-    disconnect: () => void
-    accountsChanged: (accounts: string[]) => void
-    chainChanged: (chainId: string) => void
-  } | null = null
+  private emitFirstAvailableConnection() {
```
tomiir: can't this be used in `connect` ?

packages/adapters/ethers5/src/client.ts
```diff
@@ -544,22 +802,85 @@ export class Ethers5Adapter extends AdapterBlueprint {
...
+      this.deleteConnection(connectorId)
+
+      if (
+        HelpersUtil.isLowerCaseMatch(
+          this.getConnectorId(CommonConstantsUtil.CHAIN.EVM),
+          connectorId
+        )
+      ) {
+        this.emitFirstAvailableConnection()
+      }
```
tomiir: why are you emmitting here?

packages/adapters/ethers5/src/client.ts
```diff
@@ -544,22 +802,85 @@ export class Ethers5Adapter extends AdapterBlueprint {
...
+    string,
+    {
+      disconnect: () => void
+      accountsChanged: (accounts: string[]) => void
+      chainChanged: (chainId: string) => void
+      provider: Provider | CombinedProvider
+    } | null
+  > = {}
+
+  private listenProviderEvents(connectorId: string, provider: Provider | CombinedProvider) {
```
tomiir: same as before here

packages/adapters/solana/src/client.ts
```diff
@@ -319,52 +371,176 @@ export class SolanaAdapter extends AdapterBlueprint<SolanaProvider> {
...
+  private providerHandlers: Record<
+    string,
+    {
+      disconnect: () => void
+      accountsChanged: (publicKey: PublicKey) => void
+      provider: SolanaProvider
+    } | null
+  > = {}
+
+  private listenProviderEvents(connectorId: string, provider: SolanaProvider) {
```
tomiir: same

packages/appkit/src/client/appkit-base-client.ts
tomiir: I'd rename appropiately

packages/appkit/src/client/appkit-base-client.ts
```diff
@@ -456,25 +453,45 @@ export abstract class AppKitBaseClient {
         const activeChain = ChainController.state.activeChain
         const adapter = this.getAdapter(activeChain)
         const chainId = this.getCaipNetwork(activeChain)?.id
+        const connections = activeChain
+          ? (ConnectionController.state.connections.get(activeChain) ?? [])
+          : []
```
tomiir: isnt the ?: redundant?

packages/appkit/src/connections/Connection.ts
```diff
@@ -0,0 +1,321 @@
...
+  public namespace: ChainNamespace
+
+  constructor(params: { namespace: ChainNamespace }) {
+    this.namespace = params.namespace
+  }
+
+  public async syncConnections(
+    params: SyncEvmConnections | SyncSolanaConnections | SyncBitcoinConnections
+  ) {
+    switch (this.namespace) {
```
tomiir: do we ever want to update othe namespaces connections ?

packages/appkit/src/connections/Connection.ts
```diff
@@ -0,0 +1,321 @@
...
+            })
+
+            if (address) {
+              onConnection({
+                connectorId: connector.id,
+                accounts: [{ address }],
+                caipNetwork
+              })
+
+              onListenProvider(connector.id, connector.provider as SolanaProvider)
```
tomiir: why do you. check the connector type for EVM and not do this dfor auth and wc but do it here?

packages/appkit/src/connections/Connection.ts
```diff
@@ -0,0 +1,321 @@
...
+          }
+
+          const chain = connector.chains.find(c => c.id === caipNetwork?.id) || connector.chains[0]
+
+          if (!chain) {
+            throw new Error('The connector does not support any of the requested chains')
+          }
+
+          if (address) {
+            onListenProvider(connector.id, connector.provider as BitcoinConnector)
```
tomiir: same

packages/adapters/bitcoin/src/adapter.ts
0xmkh: yeah i think `ConnectionManager` makes more sense here

packages/adapters/ethers/src/client.ts
0xmkh: good idea, would it make sense to refactor it in another PR ? I guess it'd be a huge change if i do it now

packages/adapters/ethers/src/client.ts
0xmkh: ah good catch, simplified it a bit 🙏

packages/adapters/ethers5/src/client.ts
0xmkh: cause if you connect first time we do `addConnection(...)` and it's already there on the adapter level

packages/adapters/ethers5/src/client.ts
0xmkh: it does throw type error, it was like this before as well

packages/adapters/ethers5/src/client.ts
0xmkh: agree, left a comment here as well https://github.com/reown-com/appkit/pull/4482#discussion_r2151987040

packages/adapters/ethers5/src/client.ts
0xmkh: this is actually used everywhere in connect, disconnect and in connector events

packages/adapters/ethers5/src/client.ts
0xmkh: so i'm checking if the current connector gets disconnected and if it does we need to immediately find the first available connection and emit to appkit client

packages/appkit/src/client/appkit-base-client.ts
0xmkh: pushed a change

packages/appkit/src/connections/Connection.ts
0xmkh: not atm, i think this is enough

packages/appkit/src/connections/Connection.ts
0xmkh: ah cause these are already handled on the adapter / appkit client level

packages/adapters/wagmi/src/client.ts
```diff
@@ -296,6 +329,36 @@ export class WagmiAdapter extends AdapterBlueprint {
...
+    chainId,
+    connector
+  }: {
+    address: string
+    chainId: number | string
+    connector: Connector
+  }) {
+    const provider = (await connector.getProvider().catch(() => undefined)) as Provider | undefined
+
+    this.emit('accountChanged', {
```
enesozturk, line 343: Should we still emit this event if the provider is undefined?

packages/adapters/wagmi/src/client.ts
```diff
@@ -225,20 +227,51 @@ export class WagmiAdapter extends AdapterBlueprint {
...
+
+            return {
+              accounts: connection.accounts.map(account => ({
+                address: account
+              })),
+              caipNetwork,
+              connectorId: connection.connector.id,
+              auth: isAuth
+                ? {
+                    name: StorageUtil.getConnectedSocialProvider(),
```
enesozturk, line 266: Same here, if using these values from Controllers problematic, we should fixe that first place - iirc there was an syncing issue with these?

packages/scaffold-ui/src/views/w3m-connecting-wc-view/index.ts
0xmkh: yeah pushed a getter function for this 🙏

packages/adapters/wagmi/src/client.ts
0xmkh: yeah i can create a followup PR where we use controllers, but we use these everywhere because we don't have controller state for socials

packages/adapters/wagmi/src/client.ts
0xmkh: good catch, we shouldn't

apps/laboratory/src/components/Connection/ConnectionAccount.tsx
0xmkh: `account.address` can't be undefined

packages/appkit/src/adapters/ChainAdapterBlueprint.ts
```diff
@@ -84,6 +102,12 @@ export abstract class AdapterBlueprint<
     this.projectId = params.projectId
     this.namespace = params.namespace
     this.adapterType = params.adapterType
+
+    if (this.namespace) {
+      this.connectionManager = new ConnectionManager({
+        namespace: this.namespace
+      })
+    }
```
tomiir: why here instead of constructore?

packages/appkit/src/adapters/ChainAdapterBlueprint.ts
```diff
@@ -455,6 +479,224 @@ export abstract class AdapterBlueprint<
...
+  /**
+   * Handles connect event for a specific connector.
+   * @param {string[]} accounts - The accounts that changed
+   * @param {string} connectorId - The ID of the connector
+   */
+  protected onConnect(accounts: (ParsedCaipAddress | string)[], connectorId: string) {
+    if (accounts.length > 0) {
+      const account = accounts[0]
+      const address = typeof account === 'string' ? account : account?.address
+      const chainId = typeof account === 'string' ? undefined : account?.chainId
```
tomiir: why do you need these parses ?

packages/appkit/src/adapters/ChainAdapterBlueprint.ts
```diff
@@ -455,6 +479,224 @@ export abstract class AdapterBlueprint<
...
+        })
+
+        this.addConnection({
+          connectorId,
+          accounts: accounts.map(_account => {
+            if (typeof _account === 'string') {
+              return { address: _account }
+            }
+
+            return { address: _account.address }
```
tomiir: same. 

can we make it a util? maybe it's responsibility of the specific namespace adapter to parse it to a bare addresS?

packages/appkit/src/universal-adapter/client.ts
```diff
@@ -179,6 +179,10 @@ export class UniversalAdapter extends AdapterBlueprint {
     })
   }
 
+  public override emitFirstAvailableConnection(): void {
+    return undefined
+  }
```
tomiir, line 186: is this ok?

packages/appkit/src/adapters/ChainAdapterBlueprint.ts
0xmkh: ah since `account` can either be a string or `ParsedCaipAddress`

0xmkh: in `WcHelperUtil.listenProvider` we parse as `ParsedCaipAddress`

packages/appkit/src/universal-adapter/client.ts
0xmkh: yep, we don't manage connections in universal adapter

packages/appkit/src/universal-adapter/client.ts
```diff
@@ -41,9 +41,16 @@ export class UniversalAdapter extends AdapterBlueprint {
     try {
       const connector = this.getWalletConnectConnector()
       await connector.disconnect()
+      this.emit('disconnect')
```
svenvoskamp, line 46: Could this result in a loop where it keeps disconnecting?

0xmkh: mhmm can't think of a situation that happening, it's basically disconnecting and emits disconnect event and that's pretty much it

packages/appkit/src/adapters/ChainAdapterBlueprint.ts
0xmkh: put it under `constructor`

packages/appkit/src/adapters/ChainAdapterBlueprint.ts
```diff
@@ -90,6 +91,12 @@ export abstract class AdapterBlueprint<
       ConnectorController.getConnectorId(namespace)
 
     if (params) {
+      if (params.namespace) {
```
tomiir:
```suggestion
    if (params?.namespace) {
```

tomiir: careful with committing from here, indentation is not correct

packages/appkit/src/adapters/ChainAdapterBlueprint.ts
```diff
@@ -488,8 +489,8 @@ export abstract class AdapterBlueprint<
   protected onConnect(accounts: (ParsedCaipAddress | string)[], connectorId: string) {
     if (accounts.length > 0) {
       const account = accounts[0]
-      const address = typeof account === 'string' ? account : account?.address
-      const chainId = typeof account === 'string' ? undefined : account?.chainId
+      const address = CoreHelperUtil.isString(account) ? account : account?.address
+      const chainId = CoreHelperUtil.isString(account) ? undefined : account?.chainId
```
tomiir: I actually meant something like 
`getAddress(string | { address}): string`

packages/scaffold-ui/src/views/w3m-profile-wallets-view/index.ts
```diff
@@ -165,6 +167,30 @@ export class W3mProfileWalletsView extends LitElement {
...
+          const tabWidth = availableWidth / tabCount
+          this.tabWidth = `${tabWidth}px`
+          this.requestUpdate()
+        }
+      }
+
+      this.tabsResizeObserver = new ResizeObserver(handleTabsResize)
+      this.tabsResizeObserver.observe(this)
+      handleTabsResize()
+    }
```
tomiir, line 193: why add this ?

packages/adapters/bitcoin/src/adapter.ts
```diff
@@ -79,6 +124,25 @@ export class BitcoinAdapter extends AdapterBlueprint<BitcoinConnector> {
       )
     )
 
+    if (accounts && accounts.length > 1) {
```
enesozturk, line 127: Why do we expect accounts to be more than 1? Wouldn't single account be enough? I mean some wallets might return signle account maybe?

packages/adapters/ethers/src/client.ts
```diff
@@ -456,44 +589,73 @@ export class EthersAdapter extends AdapterBlueprint {
...
+
+      const connection = this.connectionManager?.getConnection({
+        connectorId: params.id,
+        connections: this.connections,
+        connectors: this.connectors
+      })
+
+      switch (connector.type) {
+        case 'WALLET_CONNECT':
```
enesozturk: Can you use constants instead of hard coded string? Same for 'AUTH' and others

packages/adapters/bitcoin/src/adapter.ts
tomiir: Good point, at least 1 account should suffice

0xmkh: ah yeah this is used to format ordinal and payment bitcoin addresses if more than 1 account is returned

packages/scaffold-ui/src/views/w3m-profile-wallets-view/index.ts
0xmkh: i was trying out demo and the tab width was messed up, so created this fn to determine what the tab width should be based off of its offsetWidth

packages/scaffold-ui/src/partials/w3m-connect-featured-widget/index.ts
```diff
@@ -23,6 +24,10 @@ export class W3mConnectFeaturedWidget extends LitElement {
       return null
     }
 
+    const hasConnection = ConnectionController.hasAnyConnection(
```
enesozturk: These might be
```suggestion
    const hasWcConnection = ConnectionController.hasAnyConnection(
```

enesozturk: And for other files too

0xmkh: yep

enesozturk (approved):

bogdan:
You introduced this big feature that changes everything when I was updating from 1.7.3 to 1.7.15 - suppose to be patches only. Did I miss something?

PR #4509, opened 2025-06-16
fix: syncning connections when using different connectors on different namespaces on initialization

enesozturk:
# Description

Fixes issue on initialization when syncing the connection for multi-chain use cases. Two issues connected to each other:

1. Bitcoin adapter does get active network from all namespaces; if active namespace is not bip122, Bitcoin adapter's `connect` method fails and it cannot reconnect the connector - fixes this as to filter networks by namespace in the Bitcoin adapter's `getActiveNetwork` method
2. UP's `chainChanged` method calling `setCaipNetwork` where it's not respecting the active network on AppKit; see the following example flow:
   - Connect to Bitcoin adapter with an extension
   - Connect to EVM and Solana with WC (i.e Trust iOS supports both)
   - Make active network Bitcoin on AppKit
   - Refresh page
   - The 'chainChanged' is triggered and AppKit switches back to EVM, which is unexpected
3. Removed `setConnectorId` call in `connectWalletConnect`; In the flow mentioned in 2), when connecting with Trust iOS app, AppKit set's the Bitcoin's `@appkit/bip122_connected_connector_id` storage value to `WALLET_CONNECT`, which is wrong. Since Trust connection supports only EIP155 and Solana, it shouldn't update BIP122 values.

## Type of change

# Associated Issues

For GH issues: closes #...

# Showcase (Optional)

Before

https://github.com/user-attachments/assets/146b1f14-cc27-4aaa-9e49-513839000081

After

https://github.com/user-attachments/assets/c67aa055-5d45-4dd7-9f67-bc5ff14daada

# Checklist


packages/appkit/src/client/appkit-base-client.ts
```diff
@@ -466,12 +466,6 @@ export abstract class AppKitBaseClient {
         this.close()
         this.setClientId(result?.clientId || null)
         StorageUtil.setConnectedNamespaces([...ChainController.state.chains.keys()])
-        this.chainNamespaces.forEach(namespace => {
```
enesozturk, line 469: We are already doing the same inside the `syncWalletConnectAccount` method

packages/appkit/src/client/appkit-base-client.ts
```diff
@@ -1380,7 +1374,10 @@ export abstract class AppKitBaseClient {
           return
         }
 
-        if (currentCaipNetwork?.id !== caipNetwork?.id) {
+        if (
```
enesozturk, line 1377: Adding namespace clause here to make sure to change active namespace if I'm already on the same namespace - this will prevent unnecessary network switches on multi-chain use cases

tomiir (approved):

PR #4519, opened 2025-06-17
Chore: add script to check version mismatch postinstall 

svenvoskamp:
# Description

Chore: add script to check version mismatch postinstall 

I tested it with the following config in lab:

```
 "@reown/appkit": "workspace:*",
 "@reown/appkit-adapter-ethers": "1.7.0",
 "@reown/appkit-adapter-wagmi": "1.6.9",
 ```

Result:

``` 
Reown AppKit version mismatch detected!
│    Expected all @reown/appkit packages to be version 1.7.9
│    Mismatched packages:
│      • @reown/appkit-adapter-ethers: 1.7.0 (expected 1.7.9)
│      • @reown/appkit-adapter-wagmi: 1.6.9 (expected 1.7.9)
│    Please update all @reown/appkit packages to version 1.7.9
│    You can run the following commands:
│      npm install @reown/appkit-adapter-ethers@1.7.9
│      npm install @reown/appkit-adapter-wagmi@1.7.9
└─ Failed in 146ms
```

Happy path:

```
installed @reown/appkit packages and versions:
│   @reown/appkit: 1.7.10
|   @reown/appkit-adapter-bitcoin: 1.7.10
│   @reown/appkit-adapter-ethers: 1.7.10
│   @reown/appkit-adapter-ethers5: 1.7.10
│   @reown/appkit-adapter-solana: 1.7.10
│   @reown/appkit-adapter-wagmi: 1.7.10
│   @reown/appkit-common: 1.7.10
│   @reown/appkit-experimental: 1.7.10
│   @reown/appkit-siwe: 1.7.10
│   @reown/appkit-siwx: 1.7.10
│   @reown/appkit-wallet: 1.7.10
│   @reown/appkit-wallet-button: 1.7.10
│   @reown/appkit-pay: 1.7.10
└─ Done in 87ms appkit packages are in sync (1.7.10)
```

## Type of change

# Associated Issues

For GH issues: closes #...

# Showcase (Optional)

# Checklist


packages/appkit/scripts/appkit-version-check.js
```diff
@@ -0,0 +1,176 @@
...
+      } catch (error) {
+        // Invalid package.json, continue searching
+      }
+    }
+
+    current = dirname(current)
+  }
+
+  return null
+}
```
tomiir, line 48: we should have a global trycatch, else users might be unable to install if something fails here

packages/appkit/scripts/appkit-version-check.js
```diff
@@ -0,0 +1,176 @@
...
+    } catch (error) {
+      // Ignore error
+    }
+  }
+
+  return null
+}
+
+// Try multiple approaches to find the consumer project
+let project = findProjectRoot(__dirname) || findProjectRootAlternative()
```
tomiir: can we wrap this code into functions instead of having it at top level? might cause some issues with older node versions

tomiir: also same as above everything should have a try catch

svenvoskamp:
> @svenvoskamp what happens when postinstall fails. is it blocking for the dev?

@rocky yes it blocks the install, see `process.exit(1)`, I can remove that one? 

Will add a try catch, so it never blocks the dev if it fails and can also remove the process.exit, but then it might get easily ignored tho

packages/appkit/scripts/appkit-version-check.js
```diff
@@ -0,0 +1,213 @@
...
+
+const require = createRequire(import.meta.url)
+const __dirname = dirname(fileURLToPath(import.meta.url))
+
+/* CONFIG ------------------------------------------------------------------ */
+const REOWN_SCOPE_PREFIX = '@reown/appkit'
+const sections = ['dependencies', 'devDependencies', 'peerDependencies', 'optionalDependencies']
+/* ------------------------------------------------------------------------- */
+
+function findProjectRoot(startPath) {
```
enesozturk, line 17: Is this logics working on monorepos? Also since this is going to work on consumer's devices we should be extra careful that this logic is not going on a endless loop, will ignore files like node_modules etc. Would be nice to cover with unit tests

svenvoskamp: If you install your app in the monorepo it will work, but not if you call it from the root. (It won't find your specific app package.json)

It will only check if it can find the root package.json and then check the version. It will only check the sections and won't go in the node_modules

PR #4544, opened 2025-06-20
fix: show unsupportedChainUI if wagmi is unable to switch network and enableNetworkSwitch is disabled 

svenvoskamp:
# Description

In some cases when a wallet is connecting to AppKit and is not connecting to the right network and there are only a few networks enabled; Wagmi is unable to switch to the chain because this wallet is not enabled in wagmi. In this case AppKit will still connect and will switch to the wallets chain. In some cases this will open the `unsupportedChainUI` if the network is not supported.

But there are also cases when `enableNetworkSwitch` is turned off. In this case wagmi will still connect, fail to switch chain, but AppKit keeps in the wrong chain. We need to simulate a `showUnsupportedChainUI` to show the user the correct error on the screen, so they can manually switch to the correct network.

## Type of change

# Associated Issues

For Linear issues: Closes APKT-3076

# Showcase (Optional)

# Checklist

### DIFF
--- a/packages/scaffold-ui/src/views/w3m-connecting-wc-view/index.ts
+++ b/packages/scaffold-ui/src/views/w3m-connecting-wc-view/index.ts
@@ -111,7 +111,6 @@ export class W3mConnectingWcView extends LitElement {
       }
     } catch (error) {
       /*
-
        * In some cases when a wallet is connecting to AppKit and is not connecting to the right network and there are only a few networks enabled; Wagmi is unable to switch to the chain because this wallet is not enabled in wagmi. In this case AppKit will still connect and will switch to the wallets chain. In some cases this will open the unsupportedChainUI if the network is not supported.
        *
        * But there are also cases when enableNetworkSwitch is turned off. In this case wagmi will still connect, fail to switch chain, but AppKit keeps in the wrong chain. We need to simulate a showUnsupportedChain to show the user the correct error on the screen, so they can manually switch to the correct network.


packages/scaffold-ui/src/views/w3m-connecting-wc-view/index.ts
```diff
@@ -108,6 +110,28 @@ export class W3mConnectingWcView extends LitElement {
...
+        error.message.includes('An error occurred when attempting to switch chain') &&
+        !OptionsController.state.enableNetworkSwitch
+      ) {
+        if (ChainController.state.activeChain) {
+          ChainController.setActiveCaipNetwork(
+            CaipNetworksUtil.getUnsupportedNetwork(
+              `${ChainController.state.activeChain}:${ChainController.state.activeCaipNetwork?.id}`
+            )
+          )
+          ChainController.showUnsupportedChainUI()
```
0xmkh, line 129: Do you know if we can do this from the wagmi adapter ?

```ts
this.wagmiConfig?.setState(x => ({
  ...x,
  chainId: 999999999999
}))
```

svenvoskamp: We can do that but then still AppKit won't trigger the unsupportedNetwork view as `enableNetworkSwitch` is off. We need to manually trigger it with this check

packages/scaffold-ui/src/views/w3m-connecting-wc-view/index.ts
```diff
@@ -108,6 +110,28 @@ export class W3mConnectingWcView extends LitElement {
...
+      /*
+       * In some cases when a wallet is connecting to AppKit and is not connecting to the right network and there are only a few networks enabled; Wagmi is unable to switch to the chain because this wallet is not enabled in wagmi. In this case AppKit will still connect and will switch to the wallets chain. In some cases this will open the unsupportedChainUI if the network is not supported.
+       *
+       * But there are also cases when enableNetworkSwitch is turned off. In this case wagmi will still connect, fail to switch chain, but AppKit keeps in the wrong chain. We need to simulate a showUnsupportedChain to show the user the correct error on the screen, so they can manually switch to the correct network.
+       */
+      if (
+        error instanceof Error &&
+        error.message.includes('An error occurred when attempting to switch chain') &&
+        !OptionsController.state.enableNetworkSwitch
+      ) {
```
tomiir, line 122: given this is specific to wagmi, why not do it in the wagmi adapter directly?

svenvoskamp: Because then still the other code from the catch block will execute and causes strange behaviour

tomiir (approved):

packages/scaffold-ui/src/views/w3m-connecting-wc-view/index.ts
tomiir: if we can avoid bubbling up the error so that catch block never executes, I'd prefer that, else I left an approve

svenvoskamp: Tried a few things and still executes, so I prefer to keep it like this

PR #4575, opened 2025-06-27
feat: add 1CA to embedded wallets 

svenvoskamp:
# Description

This PR adds 1CA flow to the embedded wallet. Instead of directly calling the `w3mFrameProvider`, we now call a new introduced method in the SIWXUtil called `authConnectorAuthenticate`. This method will check if `SIWX` is enabled and will then add the `SIWXMessage` to the `w3mFrameProvider` connect method. 

It will also handle the signature that we get back from the `w3mFrameProvider` and will add the session accordingly. 

More changes:

* Test changes 
* Call `SIWXUtil` method in clients 
* Handle a case where `onSocialConnected` is triggering before we've added the session to `SIWX`

## Type of change

# Associated Issues

For GH issues: closes #...

# Showcase (Optional)

# Checklist


packages/appkit/src/client/appkit.ts
```diff
@@ -64,7 +65,30 @@ export class AppKit extends AppKitBaseClient {
 
   // -- Private ------------------------------------------------------------------
 
-  private onAuthProviderConnected(user: W3mFrameTypes.Responses['FrameGetUserResponse']) {
+  private async onAuthProviderConnected(user: W3mFrameTypes.Responses['FrameGetUserResponse']) {
+    if (user.message && user.signature && user.siwxMessage) {
+      // OnAuthProviderConnected is getting triggered when we receive a success event on Social / Email login. Add this moment, if SIWX is enabled, we are still adding the session to SIWX. Await this promise to make sure that the modal doesn't show the SIWX Sign Message UI
```
enesozturk:
```suggestion
      // OnAuthProviderConnected is getting triggered when we receive a success event on Social / Email login. At this moment, if SIWX is enabled, we are still adding the session to SIWX. Await this promise to make sure that the modal doesn't show the SIWX Sign Message UI
```

packages/controllers/src/utils/SIWXUtil.ts
```diff
@@ -215,6 +218,110 @@ export const SIWXUtil = {
...
+
+      return {
+        address: result.address,
+        chainId: result.chainId,
+        accounts: result.accounts
+      }
+    }
+
+    const siwxMessage = await siwx.createMessage({
+      chainId: ChainController.getActiveCaipNetwork()?.caipNetworkId || ('' as CaipNetworkId),
```
enesozturk, line 251: Can we handle possible null value as a early return or error throw?

packages/controllers/src/utils/SIWXUtil.ts
```diff
@@ -215,6 +218,110 @@ export const SIWXUtil = {
...
+    })
+
+    if (result.signature && result.message) {
+      const promise = SIWXUtil.addEmbeddedWalletSession(
+        siwxMessageData,
+        result.message,
+        result.signature
+      )
+
+      await promise
```
enesozturk, line 285: Why not
```suggestion
      await SIWXUtil.addEmbeddedWalletSession(
        siwxMessageData,
        result.message,
        result.signature
      )

      await promise
```

enesozturk: Same for the one on the `appkit.ts`

packages/scaffold-ui/test/views/w3m-email-verify-otp-view.test.ts
```diff
@@ -145,7 +145,6 @@ describe('W3mEmailVerifyOtpView', () => {
 
       expect(RouterController.reset).not.toHaveBeenCalled()
       expect(RouterController.push).not.toHaveBeenCalled()
-      expect(ModalController.close).not.toHaveBeenCalled()
```
enesozturk, line 148: Is it bc it's directly navigating to SIWX message?

enesozturk (approved):
Left non blocking comments

packages/scaffold-ui/test/views/w3m-email-verify-otp-view.test.ts
svenvoskamp: Normally we expect the Sign Messag SIWE screen, but now it should close as we directly authenticate

packages/controllers/src/utils/SIWXUtil.ts
svenvoskamp: If we `await` here, there can be a race condition where `onSocialConnect` triggers and triggers this method, but also the response from `connect` can trigger this. We need to await in these specific methods and return the promise immediately to prevent this

packages/controllers/src/utils/SIWXUtil.ts
svenvoskamp: It will always create a message and will otherwise be handled on secure site, so we default back to a specific value.

PR #4600, opened 2025-07-02
refactor: remove type castings and handle errors instead

enesozturk:
# Description

- Removes type casting for making sure we don't override any possible issues and throw errors if values are not expected
- Removes type casting for custom `0x${string}` and use `@reown/appkit-common` package for for types and namespace values

## Type of change

# Associated Issues

For GH issues: closes #...

# Showcase (Optional)

# Checklist

### DIFF
--- a/packages/experimental/src/smart-session/controllers/SmartSessionsController.ts
+++ b/packages/experimental/src/smart-session/controllers/SmartSessionsController.ts
@@ -226,7 +226,7 @@ export const SmartSessionsController = {
         pci: session.pci,
         permissions: [...session.permissions.map(p => JSON.parse(JSON.stringify(p)))],
         expiry: Math.floor(session.expiry / 1000),
-        address: chainAndAddress.address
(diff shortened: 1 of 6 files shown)


packages/experimental/src/smart-session/controllers/SmartSessionsController.ts
```diff
@@ -224,15 +226,11 @@ export const SmartSessionsController = {
         pci: session.pci,
         permissions: [...session.permissions.map(p => JSON.parse(JSON.stringify(p)))],
         expiry: Math.floor(session.expiry / 1000),
-        address: activeCaipAddress as `0x${string}`
+        address: chainAndAddress.address
```
enesozturk: Found here that `revokePermissions` expect address value but we are passing CAIP address, so this should be the right one. But maybe cannot see something @tomiir this is an old code would like to validate with u if we should pass CAIP address or Hex value?

PR #4605, opened 2025-07-03
refactor: separating testing utils, update bad practice on unit tests, remove redundant controller utils

enesozturk:
# Description

- Updating testing reset approach: We are calling `clearAllMocks` in many places of the unit tests where this function doesn't reset the state of the controllers and we should avoid. Instead, `restoreAllMocks` brings the controllers to the initial state. Where it helps to isolate every test.
- Removing redundant utils from controllers: We have `getAccountProp`, `getActiveCaipNetwork` and such utils in the controllers. Controllers should be responsible for only managing the state. These kind of utils should be in the utils folder.
- Testing utils: Exporting utils from the controllers package to use mocking the controllers in the unit tests

## Type of change

# Associated Issues

For GH issues: closes #...

# Showcase (Optional)

# Checklist


packages/appkit/src/client/appkit-base-client.ts
```diff
@@ -431,7 +431,7 @@ export abstract class AppKitBaseClient {
       }
 
       const adapter = this.getAdapter(namespace)
-      const { caipAddress } = ChainController.getAccountData(namespace) || {}
+      const caipAddress = ChainController.state.chains.get(namespace)?.accountState?.caipAddress
```
0xmkh, line 434: wouldn't `ChainController.getAccountData` be cleaner ?

packages/controllers/tests/hooks/vue.test.ts
```diff
@@ -64,33 +54,71 @@ describe('useAppKitAccount', () => {
...
     const [state] = withSetup(() => useAppKitAccount())
 
-    AccountController.setCaipAddress('eip155:1:0x123...', 'eip155')
-    AccountController.setStatus('connected', 'eip155')
-    AccountController.setUser({ username: 'test', email: 'testuser@example.com' }, 'eip155')
-    AccountController.setSmartAccountDeployed(true, 'eip155')
-    AccountController.setPreferredAccountType('smartAccount', 'eip155')
+    ChainController.state.activeChain = ConstantsUtil.CHAIN.EVM
+    // @ts-expect-error ignore type error
+    ChainController.state.chains = new Map([
```
0xmkh, line 105: wouldn't mocking here be better ?

magiziz (approved):

packages/appkit/src/client/appkit.ts
```diff
@@ -374,8 +375,7 @@ export class AppKit extends AppKitBaseClient {
             AlertController.open(ErrorUtil.ALERT_ERRORS.UNVERIFIED_DOMAIN, 'error')
           }
         },
-        getActiveCaipNetwork: (namespace?: ChainNamespace) =>
-          ChainController.getActiveCaipNetwork(namespace)
+        getActiveCaipNetwork: (namespace?: ChainNamespace) => getActiveCaipNetwork(namespace)
```
svenvoskamp, line 378: Isn't this more confusing now, having several `getActiveCaipNetwork` methods that are not tied to a controller anymore?

packages/appkit/src/client/appkit-base-client.ts
enesozturk: We are using `getAccountData` in really few places, doesn't seem very necessary. And it's just a getter where I think shouldn't be located in ChainController

packages/appkit/src/client/appkit.ts
enesozturk: I think `getActiveCaipNetwork` tells enough what it does, doesn't matter where you're calling it anywhere in the code, it's just returns the active network object from state. And I think later we'd need more separated components where they are tree-shakable, correct me if I'm wrong @tomiir

packages/controllers/tests/hooks/vue.test.ts
enesozturk: I've tried that, somehow the mockChainControllerState() is not working with Vue tests, I'm thinking it's somehow related Vue's life cycle methods when it comes to testing, it's not respecting to `spyOn` and the hooks not getting triggerred with new data. So specificially to Vue tests, mutating the state directly

PR #4744, opened 2025-07-31
chore: fix missing `address` in WC success event

magiziz:
# Description

Fixed an issue where WalletConnect connections didn't include the `address` value when sending the `CONNECT_SUCCESS` event

## Type of change

# Associated Issues

For Linear issues: Closes APKT-3440

# Checklist

### DIFF
--- a/packages/controllers/src/controllers/EventsController.ts
+++ b/packages/controllers/src/controllers/EventsController.ts
@@ -51,7 +51,8 @@ export const EventsController = {
 
   async _sendAnalyticsEvent(payload: EventsControllerState) {
     try {
-      const address = AccountController.state.address
+      const address =
+        'address' in payload.data ? payload.data.address : AccountController.state.address
       if (excluded.includes(payload.data.event) || typeof window === 'undefined') {
         return
       }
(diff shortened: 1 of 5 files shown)


packages/controllers/src/controllers/EventsController.ts
```diff
@@ -51,7 +51,12 @@ export const EventsController = {
 
   async _sendAnalyticsEvent(payload: EventsControllerState) {
     try {
-      const address = AccountController.state.address
+      let address = AccountController.state.address
+
+      if ('address' in payload.data && payload.data.address) {
+        address = payload.data.address
+      }
+
```
enesozturk, line 59: This is a bit making things confusing, what are the cases where we send address in payload and what are the cases we don't - if we have both cases, is that intentional? 

I mean when we call `CONNECT_SUCCESS` event, why AccountController.state.address not working?

enesozturk: Ah okok - prob we don't even set the address yet

0xmkh: Yeah there's a slight delay on WC connection. In our appkit class when WC triggers `onConnect` the `AccountController.state.address` doesn't update right away it happens a bit later

svenvoskamp (approved):

PR #4750, opened 2025-07-31
feat: reown authentication as remote feature

zoruka:
# Description

This PR is moving `ReownAuthentication` class into `@reown/appkit-controllers` package and implementing an easy integration through AppKit features and remote features.

## Type of change

# Associated Issues

For GH issues: closes #...

# Showcase (Optional)

# Checklist

### DIFF
--- a/packages/scaffold-ui/test/views/w3m-data-capture-view.test.ts
+++ b/packages/scaffold-ui/test/views/w3m-data-capture-view.test.ts
@@ -216,14 +216,14 @@ describe('W3mDataCaptureView', () => {
       const view = await createView()
 
       // Set a valid email and attempt submit
-      const emailInput = HelpersUtil.querySelector(view, 'wui-email-input')
+      const emailInput = HelpersUtil.querySelect(view, 'wui-email-input')
       emailInput?.dispatchEvent(new CustomEvent('inputChange', { detail: 'test@example.com' }))
       await view.updateComplete
 
       // Spy on requestEmailOtp on prototype to detect accidental calls
       const spy = vi.spyOn(ReownAuthentication.prototype as any, 'requestEmailOtp')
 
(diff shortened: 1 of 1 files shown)


packages/controllers/exports/features.ts
```diff
@@ -0,0 +1 @@
+export * from '../src/features/reown-authentication/index.js'
```
tomiir, line 1: why add this here?

tomiir: i mean it's not wrong, just weird we are exporting a `features` export in our `controllers` package which is intended as a central package for controlling states of ak

packages/appkit/src/utils/ConfigUtil.ts
```diff
@@ -194,6 +195,21 @@ const featureConfig = {
...
+  },
+  reownAuthentication: {
+    apiFeatureName: 'reown_authentication' as const,
+    localFeatureName: 'reownAuthentication',
+    returnType: false as boolean,
+    isLegacy: false,
+    isAvailableOnBasic: false,
+    processApi: (apiConfig: TypedFeatureConfig) => Boolean(apiConfig.isEnabled),
+    processFallback: (localValue: unknown): boolean => {
+      if (typeof localValue === 'undefined') {
```
tomiir, line 207: why do you need this check ?

packages/controllers/exports/features.ts
zoruka: The goal is to have ReownAuthentication somewhere inside the main packages, so we can enable it through the dashboard without installing a new package. Would you suggest a new place where we can put this file? Should convert it into a controller?

packages/appkit/src/utils/ConfigUtil.ts
zoruka: I did get it from the others features, it verifies if something was not set in the local configuration before using the default value.

tomiir (approved):

PR #4885, opened 2025-08-21
refactor: remove redundant project id not configured error triggers

enesozturk:
# Description

In AppKit base client we have `checkAlloweOrigins` method where we call redundant `ErrorUtil.ALERT_ERRORS.PROJECT_ID_NOT_CONFIGURED` alerts:

- `!allowedOrigins || !CoreHelperUtil.isClient()` clause: The `!allowedOrigins` here is deadly code where it never goes falsy as we handle the errors on `fetchAllowedOrigins` method. If the error is not known, we return empty array, where it's not falsy. Only isClient() check left, and if the code runs on server side, we should not proceed running the function.
- As we handle errors on the `fetchAllowedOrigins` and return empty array for other types of errors, calling `ErrorUtil.ALERT_ERRORS.PROJECT_ID_NOT_CONFIGURED` on switch case's `default` is dead code.
 
## Type of change

# Associated Issues

For GH issues: closes #...

# Showcase (Optional)

# Checklist

### DIFF
--- a/packages/appkit/src/client/appkit-base-client.ts
+++ b/packages/appkit/src/client/appkit-base-client.ts
@@ -243,7 +243,7 @@ export abstract class AppKitBaseClient {
           break
         }
         default:
-          return
+          break
       }
     }
   }
--- a/packages/appkit/src/client/appkit-base-client.ts
+++ b/packages/appkit/src/client/appkit-base-client.ts
@@ -206,9 +206,7 @@ export abstract class AppKitBaseClient {
     try {
       const allowedOrigins = await ApiController.fetchAllowedOrigins()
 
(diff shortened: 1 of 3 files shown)


packages/appkit/src/client/appkit-base-client.ts
```diff
@@ -247,7 +243,7 @@ export abstract class AppKitBaseClient {
           break
         }
         default:
-          AlertController.open(ErrorUtil.ALERT_ERRORS.PROJECT_ID_NOT_CONFIGURED, 'error')
```
0xmkh, line 250: why did we always display this by default 🤔?

enesozturk: By Devin, but seems like we didn't reviewed well enough

PR #4888, opened 2025-08-22
feat: connect all namespaces for auth connector

enesozturk:
# Description

Updates connectExternal method to handle Auth connections to be connected for all namespaces. When user connected with auth provider, we should sync all supported namespace's accounts to AppKit, like WC.

In this PR, iterates remaining (non active) Auth supported namespaces to call .connect and syncs account info for that namespace. Makes relevant changes to AppKit class and some controllers and adds e2e tests

## Type of change

# Associated Issues

For GH issues: closes #...

# Showcase (Optional)

# Checklist


apps/laboratory/tests/shared/pages/ModalPage.ts
```diff
@@ -407,7 +407,7 @@ export class ModalPage {
     }
 
     await expect(this.page.getByText(headerTitle)).not.toBeVisible({
-      timeout: 20_000
+      timeout: 30_000
```
enesozturk, line 410: This one is weird - for last few weeks we had issues on Farcaster tests including this PR.

I've did some investigation and seemed like after pasting OTP, it's not resolving the getUser in 20 secs. But actually it's resolving eventually, it's just taking more. I've updated the timeout slightly and it helped.

What we do in Farcaster is that we reload the secure site if the user opens Farcaster page and go back, just to clear Farcaster state on secure site, we reload it - introduced here: https://github.com/reown-com/appkit/pull/3527/files But tbh I don't really now why just reloading secure site causes requests to be resolved slower.

apps/laboratory/tests/email-after-farcaster.spec.ts
```diff
@@ -94,6 +94,7 @@ emailTestAfterFarcaster(
     await page.closeModal()
     await validator.expectCaipAddressHaveCorrectNetworkId(caipNetworkId as CaipNetworkId)
 
+    await page.page.waitForTimeout(1000)
```
enesozturk: Similar for this one, after switching network clicking sign immediately causes error bc actually the switching not completed on secure site yet. Putting a 1s delay before signing message helps, but it's synthetic for sure. I'm curious why this is not happening on other tests but Farcaster.

0xmkh: is it possible to do something like `waitFor(...)`?

enesozturk: You mean waiting for some element to be visible before signing? We already doing that - and that's the problem actually, on AppKit the network seems switched, but on secure site it takes additional time to be switched @magiziz

packages/appkit-utils/src/HelpersUtil.ts
```diff
@@ -75,5 +79,44 @@ export const HelpersUtil = {
...
+      case 'eip155':
+        return CommonConstantsUtil.CHAIN.EVM
+      case 'solana':
+        return CommonConstantsUtil.CHAIN.SOLANA
+      case 'bip122':
+        return CommonConstantsUtil.CHAIN.BITCOIN
+      default:
+        return namespace as ChainNamespace
+    }
+  },
```
0xmkh: maybe we do this?
```ts
userChainIdToChainNamespace(chainId: number | string) {
  if (typeof chainId === 'number') {
    return CommonConstantsUtil.CHAIN.EVM
  }

  const [namespace] = chainId.split(':')

  const NAMESPACE_TO_CHAIN: Record<string, ChainNamespace> = {
    eip155: CommonConstantsUtil.CHAIN.EVM,
    solana: CommonConstantsUtil.CHAIN.SOLANA,
    bip122: CommonConstantsUtil.CHAIN.BITCOIN,
  }

  return NAMESPACE_TO_CHAIN[namespace] ?? (namespace as ChainNamespace)
}
```

apps/laboratory/tests/email-after-farcaster.spec.ts
```diff
@@ -94,6 +94,7 @@ emailTestAfterFarcaster(
     await page.closeModal()
     await validator.expectCaipAddressHaveCorrectNetworkId(caipNetworkId as CaipNetworkId)
 
+    await page.page.waitForTimeout(2000)
```
0xmkh, line 97: can we avoid using timeout and instead use `waitFor(...)`?

0xmkh: as in like we can wait until element has performed the action

magiziz (approved):

tomiir (approved):
the appkit-base-client chagnes are code formatting / nicer code but no functional changes right?

PR #4947, opened 2025-09-03
fix: bitcoin account publickey doesn't set on reconnect

enesozturk:
# Description

When reconnecting the adapter connections, the publicKey field of the Bitcoin account doesn't get set as expected.

Two problem identified on the base class' `initialize` method:
- We are calling both `await this.syncExistingConnection()` and `await this.syncAdapterConnections()` and this causes getting the connection two times which is unnecessary.
- When getting the account with `await this.syncExistingConnection()`, the publicKey returns under allAccounts objects, but `await this.syncAdapterConnections()` doesn't return the `publicKey` as expected. So when this function run after `syncExistingConnection`, the publicKey lost.

## Type of change

# Associated Issues

For GH issues: closes #...

# Showcase (Optional)

# Checklist


apps/laboratory/src/components/AppKitInfoMultiChain.tsx
```diff
@@ -150,7 +150,7 @@ function AccountCard({
                 <Box display="flex" alignItems="center" gap={2}>
                   {walletInfo?.icon ? (
                     <Image
-                      src={walletInfo.icon}
+                      src={walletInfo.icon.trim()}
```
enesozturk, line 153: Fixes issue on lab that Gancho raised recently about image src with empty lines

apps/laboratory/src/layout/LayoutHeader.tsx
```diff
@@ -44,16 +43,16 @@ function downloadLogs(toast: ReturnType<typeof useChakraToast>) {
...
   const controlsCW = useDisclosure({ id: 'customWallet' })
   const controlsNW = useDisclosure({ id: 'networks' })
   const controlsPID = useDisclosure({ id: 'projectId' })
   const toast = useChakraToast()
   const { colorMode } = useColorMode()
   const [origin, setOrigin] = useState('')
+
   useEffect(() => {
-    setOrigin(window.location.origin)
+    setOrigin(window.location.href)
```
enesozturk, line 55: Adding full url to lab

packages/adapters/bitcoin/src/adapter.ts
```diff
@@ -227,7 +226,7 @@ export class BitcoinAdapter extends AdapterBlueprint<BitcoinConnector> {
       universalProvider: this.universalProvider as UniversalProvider,
       onConnection: this.addConnection.bind(this),
       onListenProvider: this.listenProviderEvents.bind(this),
-      getConnectionStatusInfo: getConnectorStorageInfo
+      getConnectionStatusInfo: HelpersUtil.getConnectorStorageInfo.bind(this)
```
enesozturk: No need to drive `getConnectorStorageInfo`, everytime when we call `syncConnections`, I'm moving that function to `HelpersUtil` and passing here instead

packages/appkit/src/connections/ConnectionManager.ts
```diff
@@ -256,7 +259,12 @@ export class ConnectionManager {
             onListenProvider(connector.id, connector.provider as BitcoinConnector)
             onConnection({
               connectorId: connector.id,
-              accounts: accounts.map(a => ({ address: a.address, type: a.type })),
+              accounts: accounts.map(a => ({
```
enesozturk, line 261: Actual fix

packages/adapters/bitcoin/src/adapter.ts
tomiir: why bind to `this`?

packages/appkit/src/connections/ConnectionManager.ts
tomiir: do we need to manually map? can we just spread?

packages/adapters/bitcoin/src/adapter.ts
0xmkh: yh same question

enesozturk: I think no need yeah, at first the functions was a bit different, not was from the HelpersUtil, updating

packages/appkit-utils/src/HelpersUtil.ts
```diff
@@ -110,5 +115,23 @@ export const HelpersUtil = {
...
+  getConnectorStorageInfo(connectorId: string, namespace: ChainNamespace) {
+    const storageConnectionsByNamespace = StorageUtil.getConnections()
+    const storageConnections = storageConnectionsByNamespace[namespace] ?? []
+
+    return {
+      hasDisconnected: StorageUtil.isConnectorDisconnected(connectorId, namespace),
+      hasConnected: storageConnections.some(c =>
+        HelpersUtil.isLowerCaseMatch(c.connectorId, connectorId)
+      )
+    }
```
0xmkh, line 135: Is it possible to use `ConnectionController.state.recentConnections`?

packages/appkit/src/adapters/ChainAdapterBlueprint.ts
```diff
@@ -755,7 +755,10 @@ export namespace AdapterBlueprint {
   export type SyncConnectionsParams = {
     connectToFirstConnector: boolean
     caipNetwork?: CaipNetwork
-    getConnectorStorageInfo: (connectorId: string) => {
+    getConnectorStorageInfo: (
+      connectorId: string,
+      namespace: ChainNamespace
+    ) => {
```
0xmkh: maybe it can be a param?

packages/appkit/src/client/appkit-base-client.ts
```diff
@@ -1241,34 +1241,24 @@ export abstract class AppKitBaseClient {
...
-            const storageConnections = storageConnectionsByNamespace[namespace] ?? []
-
-            return {
-              hasDisconnected: StorageUtil.isConnectorDisconnected(connectorId, namespace),
-              hasConnected: storageConnections.some(c =>
-                HelpersUtil.isLowerCaseMatch(c.connectorId, connectorId)
-              )
-            }
-          }
+          getConnectorStorageInfo: HelpersUtil.getConnectorStorageInfo.bind(this)
```
0xmkh: why `.bind(this)`?

packages/appkit/src/connections/ConnectionManager.ts
enesozturk: `address` obj has `namespace` as well, which we don't need. Assuming we are storing these values on storage, better to keep it smaller

0xmkh: Do we need to migrate current `@appkit/connections` local storage connections so it's compatible with the new one? E.g some old accounts won't have path, publicKey and etc

packages/appkit-utils/src/HelpersUtil.ts
enesozturk: Seems like we don't have that function?

magiziz (approved):

PR #4962, opened 2025-09-04
feat: `openSend`

magiziz:
# Description

Added `openSend` function to customize the send flow experience by pre-configuring the token, amount, chain, and recipient address

## Type of change

# Associated Issues

For Linear issues: Closes APKT-3632,APKT-3628
For GH issues: closes #...

# Showcase (Optional)

example

image

image

# Checklist


packages/appkit-utils/src/PresetsUtil.ts
```diff
@@ -115,6 +115,13 @@ export const PresetsUtil = {
     '00000008819873e925422c1ff0f99f7c': 'b3406e4a-bbfc-44fb-e3a6-89673c78b700'
   } as Record<string, string>,
 
+  TOKEN_ADDRESSES_BY_SYMBOL: {
+    USDC: {
+      8453: '0x833589fcd6edb6e08f4c7c32d4f71b54bda02913',
+      84532: '0x036CbD53842c5426634e7929541eC2318f3dCF7e'
+    }
```
tomiir: can we map to `baseUsdc` from `controllers` package?  In `ExchangeUtil` we have a bunch of tokens. 
We could maybe move those to a constants file like `TokenUtil.ts` or simlar and use it. wyt?

packages/appkit/tests/client/appkit-base-client.test.ts
```diff
@@ -51,6 +54,10 @@ describe('AppKitBaseClient.checkAllowedOrigins', () => {
 
       async injectModalUi() {}
       async syncIdentity() {}
+
+      override async syncAdapterConnections() {
```
enesozturk, line 58: Why is this for?

packages/controllers/src/utils/ViemUtil.ts
```diff
@@ -0,0 +1,42 @@
+import { type Chain as ViemChain } from 'viem'
+import { createPublicClient, http } from 'viem'
+import * as viemChains from 'viem/chains'
```
enesozturk: Is this safe to do? It might import all the network objects where might increase the bundle size significantly

packages/appkit/src/client/appkit-base-client.ts
```diff
@@ -2029,14 +2048,77 @@ export abstract class AppKitBaseClient {
...
         case 'Swap':
-          return ModalController.open({ ...options, data: { swap: options.arguments } })
+          return ModalController.open({
+            ...options,
+            data: { swap: options.arguments as ViewArguments['Swap'] }
+          })
+        case 'WalletSend':
+          return ModalController.open({
+            ...options,
+            data: { send: options.arguments as ViewArguments['WalletSend'] }
```
tomiir: why do we need to cast the types? shouldn't be necessary given we are using generics.

packages/appkit/src/client/appkit-base-client.ts
```diff
@@ -1853,6 +1867,11 @@ export abstract class AppKitBaseClient {
   public getCaipNetworks = (namespace?: ChainNamespace) =>
     ChainController.getCaipNetworks(namespace)
 
+  public getCaipNetworkById = (id: string | number) =>
+    this.getCaipNetworks().find(
+      n => n.id.toString() === id.toString() || n.caipNetworkId.toString() === id.toString()
+    )
```
tomiir: why do we need this in the public interface?

packages/appkit/src/client/appkit-base-client.ts
```diff
@@ -2029,14 +2048,77 @@ export abstract class AppKitBaseClient {
...
+
+      if (!caipNetwork) {
+        throw new Error(`openSend: caipNetwork with chainId ${args.chainId} not found`)
+      }
+
+      await this.switchNetwork(caipNetwork)
+    }
+
+    try {
+      const symbol = TokenUtil.getTokenSymbolByAddress(args.assetAddress)
```
tomiir: what happens if `assetAddress` is not supported by balance api ?

packages/appkit/src/client/appkit-base-client.ts
```diff
@@ -2029,14 +2048,77 @@ export abstract class AppKitBaseClient {
...
+      })
+
+      const unsubscribeModal = ModalController.subscribe(modal => {
+        if (!modal.open) {
+          cleanup()
+          reject(new Error('Modal closed'))
+        }
+      })
+
+      const cleanup = this.createCleanupHandler([unsubscribe, unsubscribeModal])
```
tomiir: how is this being used before initialization ??

packages/appkit/src/library/react/index.ts
```diff
@@ -121,11 +121,17 @@ export function useAppKit() {
     await modal?.open(options)
   }
 
+  async function openSend(args: NonNullable<OpenOptions<'WalletSend'>['arguments']>) {
+    const w3mModal = modal as AppKit
```
tomiir: is this needed?

packages/appkit/src/library/react/index.ts
```diff
@@ -121,11 +121,17 @@ export function useAppKit() {
...
+
+    return w3mModal.openSend(args)
+  }
+
   async function close() {
     await modal?.close()
   }
 
-  return { open, close }
+  return { open, openSend, close }
```
tomiir: don't think this should be exposed here 🤔 

either:
- opening send is a specia lcase of `open` 
- it should go on a separate hook

packages/controllers/src/controllers/SendController.ts
```diff
@@ -79,6 +88,10 @@ const controller = {
     }
   },
 
+  setHash(hash: SendControllerState['hash']) {
+    state.hash = hash
+  },
```
tomiir: why do we need this? state should be updated from handlers, not via setters (incoming rule for the repo in the following weeks) we should not expose setters directly as it makes it prone for issues. 
`hash` should be updated by one of the handlers ('d think it would be done on the actual send transaction step)

packages/controllers/src/controllers/SendController.ts
```diff
@@ -266,10 +279,15 @@ const controller = {
...
         amount: params.sendTokenAmount,
-        network: ChainController.state.activeCaipNetwork?.caipNetworkId || ''
+        network: ChainController.state.activeCaipNetwork?.caipNetworkId || '',
+        hash: hash || ''
       }
     })
 
+    if (hash) {
+      SendController.setHash(hash)
+    }
```
tomiir: `state.hash = hash`

packages/controllers/src/utils/AssetUtil.ts
```diff
@@ -125,5 +125,13 @@ export const AssetUtil = {
...
   getChainImage(chain: ChainNamespace) {
     return AssetController.state.networkImages[namespaceImageIds[chain]]
+  },
+
+  getTokenImage(symbol?: string) {
+    if (!symbol) {
+      return undefined
+    }
+
+    return AssetController.state.tokenImages[symbol]
```
tomiir, line 135: wouldn't it be better to fetch it if not available? instead of relying on external flows having set this

packages/controllers/src/utils/BalanceUtil.ts
```diff
@@ -105,5 +122,47 @@ export const BalanceUtil = {
    */
   filterLowQualityTokens(balances: BlockchainApiBalanceResponse['balances']) {
     return balances.filter(balance => balance.quantity.decimals !== '0')
+  },
+  async fetchERC20Balance({ caipAddress, assetAddress, caipNetworkId }: FetchER20BalanceParams) {
```
tomiir: beautiful 👨‍🍳

packages/controllers/src/utils/ConstantsUtil.ts
```diff
@@ -153,6 +153,7 @@ export const ConstantsUtil = {
     CommonConstantsUtil.CHAIN.EVM,
     CommonConstantsUtil.CHAIN.SOLANA
   ] as ChainNamespace[],
+  SEND_PARAMS_SUPPORTED_CHAINS: [CommonConstantsUtil.CHAIN.EVM] as ChainNamespace[],
```
tomiir, line 156: why `SEND_PARAMS` ?

packages/controllers/src/utils/ViemUtil.ts
tomiir: very careful with where we import this file. It's pulling in ALL networks so whatever file imports this will import all of the network list (which is HUGE)

packages/scaffold-ui/src/partials/w3m-input-address/index.ts
```diff
@@ -42,6 +44,25 @@ export class W3mInputAddress extends LitElement {
...
+        <textarea
+          spellcheck="false"
+          ?disabled=${true}
+          autocomplete="off"
+          .value=${this.value ?? ''}
+        >
+           ${this.value ?? ''}</textarea
+        >
+      </wui-flex>`
+    }
```
tomiir: is it a diff ui? If it's the same why not make the other elements uneditable instead of rendering a copy?

packages/scaffold-ui/src/views/w3m-send-confirmed-view/index.ts
```diff
@@ -0,0 +1,61 @@
...
+        flexDirection="column"
+        alignItems="center"
+        gap="4"
+        .padding="${['1', '3', '4', '3'] as const}"
+      >
+        <wui-flex justifyContent="center" alignItems="center" class="icon-box">
+          <wui-icon size="xxl" color="success" name="checkmark"></wui-icon>
+        </wui-flex>
+
+        <wui-text variant="h6-medium" color="primary">You successfully sent asset</wui-text>
```
tomiir, line 37: Can we use the asset's name here? 
Also in case we don't have it, typo? `You have successfully sent the asset` (I'd check w sam)

packages/scaffold-ui/src/views/w3m-wallet-send-preview-view/index.ts
```diff
@@ -160,8 +162,12 @@ export class W3mWalletSendPreviewView extends LitElement {
 
     try {
       await SendController.sendToken()
-      SnackController.showSuccess('Transaction started')
-      RouterController.replace('Account')
+      if (this.params) {
```
tomiir, line 165: do you need to check which params?

packages/scaffold-ui/src/views/w3m-wallet-send-view/index.ts
```diff
@@ -69,18 +82,30 @@ export class W3mWalletSendView extends LitElement {
...
+
   // -- Render -------------------------------------------- //
   public override render() {
     this.getMessage()
 
+    const isReadOnly = Boolean(this.params)
+
+    if (!this.address) {
+      throw new Error('w3m-wallet-send-view: No account connected')
+    }
```
tomiir: we said we were not doing the route to connection screen flow now right? 
would be cool to bring it up to sam/chris

packages/appkit/tests/client/appkit-base-client.test.ts
0xmkh: i included this since otherwise it'd crash the test file because `syncAdapterConnections` is not defined

packages/controllers/src/utils/ViemUtil.ts
0xmkh: nope not safe at all, refactored 🙏

packages/appkit-utils/src/PresetsUtil.ts
0xmkh: oh good idea

packages/appkit/src/client/appkit-base-client.ts
0xmkh: ah yeah i got bunch of type errors, refactored it 🙏

packages/appkit/src/client/appkit-base-client.ts
0xmkh: good catch, made it private

packages/appkit/src/client/appkit-base-client.ts
0xmkh: I think it's used after initialization. Idea here is when modal closes or payment is successful we finish listening to the subscribe key events

packages/appkit/src/client/appkit-base-client.ts
0xmkh: then the token image would not be displayed

0xmkh: i think we'd need to find a better solution for this in future, since we heavily rely on blockchain api icon urls which we're not using for our `openSend` feature

packages/appkit/src/library/react/index.ts
0xmkh: not needed

packages/controllers/src/utils/AssetUtil.ts
0xmkh: mhmm not sure if it'd be a good idea tbh, since this is only a getter function and we follow same rule for other fn's such as `getChainImage`, `getNetworkImageById`, `getConnectorImage`, `getNetworkImage` and etc

packages/controllers/src/utils/ConstantsUtil.ts
0xmkh: Didn't want to do `SEND_SUPPORTED_CHAINS` since it wouldn't be correct as it implies to the modal view instead of `WalletSend` params view version

packages/scaffold-ui/src/partials/w3m-input-address/index.ts
0xmkh: ah it's not the same ui

packages/scaffold-ui/src/views/w3m-send-confirmed-view/index.ts
0xmkh: yeah they said it's fine to not include the asset there, we can create new UI for that probably

packages/scaffold-ui/src/views/w3m-wallet-send-preview-view/index.ts
0xmkh: nah no need to check which param

packages/scaffold-ui/src/views/w3m-wallet-send-view/index.ts
0xmkh: no we haven't decided to do that, let's definitely discuss about that 🙏

packages/scaffold-ui/src/views/w3m-wallet-send-view/index.ts
```diff
@@ -150,6 +173,72 @@ export class W3mWalletSendView extends LitElement {
...
+        caipAddress: `${caipNetworkId}:${this.address}`,
+        assetAddress,
+        caipNetworkId
+      })
+
+      if (!name || !symbol || !decimals || !balance) {
+        SnackController.showError('Token not found')
+
+        return
+      }
```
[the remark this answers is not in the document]
0xmkh: its ok since we set loading to false in `finally` block

packages/appkit/src/client/appkit-base-client.ts
```diff
@@ -106,13 +107,18 @@ export type Views =
...
+export interface OpenOptions<V extends Views | undefined = Views> {
+  view?: V
   uri?: string
   namespace?: ChainNamespace
-  arguments?: View extends keyof ViewArguments ? ViewArguments[View] : never
+  arguments?: V extends 'Swap'
+    ? ViewArguments['Swap']
+    : V extends 'WalletSend'
+      ? ViewArguments['WalletSend']
+      : never
```
enesozturk, line 122: Normally linter should warn for nested ternary, isn't case for TS maybe? 🤔

packages/appkit/src/client/appkit-base-client.ts
```diff
@@ -204,6 +210,81 @@ export abstract class AppKitBaseClient {
...
+
+    function isSend(options?: OpenOptions): options is OpenOptions<'WalletSend'> {
+      return options?.view === 'WalletSend'
+    }
+
+    return {
+      isSwap,
+      isSend
+    }
+  }
```
enesozturk, line 284: We do we need to create functions if we just check the view names?

packages/appkit/src/client/appkit-base-client.ts
0xmkh: i think for typescript its ok

0xmkh: this was the only way i could solve it using literals, since the current approach we did wasn't correct

packages/controllers/src/utils/BalanceUtil.ts
```diff
@@ -105,5 +122,47 @@ export const BalanceUtil = {
...
+    const publicClient = await ViemUtil.createViemPublicClient(caipNetworkId)
```
enesozturk: Q: cannot we create Viem chain with `defineChain({ ...caipNetwork })` instead of passing caipNetworkId to `createViemPublicClient` and look for the chain object from all the Viem chains?

enesozturk (approved):

packages/appkit/src/client/appkit-base-client.ts
0xmkh: yeah so each of them has return type boolean, e.g `options is OpenOptions<'WalletSend'>`

packages/controllers/src/utils/BalanceUtil.ts
0xmkh: mhmm good question, i'll try that

packages/controllers/src/utils/ViemUtil.ts
```diff
@@ -1,11 +1,39 @@
...
+
+    cachedViemUtils = {
+      chains: viemChains as unknown as Record<string, ViemChain>,
+      createPublicClient,
+      http
+    }
+  }
+
+  return cachedViemUtils
+}
```
tomiir: what's this for? this will still load all of viem chains package

0xmkh: refactored it to use `defineChain` only, and `cachedViemUtils` is used as singleton to store function references from viem

PR #5003, opened 2025-09-11
chore: update error message when insufficient liquidity

enesozturk:
# Description

On SwapController, we are fetching quotes but when there is insufficient liquidity, we are showing insufficient balance error instead, which is misleading

## Type of change

# Associated Issues

For GH issues: closes #...

# Showcase (Optional)

# Checklist

### DIFF
--- a/packages/controllers/src/utils/SwapApiUtil.ts
+++ b/packages/controllers/src/utils/SwapApiUtil.ts
@@ -132,13 +132,22 @@ export const SwapApiUtil = {
   },
 
   async handleSwapError(error: unknown) {
-    const response = await ((error as ErrorOptions)?.cause as Response)?.json()
-    const reason = response?.reasons?.[0]?.description
+    try {
+      const cause = (error as ErrorOptions)?.cause as Response
+      if (!cause?.json) {
+        return undefined
+      }
 
-    if (reason?.includes('insufficient liquidity')) {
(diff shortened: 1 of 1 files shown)


packages/controllers/src/utils/SwapApiUtil.ts
```diff
@@ -129,5 +129,25 @@ export const SwapApiUtil = {
...
+      }
+
+      const response = await cause.json()
+      const reason = response?.reasons?.[0]?.description
+
+      if (reason?.includes('insufficient liquidity')) {
+        return 'Insufficient liquidity'
+      }
+
+      return undefined
```
tomiir, line 148: why not return the reason here?

enesozturk: Bc we don't know the all the possible errors from 1Inch. If the error reason is not the insufficient liquidity, we fall back to previously shown error @tomiir

packages/controllers/src/controllers/SwapController.ts
```diff
@@ -578,8 +578,9 @@ const controller = {
         SwapController.setTransactionDetails()
       }
     } catch (error) {
+      const response = await SwapApiUtil.handleSwapError(error)
```
0xmkh, line 581: is it possible to avoid doing `await` here when handling errors?

PR #5021, opened 2025-09-15
chore: add `SEND_REJECTED` event when user rejects transaction request

0xmkh:
# Description

Fixed an issue where a `SEND_ERROR` event was logged when a user rejected a transaction. It now logs a `SEND_REJECT` event instead

## Type of change

# Associated Issues

For Linear issues: Closes APKT-3802

# Checklist

### DIFF
--- a/.github/workflows/release-publish.yml
+++ b/.github/workflows/release-publish.yml
@@ -55,7 +55,7 @@ jobs:
             --head "${{ github.ref_name }}" \
             --base main \
             --title "Merge ${{ github.ref_name }} back to main" \
-            --body  "Auto‑generated after publishing."
+            --body  "Auto‑generated after publishing." \
             --verbose
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}


packages/scaffold-ui/src/views/w3m-wallet-send-preview-view/index.ts
```diff
@@ -180,23 +181,39 @@ export class W3mWalletSendPreviewView extends LitElement {
...
+      } else {
+        // eslint-disable-next-line no-console
+        console.error('SendController:sendToken - failed to send transaction', error)
+        EventsController.sendEvent({
+          type: 'track',
+          event: 'SEND_ERROR',
+          properties: eventProperties
+        })
+      }
+
```
lukaisailovic: I think this is too much logic in the views, can we extract it to the `SendControlle`?

packages/adapters/wagmi/src/client.ts
```diff
@@ -392,44 +392,60 @@ export class WagmiAdapter extends AdapterBlueprint {
...
-      abi: data.abi,
-      functionName: data.method,
-      args: data.args,
-      __mode: 'prepared'
-    })
+      return { hash: tx }
+    } catch (err) {
+      if (ErrorUtil.isUserRejectedRequestError(err)) {
+        throw new UserRejectedRequestError(err)
+      }
```
lukaisailovic: We've added this logic for every adapter but we're using the same utils and test patterns for error messages. Why not catch the error in the downstream service that calls the provider?

packages/scaffold-ui/src/views/w3m-wallet-send-preview-view/index.ts
```diff
@@ -180,23 +181,39 @@ export class W3mWalletSendPreviewView extends LitElement {
...
+        errMessage = error.message
+      } else {
+        // eslint-disable-next-line no-console
+        console.error('SendController:sendToken - failed to send transaction', error)
+        EventsController.sendEvent({
+          type: 'track',
+          event: 'SEND_ERROR',
+          properties: eventProperties
+        })
+      }
```
lukaisailovic: can we just do the ternary operator on the `event` property in order to avoid entire branch with duplicated `sendEvent`?

packages/adapters/wagmi/src/client.ts
0xmkh: Good idea, we might need to create a service called `WagmiService` that can throw specific errors on provider level. I guess we can do that in next followup PR just to keep this PR short?

lukaisailovic: I didn't mean that we should create `WagmiService`, but that every single provider is called in the Send utility/service/controller (or view in our case) and that instead of implementing this logic for every adapter, we can just keep it there.

Unless we want to reuse `UserRejectedRequestError` for something else, then its fine

0xmkh: ahh i see, ok pushing a change now 🙏

packages/controllers/src/controllers/SendController.ts
```diff
@@ -114,6 +116,18 @@ const controller = {
     state.loading = loading
   },
 
+  getSdkEventProperties(error: unknown) {
```
enesozturk, line 119: There should be similar utility already no?

enesozturk: I remember I've used it recently in several error events

packages/scaffold-ui/src/views/w3m-wallet-send-preview-view/index.ts
```diff
@@ -172,31 +170,28 @@ export class W3mWalletSendPreviewView extends LitElement {
...
-        properties: {
-          message: CoreHelperUtil.parseError(error),
-          isSmartAccount:
-            getPreferredAccountType(ChainController.state.activeChain) ===
-            W3mFrameRpcConstants.ACCOUNT_TYPES.SMART_ACCOUNT,
-          token: this.token?.symbol || '',
-          amount: this.sendTokenAmount,
-          network: ChainController.state.activeCaipNetwork?.caipNetworkId || ''
-        }
+        event: isUserRejectedRequestError ? 'SEND_REJECTED' : 'SEND_ERROR',
```
enesozturk, line 190: Maybe we could apply the same for the swap as well (let's track as another issue/pr)

packages/controllers/src/controllers/SendController.ts
0xmkh: mhmm `CoreHelperUtil.parseError` you mean?

packages/common/src/utils/ErrorUtil.ts
```diff
@@ -42,12 +42,22 @@ export const ErrorUtil = {
       return false
     }
   },
+  isUserRejectedMessage(message: string) {
+    return (
+      message.toLowerCase().includes('user rejected') ||
+      message.toLowerCase().includes('user cancelled') ||
+      message.toLowerCase().includes('user canceled')
```
tomiir, line 49: lol, english

packages/controllers/src/controllers/SendController.ts
tomiir: yea I remembeer there was one utiliity getting the headers at least so we should expand that one

packages/common/src/utils/ErrorUtil.ts
0xmkh: yeah i know i put it there because trust wallet throws `user canceled` instead of `user cancelled` 😭

tomiir (approved):
Looks good but I'd check if the utility already exists 🙏

PR #5189, opened 2025-10-08
fix: wagmi cross-namespace request

tomiir:
# Description

- Fixes issue where `AuthConnector` would target mainnet instead of current network if a request came while the active namespace was different than `eip155`

## Type of change

# Associated Issues

Closes APKT-3997

# Checklist


packages/adapters/wagmi/src/connectors/AuthConnector.ts
```diff
@@ -51,6 +51,16 @@ export function authConnector(parameters: AuthParameters) {
   }
 
   function parseChainId(chainId: string | number) {
+    const networks = ChainController.getCaipNetworks('eip155')
```
enesozturk: Can we use `ConstantsUtil.CHAIN.EVM` (from `common` package)?

enesozturk: Same below

packages/adapters/wagmi/src/connectors/AuthConnector.ts
```diff
@@ -51,6 +51,16 @@ export function authConnector(parameters: AuthParameters) {
...
+    if (!networks.some(network => String(network.id) === String(chainId))) {
```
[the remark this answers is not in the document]
enesozturk: I didn't get why this could be problem, seems ok? Can you give example for the cases where this could cause issues @cursoragent?

packages/adapters/wagmi/src/connectors/AuthConnector.ts
```diff
@@ -51,6 +51,16 @@ export function authConnector(parameters: AuthParameters) {
...
+      const currentChainId = ChainController.getActiveCaipNetwork('eip155')?.id
+      if (currentChainId) {
+        return currentChainId as number
+      }
+
+      return (networks[0]?.id as number) || 1
+    }
+
     return NetworkUtil.parseEvmChainId(chainId) || 1
```
lukaisailovic: Why return mainnet as default? Can we set up a const and return it instead in order to be more explicit here? 
`return NetworkUtil.parseEvmChainId(chainId) || ETH_MAINNET`

tomiir: Default was there all along, but removed it altogether now as it could cause issues on setups without mainnet included

PR #5220, opened 2025-10-13
fix: remove unused client parameter from initializeModal function

Forostovec:
Remove unused client parameter from initializeModal function in sign-client example

The client parameter was never used in the function implementation and was never
passed by any callers. This parameter was misleading as it suggested SignClient
should be passed to AppKit when using manualWCControl, but the architecture
actually requires SignClient and AppKit to work independently. The parameter
was removed to clean up the API and eliminate confusion about the integration
pattern between SignClient and AppKit in manual control mode.


Forostovec:

ignaciosantise:
Thank you for your contribution. At this time, Reown does not generally accept these types of contributions for security reasons.

PR #5222, opened 2025-10-14
Fix: Respect enableWallets overrides in demo config

sky-coderay:
Ensure that the demo configuration keeps wallet support disabled when explicitly set to false instead of forcing it back to true


sky-coderay:

Issue #5365, opened 2025-11-26
[bug] - several multi-chain issues.


glitch-txs:
### Link to minimal reproducible example

[https://lab.reown.com/appkit/?name=multichain-all](<https://lab.reown.com/appkit/?name=multichain-all>)

### Steps to Reproduce

- [X] AppKit requests "switchChain" method after connecting with EVM (this happens only when Solana is already connected) [https://github.com/reown-com/appkit/pull/5372](<https://github.com/reown-com/appkit/pull/5372>)

Steps to repro:

1. Go to Labs
2. Connect to Solana using Trust Wallet
3. Open the modal, go to connections, connect with EVM
4. After connecting the wallet requests to switch to Ethereum

- [X] The `ProfileWallets` view is not typed. We want to use it to show the different connected wallets. [https://github.com/reown-com/appkit/pull/5373](<https://github.com/reown-com/appkit/pull/5373>)
- [X] AppKit doesn't detect when the user disconnects Solana manually from the wallet. [https://github.com/reown-com/appkit/pull/5374](<https://github.com/reown-com/appkit/pull/5374>)
- [ ] When Solana and EVM are active and connected, `allAccounts` from `useAppKitAccount` only shows a single active network. If the user disconnects EVM then allAccounts becomes undefined.

Comment: I see on the source code that `useAppKitAccount` differs from documentation, in the source code it was taken a single-chain approach while the documentation states that [`allAccounts`: A list of connected accounts](<https://docs.reown.com/appkit/react/core/hooks#returns-2>) I won't touch this since addressing it would mean a breaking change.

PR #5372, opened 2025-11-29
fix: Unwanted switchChain requests on connect for EVM

glitch-txs:
# Description

- Added conditionals for chain ID in Wagmi's connect function
- Added `defaultNetwork` as part of the `OptionsController` state
- Rename `getDefaultNetwork` to `getLastActiveOrDefaultNetwork` and added comments.

## Type of change

# Associated Issues

For GH issues: closes https://github.com/reown-com/appkit/issues/5365

# Showcase (Optional)

# Checklist

### DIFF
--- a/.changeset/funny-windows-rest.md
+++ b/.changeset/funny-windows-rest.md
@@ -1,0 +1,30 @@
+---
+'@reown/appkit-scaffold-ui': patch
+'pay-test-exchange': patch
+'@reown/appkit-adapter-bitcoin': patch
+'@reown/appkit-adapter-ethers': patch
+'@reown/appkit-adapter-ethers5': patch
+'@reown/appkit-adapter-solana': patch
+'@reown/appkit-adapter-ton': patch
+'@reown/appkit-adapter-wagmi': patch
+'@reown/appkit': patch
+'@reown/appkit-utils': patch
+'@reown/appkit-cdn': patch
+'@reown/appkit-cli': patch
+'@reown/appkit-codemod': patch
+'@reown/appkit-common': patch
+'@reown/appkit-controllers': patch
+'@reown/appkit-core': patch
+'@reown/appkit-experimental': patch
+'@reown/appkit-pay': patch
+'@reown/appkit-polyfills': patch
+'@reown/appkit-siwe': patch
+'@reown/appkit-siwx': patch
+'@reown/appkit-testing': patch
+'@reown/appkit-ui': patch
+'@reown/appkit-universal-connector': patch
+'@reown/appkit-wallet': patch
+'@reown/appkit-wallet-button': patch
+---
+
+Fixed an issue where profile wallets tabs were shown in the wrong order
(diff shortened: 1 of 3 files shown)


packages/adapters/wagmi/src/client.ts
```diff
@@ -690,6 +690,15 @@ export class WagmiAdapter extends AdapterBlueprint {
         }
       }
 
+      /** The default chainId to connect to, if set by the user as `defaultNetwork` in config options. */
+      let chainId: number | undefined = undefined
+      const defaultNetwork = OptionsController.state.defaultNetwork
+
+      if (defaultNetwork && defaultNetwork.chainNamespace === 'eip155') {
+        const defaultChainId = Number(defaultNetwork?.caipNetworkId.split(':')[1])
+        chainId = Number.isFinite(defaultChainId) ? defaultChainId : undefined
```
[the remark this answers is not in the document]
glitch-txs: The concept of an “active namespace” or “switching namespace” makes no sense in a multi-chain architecture.

Chain switching is a UX friction that exists primarily in EVM chains because multiple EVM chains share the same provider instance. In a multi-chain model, this problem disappears because each connection is handled independently through its own provider, and no implicit switching is required.

For example, if a user is connected to both Solana and Ethereum, both namespaces are active simultaneously. Switching is only necessary when moving between EVM chains that cannot be connected in parallel (e.g., Ethereum → Base), because those chains cannot expose two separate providers in the same session.

However, the SDK currently forces a single-chain workflow by enforcing an “active network” state in the Network modal. This results in the modal triggering an automatic switchChain on connect whenever the wallet’s current chain differs from the previously connected one.

Also, the Network modal previously displayed a “connect wallet” view after selecting a chain, which is no longer the case, if the intention behind the Network modal is to provide a global “selected chain” state to help developers conditionally render UI or adjust logic, then this flow should be brought back and isolate it from the main modal view.

packages/adapters/wagmi/src/client.ts
```diff
@@ -641,7 +641,7 @@ export class WagmiAdapter extends AdapterBlueprint {
     params: AdapterBlueprint.ConnectParams
   ): Promise<AdapterBlueprint.ConnectResult> {
     try {
-      const { id, address, provider, type, info, chainId, socialUri } = params
+      const { id, address, provider, type, info, socialUri } = params
```
0xmkh, line 644: I think we still the `chainId` here, would it be possible to prioritize the `chainId` over `defaultNetwork`?

packages/adapters/wagmi/src/tests/client.test.ts
```diff
@@ -665,6 +665,57 @@ describe('WagmiAdapter', () => {
       expect(result.chainId).toBe(1)
     })
 
+    it('should connect with chainId as undefined when defaultNetwork is not set', async () => {
+      // Mock OptionsController with no defaultNetwork
+      const originalDefaultNetwork = OptionsController.state.defaultNetwork
+      OptionsController.state.defaultNetwork = undefined
```
0xmkh, line 671: we usually use `vi.spyOn` when mocking states, e.g good example is `ChainController.test.ts`

packages/adapters/wagmi/src/client.ts
glitch-txs: this is what was causing the issue, it would force the user to connect to a specific chain every time connect is triggered. This should be optional.

PR #4735, opened 2025-07-29
fix: mark networks as readonly

olehmisar:
# Description

Mark networks as readonly to be able to use a tuple of chains with `as const`

## Type of change

- [ ] Chore (non-breaking change that addresses non-functional tasks, maintenance, or code quality improvements)
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# Associated Issues

No issues, just a bug i found

# Showcase (Optional)

Fixes the following error
image

# Checklist

- [ ] Code in this PR is covered by automated tests (Unit tests, E2E tests)
- [ ] My changes generate no new warnings
- [x] I have reviewed my own code
- [x] I have filled out all required sections
- [ ] I have tested my changes on the preview link
- [ ] Approver of this PR confirms that the changes are tested on the preview link

PR #5454, opened 2026-01-09
fix: mark networks as readonly

olehmisar:
reopen of #4735


olehmisar:
@arein re-opened this as requested

PR #5456, opened 2026-01-09
feat(headless): improve deeplink handling

enesozturk:
## Summary

Refactors the headless `useAppKitWallets()` hook to pre-generate WalletConnect URIs on page load, fixing the iOS deeplink blocking issue that prevented wallet switching.

---

### How iOS/Safari Handles Deeplinks

iOS enforces a strict security model for triggering deeplinks (custom URL schemes like `metamask://`):

- **Allowed**: Deeplinks triggered **synchronously** within a user gesture (tap/click)
- **Blocked**: Deeplinks triggered from **async callbacks** (promises, setTimeout, fetch responses)

This is intentional to prevent websites from hijacking users to other apps without explicit interaction.

### The Original Flow (Broken)

```
User taps wallet → connect() called → connectWalletConnect() starts (async)
                                                ↓
                                    [Network request to fetch WC URI]
                                                ↓
                                    URI ready → trigger deeplink ❌ BLOCKED
```

When user tapped a wallet:
1. `connect()` called `ConnectionController.connectWalletConnect()`
2. This made an **async network request** to generate a WalletConnect URI
3. Once URI was ready, we tried to trigger the deeplink
4. **iOS blocked it** because we were no longer in the original user gesture context

### Why "Decline and Retry" Failed

The problem was especially bad when users declined a connection and tried a different wallet:

1. User taps Wallet A → deeplink triggered (works on first tap if lucky)
2. User declines in wallet, returns to browser
3. User taps Wallet B → `connectWalletConnect()` starts async fetch
4. iOS **immediately blocks** the deeplink attempt
5. User sees "Failed to connect" error

The async nature of URI generation meant **any wallet switch required a new async request**, which iOS would block.

---

### New Flow

```
Page loads → useAppKitWallets() initializes
                    ↓
        connectWalletConnect() called (async, no user gesture needed)
                    ↓
            [WC URI ready and cached]
                    ↓
User taps wallet → URI already exists → trigger deeplink ✅ WORKS
```

#### 1. Pre-generate WC URI on Hook Initialization

```typescript
useEffect(() => {
  if (isHeadlessEnabled && initialized && !hasConnected.current) {
    hasConnected.current = true
    ConnectionController.connectWalletConnect({ cache: 'never' })
  }
}, [isHeadlessEnabled, initialized])
```

When the headless hook initializes, we immediately start generating the WC URI in the background. By the time user taps a wallet, the URI is already available.

#### 2. Auto-refetch When URI is Cleared

```typescript
useEffect(() => {
  if (isHeadlessEnabled && initialized && hasConnected.current && !wcUri && !wcFetchingUri) {
    ConnectionController.connectWalletConnect({ cache: 'never' })
  }
}, [isHeadlessEnabled, initialized, wcUri, wcFetchingUri])
```

If the URI gets cleared (timeout, manual reset, etc.), we automatically fetch a new one so it's ready for the next attempt.

#### 3. Synchronous Deeplink Trigger on Mobile

```typescript
async function connect(_wallet: WalletItem, namespace?: ChainNamespace) {
  // ...
  if (isMobile) {
    const wcWallet = ConnectUtil.mapWalletItemToWcWallet(_wallet)

    if (wcWallet.mobile_link) {
      // URI is already cached - deeplink triggers synchronously
      ConnectionControllerUtil.onConnectMobile(wcWallet)
    } else {
      MobileWalletUtil.handleMobileDeeplinkRedirect(_wallet.id, namespace)
    }
  }
}
```

On mobile, when user taps a wallet, we use the pre-cached URI to trigger the deeplink **synchronously** within the user gesture.

#### 4. Added `mapWalletItemToWcWallet()` Helper

Converts `WalletItem` to `WcWallet` format for the mobile connection utilities:

```typescript
mapWalletItemToWcWallet(wallet: WalletItem): WcWallet {
  return {
    id: wallet.id,
    name: wallet.name,
    image_id: wallet.imageId,
    image_url: wallet.imageUrl,
    mobile_link: wallet.walletInfo.deepLink,
    link_mode: wallet.walletInfo.linkMode ?? null
  }
}
```

---

## Why WalletConnect URIs Can Be Reused

WalletConnect URIs are **not wallet-specific**. The same URI can be used to connect to any wallet. This is what makes pre-generation possible:

1. Generate URI once on page load
2. User can tap MetaMask → URI works
3. User declines, taps Rainbow → same URI still works
4. No need to generate new URIs for each wallet attempt

---

## Test Plan

- [ ] **iOS Safari**: First wallet tap triggers deeplink immediately (no double-tap needed)
- [ ] **Wallet switching**: Declining one wallet and tapping another works on first tap
- [ ] **URI auto-refresh**: After URI timeout, new URI is automatically generated
- [ ] **Desktop**: QR code flow still works as expected
- [ ] **Custom deeplink wallets**: Phantom, Solflare redirect via `handleMobileDeeplinkRedirect`

---

Co-Authored-By: Claude <noreply@anthropic.com>

---

> [!NOTE]
> Enhances headless connect flow with better mobile deeplink handling and clearer state management.
> 
> - Updates `useAppKitWallets()` to set `connectingWallet`, then on mobile trigger deeplink synchronously via `ConnectUtil.mapWalletItemToWcWallet()` and `ConnectionControllerUtil.onConnectMobile`; otherwise falls back to WalletConnect; removes URI subscription logic
> - Adds `resetConnectingWallet()` and exports it; replaces `resetWcUri()` calls in `AppkitConnectDrawer` and integrates in examples
> - Extends `WalletItem` and mapping: add `imageId`, `linkMode`; new `mapWalletItemToWcWallet()` helper
> - Example `ConnectCard`/`WalletConnectQRContent` now pass explicit props (`connectingWallet`, `wcUri`, `isFetchingWcUri`) and relax QR visibility check; keep `resetWcUri` for closing
> - Example `tsconfig.json`: set `jsx` to `preserve`
> - Tests updated to expect `resetConnectingWallet` and new behaviors
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 08b00ac5dbce92366d2c6f2142a0ddc2b1aa7758. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/controllers/tests/hooks/react.test.ts
+++ b/packages/controllers/tests/hooks/react.test.ts
@@ -799,7 +799,8 @@ describe('useAppKitWallets', () => {
       count: 0,
       connect: expect.any(Function),
       fetchWallets: expect.any(Function),
-      resetWcUri: expect.any(Function)
+      resetWcUri: expect.any(Function),
+      resetConnectingWallet: expect.any(Function)
     })
   })
 


apps/laboratory/src/layout/AppkitConnectDrawer.tsx
```diff
@@ -44,13 +44,14 @@ export function AppkitConnectDrawer({ controls }: Props) {
...
-  const { wcUri, isFetchingWcUri, connectingWallet, connect, resetWcUri } = useAppKitWallets()
+  const { wcUri, isFetchingWcUri, connectingWallet, connect, resetConnectingWallet } =
+    useAppKitWallets()
 
   const currentView = history[history.length - 1]
 
   function handleClose() {
     push('connect')
-    resetWcUri()
+    resetConnectingWallet()
```
0xmkh, line 54: `resetConnectingWallet` seems a bit confusing to me, maybe just `reset` would be fine? Wdyt?

tomiir:
@claude review

packages/controllers/exports/react.ts
```diff
@@ -433,50 +459,9 @@ export function useAppKitWallets(): UseAppKitWalletsReturn {
...
-    ) {
-      AlertController.open(
-        ConstantsUtil.REMOTE_FEATURES_ALERTS.HEADLESS_NOT_ENABLED.DEFAULT,
-        'info'
-      )
-    }
-  }, [initialized, isHeadlessEnabled, remoteFeatures?.headless])
+  function resetConnectingWallet() {
+    PublicStateController.set({ connectingWallet: undefined })
+  }
```
tomiir, line 464: Do we really need this method here? what for? shouldn't it reset other things like the selected wallet in routercontroller state?
shouldn't this be part of something like ConnectionController.cancel() or similar?

apps/laboratory/src/layout/AppkitConnectDrawer.tsx
enesozturk: `reset` doesn't give the same context to the consumer, is it? I mean what it resets? `useAppKitWallets` is stateful hook where stores `connectingWallet` object where users can use it to show loading indicator on the wallet items, and `resetConnectingWallet` is to clear it. We are exposing `wcUri` from `useAppKitWallets()`, where users might confuse if `reset` will reset it as well or not

packages/controllers/exports/react.ts
enesozturk: >  what for?
To let users to render loading indicators on the wallet items based on the connecting wallet.

We have different kind of connectors where each working differently and not really possible to handle canceled or error states (like deeplink rejections on iOS), so we need a state from `useAppKitWallets()`, it's `connectingWallet`; Deeplink are restricted if there is async method going on background, so users cannot rely on wcUri use loading state, eventually a state to indicate which wallet is currently connecting is necessary, and a function to clear is necessary

PR #5537, opened 2026-02-12
feat: implement TRON adapter

enesozturk:
# Description

Adds full TRON blockchain support to AppKit, including the adapter, two connectors, network definitions, laboratory integration, an example app, and E2E test infrastructure.

## Architecture

**TronAdapter** (`packages/adapters/tron/src/adapter.ts`):
Extends `AdapterBlueprint` for the `tron` namespace. Handles connect/disconnect, balance fetching (via `BlockchainApiController.getAddressBalance` through the WC RPC proxy), and network switching. Discovers injected TRON wallets at runtime via `TronConnectUtil.watchWalletAdapters`.

**TronConnectConnector** (`packages/adapters/tron/src/connectors/TronConnectConnector.ts`):
Connector for injected TRON wallets (TronLink, MetaMask TRON, Trust Wallet, etc.) using `@tronweb3/tronwallet-abstract-adapter`. Wraps wallet adapter methods into the `TronConnector` interface. Transaction flow: builds unsigned tx via TRON fullnode API → signs with wallet adapter → broadcasts signed tx. Uses chain-ID-to-URL mapping for fullnode resolution to avoid third-party CSP issues from wallet adapters.

**TronWalletConnectConnector** (`packages/adapters/tron/src/connectors/TronWalletConnectConnector.ts`):
Connector for WalletConnect-based connections. Extends `WalletConnectConnector<"tron">` and routes `tron_signMessage` / `tron_signTransaction` through UniversalProvider to the connected wallet.

## Changes

**Core adapter** (`packages/adapters/tron/`):
- `TronAdapter` with connect, disconnect, sign message, send transaction, balance, and sync connections
- `TronConnectConnector` for injected wallets with proper TRON transaction building/signing/broadcasting
- `TronWalletConnectConnector` for WalletConnect sessions
- `TronConnectUtil` for runtime discovery of injected TRON wallet adapters
- `TronAddressUtil` for address validation
- `ProviderEventEmitter` for connector event handling

**Network definitions** (`packages/appkit/src/networks/tron/`):
- TRON Mainnet (`0x2b6653dc`), Shasta Testnet (`0x94a9059e`), Nile Testnet (`0xcd8690dc`)
- All routed through `rpc.walletconnect.org`

**Type system and constants**:
- Add `tron` to `ChainNamespace`, connector IDs, CAIP mappings
- Add `TronConnector` interface and types (`packages/appkit-utils/src/tron/`)
- Add TRON network image IDs to `PresetsUtil` and `AssetUtil`

**Laboratory integration** (`apps/laboratory/`):
- TRON adapter config, sign message test component, send transaction test component
- CSP middleware updated with TronGrid fullnode domains (temporary, pending RPC proxy support)

**E2E test infrastructure**:
- Register `tron` in Playwright test permutation system (Desktop Chrome/Firefox + mobile)
- Add `TRX` to balance currency type, TRON skip guards for EVM-specific tests
- Test only mainnet (TRON wallets do not support programmatic network switching)

**Example app** (`examples/next-tron-app-router/`):
- Next.js App Router example with TRON wallet connection, sign message, and info display

## Known Limitations

- **Balance display**: `getAddressBalance` returns `undefined` from the RPC proxy — pending handler for TRON on blockchain-api (same pattern as TON, just needs the endpoint added)
- **Transaction CSP**: `sendTransaction` calls TronGrid directly for `createtransaction` — will route through `rpc.walletconnect.org` once the proxy supports it
- **Network switching**: TRON wallets do not support programmatic network switching; `switchNetwork` is a no-op

## Type of change

# Checklist

### DIFF
--- a/packages/adapters/tron/src/utils/TronConnectUtil.ts
+++ b/packages/adapters/tron/src/utils/TronConnectUtil.ts
@@ -29,14 +29,6 @@ function isGenuineAdapter(adapter: Adapter): boolean {
     const hasTronStandard = tron?.isTronLink === true
     const hasCompatibleWallet = tronLink?.ready === true
 
-    // eslint-disable-next-line no-console
-    console.log('[TronConnectUtil] isGenuineAdapter check:', {
-      hasTronStandard,
-      hasCompatibleWallet,
-      'tron.isTronLink': tron?.isTronLink,
-      'tronLink.ready': tronLink?.ready
-    })
-
     return hasTronStandard || hasCompatibleWallet
   }
 
@@ -88,8 +80,6 @@ export const TronConnectUtil = {
 
     for (const adapter of adapters) {
       if (adapter.readyState === WalletReadyState.Found) {
-        // eslint-disable-next-line no-console
-        console.log('[TronConnectUtil] Adapter found:', adapter.name || adapter.constructor.name)
         if (isGenuineAdapter(adapter)) {
           onFound(adapter)
         } else {


0xvgd:
Excited to see this PR is in progress. Can you please share the release estimation?

apps/browser-extension/src/core/TronProvider.ts
```diff
@@ -0,0 +1,183 @@
+/* eslint-disable @typescript-eslint/no-require-imports */
+import bs58 from 'bs58'
+import { privateKeyToAccount } from 'viem/accounts'
+
+import { AccountUtil } from '../utils/AccountUtil'
+
+/**
+ * Derive a TRON base58check address from an Ethereum hex address.
+ * TRON addresses = 0x41 + last 20 bytes of keccak256(pubkey), base58check encoded.
+ * Since we already have the ETH address (same key), we just swap the prefix.
```
tomiir, line 10: is this correct??  why do we hacve the eth address?

packages/adapters/tron/exports/index.ts
```diff
@@ -0,0 +1,4 @@
+export * from '../src/adapter.js'
+export * from '@reown/appkit-utils/tron'
+export type { Adapter as TronWalletAdapter } from '@tronweb3/tronwallet-abstract-adapter'
+// NOTE: ReownTronAdapter is not exported - it's for internal testing only
```
tomiir, line 4: is this correct?

packages/adapters/tron/src/connectors/TronConnectConnector.ts
```diff
@@ -0,0 +1,211 @@
...
+    }
+
+    if (method === 'tron_sendTransaction') {
+      const params = args.params as { transaction: Record<string, unknown> }
+
+      return this.adapter.signTransaction(params.transaction) as Promise<T>
+    }
+
+    throw new Error(`Unsupported method: ${method}`)
+  }
```
tomiir, line 76: you can probably do a smart type here where the params type is automatically recognized by the method. If that works then maybe a switch here would be cleaner

packages/adapters/tron/src/connectors/TronConnectConnector.ts
```diff
@@ -0,0 +1,211 @@
...
+    return Promise.resolve()
+  }
+
+  // -- Private ------------------------------------------------------ //
+
+  /**
+   * Resolves the TRON fullnode API URL for building and broadcasting transactions.
+   * Uses the adapter's chainId to pick the correct network, but always maps to our
+   * own controlled URLs (not the adapter's fullNode) to avoid third-party CSP issues.
+   * TODO: Route through rpc.walletconnect.org once the RPC proxy supports TRON.
```
tomiir, line 160: is this still accurate?

packages/adapters/tron/src/connectors/TronConnectConnector.ts
```diff
@@ -0,0 +1,211 @@
...
+    if (typeof adapterWithNetwork.network === 'function') {
+      try {
+        const network = await adapterWithNetwork.network()
+
+        if (network.chainId) {
+          return TRON_FULLNODE_URLS[network.chainId] || DEFAULT_TRON_FULLNODE_URL
+        }
+      } catch {
+        // Adapter may not support network() — fall through to chain mapping
+      }
```
tomiir, line 176: what happens if that's the case?

packages/adapters/tron/src/utils/TronConnectUtil.ts
```diff
@@ -0,0 +1,108 @@
...
+    const hasTronStandard = tron?.isTronLink === true
+    const hasCompatibleWallet = tronLink?.ready === true
+
+    // eslint-disable-next-line no-console
+    console.log('[TronConnectUtil] isGenuineAdapter check:', {
+      hasTronStandard,
+      hasCompatibleWallet,
+      'tron.isTronLink': tron?.isTronLink,
+      'tronLink.ready': tronLink?.ready
+    })
```
tomiir: is this intentional?

packages/adapters/tron/src/utils/TronConnectUtil.ts
```diff
@@ -0,0 +1,108 @@
...
+        isGenuine: isGenuineAdapter(a)
+      }))
+    })
+
+    const cleanups: (() => void)[] = []
+
+    for (const adapter of adapters) {
+      if (adapter.readyState === WalletReadyState.Found) {
+        // eslint-disable-next-line no-console
+        console.log('[TronConnectUtil] Adapter found:', adapter.name || adapter.constructor.name)
```
tomiir: same

packages/adapters/tron/src/adapter.ts
```diff
@@ -0,0 +1,448 @@
...
+    return Promise.resolve({ signature: '' })
+  }
+
+  override async sendTransaction(): Promise<AdapterBlueprint.SendTransactionResult> {
+    return Promise.resolve({ hash: '' })
+  }
+
+  override async writeSolanaTransaction(): Promise<AdapterBlueprint.WriteSolanaTransactionResult> {
+    return Promise.resolve({ hash: '' })
+  }
```
tomiir, line 156: why is this on the blueprint class??

packages/adapters/tron/src/adapter.ts
```diff
@@ -0,0 +1,448 @@
...
+            if (address) {
+              this.addConnection({
+                connectorId: connector.id,
+                accounts: [{ address }],
+                caipNetwork
+              })
+
+              this.listenProviderEvents(connector.id, connector)
+            }
+          }
```
tomiir, line 377: nit, for a separate pr but this does look pretty generic, seems like it could be abstracted to base class

packages/adapters/tron/src/adapter.ts
```diff
@@ -0,0 +1,448 @@
...
+    params: AdapterBlueprint.GetWalletConnectProviderParams
+  ): AdapterBlueprint.GetWalletConnectProviderResult {
+    const walletConnectProvider = new TronWalletConnectConnector({
+      provider: params.provider as UniversalProvider,
+      chains: params.caipNetworks
+    })
+
+    return walletConnectProvider as unknown as UniversalProvider
+  }
+}
```
tomiir, line 448: agh we need to remove all of these from base class at some point 😪

apps/browser-extension/src/core/TronProvider.ts
enesozturk: Yes, this is correct! TRON and Ethereum use the same ECDSA key pair and address derivation (last 20 bytes of keccak256(pubkey)). The only difference is the prefix: Ethereum uses an implicit 0x prefix, while TRON uses 0x41. So we can derive a TRON address from an ETH address by simply changing the prefix byte and re-encoding to base58check. This allows us to use viem's `privateKeyToAccount` for key management.

packages/adapters/tron/exports/index.ts
enesozturk: Yes, this comment is intentional. ReownTronAdapter is only exported from `@reown/appkit-adapter-tron/testing` for internal test use (like in laboratory). It's not part of the public API since it's a mock wallet adapter for testing purposes.

packages/adapters/tron/src/connectors/TronConnectConnector.ts
enesozturk: Good suggestion! I'll refactor this to use a switch statement with better typing in a follow-up commit.

packages/adapters/tron/src/connectors/TronConnectConnector.ts
enesozturk: If the adapter doesn't support `network()`, we fall through to use the first chain from `requestedChains` (line 177). This provides a fallback for wallet adapters that don't expose a network method.

packages/adapters/tron/src/adapter.ts
enesozturk: `writeSolanaTransaction` is on the AdapterBlueprint base class because it's part of the unified adapter interface. Even though TRON doesn't use it, we need to implement it to satisfy the interface contract. We return a stub implementation.

packages/adapters/tron/src/adapter.ts
enesozturk: Agreed! These would be good candidates for abstraction to the base class in a future cleanup PR. The `syncConnections` pattern is very similar across adapters.

enesozturk:
## Review Comments Addressed

I've responded to all review comments and made the following change:

### ✅ Refactored TronConnectConnector.request (commit 430ac9b44)
- Replaced if-else chain with switch statement for better readability
- Destructured params for cleaner code  
- Used block scope for case statements

### Summary of other comments:
- **TronProvider eth address**: Correct - TRON and ETH use same key derivation, just different address prefixes
- **ReownTronAdapter export**: Intentionally not exported - it's for testing only
- **TODO comments**: Still accurate - waiting for WalletConnect RPC proxy TRON support
- **Fallback behavior**: Documented - falls through to requestedChains when adapter doesn't support network()
- **Unsubscribe calls**: Intentional - prevents duplicate listeners on re-initialization
- **Base class methods**: Agreed these could be abstracted in future cleanup PR

Let me know if you need any other changes! 🙏

packages/adapters/tron/src/connectors/TronConnectConnector.ts
enesozturk: The TODO is still accurate. Currently, we're using TronGrid API URLs directly (api.trongrid.io, api.shasta.trongrid.io, nile.trongrid.io) for building and broadcasting transactions. The plan is to route these through rpc.walletconnect.org once the RPC proxy adds TRON support, which would provide better astructure control and consistency with other chains.

Response to reviewer: Yes, this TODO is still accurate. We're waiting for the WalletConnect RPC proxy to add TRON support before we can route through rpc.walletconnect.org instead of using TronGrid URLs directly.

packages/adapters/tron/src/adapter.ts
enesozturk: Good observation! This pattern is indeed common across adapters and could be abstracted to a base class in a future refactoring. The event mapping logic (adapter events → connector events) follows the same pattern, but the validation logic might differ slightly per chain. We can tackle this abstraction in a separate PR to keep this one focused on TRON implementation.

PR #5547, opened 2026-02-17
fix: ensure wallet switching after disconnect works for TON/TRON

enesozturk:
## Summary
Fixes the disconnect issue for TON and TRON adapters when multi-account feature is enabled. After disconnecting an active wallet, the UI now explicitly switches to the first available remaining wallet, ensuring consistent behavior across all adapters.

## Problem
When a user disconnects an active wallet with other wallets connected:
- **EVM, Solana, Bitcoin**: Works correctly - automatically switches to another wallet
- **TON, TRON**: UI doesn't update - the disconnected wallet remains shown as active

## Root Cause
All adapters (Solana, Bitcoin, TON, TRON) call `emitFirstAvailableConnection()` after disconnect, which emits an `accountChanged` event. However, for TON/TRON adapters, this event doesn't consistently trigger the UI state update in the profile wallets view.

## Solution
Added explicit wallet switching logic in `handleDisconnect` that:
1. Captures the list of connections before disconnecting
2. Disconnects the wallet normally
3. If the disconnected wallet was active AND other wallets exist:
   - Filters remaining connections (excluding the disconnected one)
   - Explicitly calls `handleSwitchWallet` with the first remaining wallet
   - Ensures UI reflects the correct active wallet

This approach works for **all adapters** (EVM, Solana, Bitcoin, TON, TRON) and provides consistent behavior.

## Changes
- `packages/scaffold-ui/src/views/w3m-profile-wallets-view/index.ts`: Updated `handleDisconnect` method to explicitly switch wallets after disconnecting an active wallet

## Test Plan
1. Connect multiple TRON wallets (e.g., TronLink + another wallet)
2. Disconnect the active wallet
3. Verify that the UI automatically switches to show the next available wallet
4. Repeat for TON adapter
5. Verify EVM, Solana, and Bitcoin adapters still work correctly

Closes APKT-4457

### DIFF
--- a/packages/scaffold-ui/src/views/w3m-profile-wallets-view/index.ts
+++ b/packages/scaffold-ui/src/views/w3m-profile-wallets-view/index.ts
@@ -566,32 +566,8 @@ export class W3mProfileWalletsView extends LitElement {
 
   private async handleDisconnect(namespace: ChainNamespace, id: string) {
     try {
-      const isActiveWallet = this.activeConnectorIds[namespace] === id
-      const { connections } = ConnectionControllerUtil.getConnectionsData(namespace)
-
       await ConnectionController.disconnect({ id, namespace })
       SnackController.showSuccess('Wallet disconnected')
-
-      /*
-       * Explicitly switch to first available wallet after disconnecting the active wallet.
-       * This is needed because TON/TRON adapters emit 'accountChanged' but the event
-       * doesn't always trigger the UI update properly. By explicitly switching here,
-       * we ensure the UI reflects the new active wallet consistently across all adapters.
-       */
-      if (isActiveWallet && connections.length > 1) {
-        const remainingConnections = connections.filter(
-          conn => !HelpersUtil.isLowerCaseMatch(conn.connectorId, id)
-        )
-
-        if (remainingConnections.length > 0) {
-          const nextConnection = remainingConnections[0]
-          const nextAddress = nextConnection?.accounts[0]?.address
-
-          if (nextConnection && nextAddress) {
-            await this.handleSwitchWallet(nextConnection, nextAddress, namespace)
-          }
-        }
-      }
     } catch {
       SnackController.showError('Failed to disconnect wallet')
     }


enesozturk:
## Root Cause Analysis Update

I've identified and fixed the actual root cause\! The issue is that **TON/TRON connectors don't expose a `provider` property**.

### The Issue Chain:

1. **Connector Type** includes optional `provider`:
```typescript
export type Connector = {
  provider?: Provider | W3mFrameProvider | UniversalProvider
}
```

2. **TON/TRON Connectors** were missing this property ❌

3. **AppKit accountChanged Handler** checks for provider:
```typescript
if (connector?.provider) {  // FALSE for TON/TRON\!
  this.syncProvider({...})
  this.syncConnectedWalletInfo(chainNamespace)  // NEVER CALLED\!
}
```

4. **syncConnectedWalletInfo** updates UI with wallet info - it was never called for TON/TRON\!

### The Fix:

Added `provider` getter to TON/TRON connectors:
```typescript
public get provider(): TronConnector {
  return this
}
```

This ensures the `accountChanged` event handler properly updates the UI when switching wallets after disconnect, just like EVM/Solana/Bitcoin adapters.

packages/scaffold-ui/src/views/w3m-profile-wallets-view/index.ts
```diff
@@ -566,8 +566,32 @@ export class W3mProfileWalletsView extends LitElement {
...
+          conn => !HelpersUtil.isLowerCaseMatch(conn.connectorId, id)
+        )
+
+        if (remainingConnections.length > 0) {
+          const nextConnection = remainingConnections[0]
+          const nextAddress = nextConnection?.accounts[0]?.address
+
+          if (nextConnection && nextAddress) {
+            await this.handleSwitchWallet(nextConnection, nextAddress, namespace)
+          }
```
0xmkh: Ideally we want to avoid switching manually, this should be handled on adapter level via`emit('connections')`

enesozturk:
✅ Agreed and fixed! Removed the manual switching logic in commit d32a92a1a.

Now relying entirely on adapter-level event handling:
- Adapters emit `accountChanged` with proper `provider` property
- AppKit client calls `syncConnectedWalletInfo()` automatically  
- UI updates through the event flow, not manual switching

The `handleDisconnect` method is now back to its simple form - just disconnect and show success message. All wallet switching is handled by the adapter layer via events. 🎯

PR #5548, opened 2026-02-18
Fix/set wcerror

tomiir:
# Description
- Handles WC Errors on mobile connections and sets `wcError` flag

## Type of change

### DIFF
--- a/packages/controllers/tests/controllers/ConnectionController.test.ts
+++ b/packages/controllers/tests/controllers/ConnectionController.test.ts
@@ -324,9 +324,9 @@ describe('ConnectionController', () => {
 
     ConnectionController.state.wcError = false
 
-    await expect(
-      ConnectionController.connectWalletConnect({ cache: 'always' })
-    ).rejects.toThrow('Connection failed')
+    await expect(ConnectionController.connectWalletConnect({ cache: 'always' })).rejects.toThrow(
+      'Connection failed'
+    )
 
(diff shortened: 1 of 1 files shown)


tomiir:
Added tests and a small refactor for the error handling:

**Refactor**: Replaced `.catch()` with `try/catch` block. The `.catch()` approach had a subtle bug — since it returned `undefined`, the promise resolved successfully and execution continued to `state.status = 'connected'`, overwriting the `'disconnected'` set in the catch. With `try/catch`, the success path is naturally skipped on error.

**Tests** (3 new):
- `wcError` and `status = 'disconnected'` when connection rejects in cached mode (Telegram)
- `wcError` and `status = 'disconnected'` when connection rejects with `cache: 'always'`
- Negative test: `wcError` stays `false` and `status = 'connected'` on success

packages/controllers/src/controllers/ConnectionController.ts
```diff
@@ -244,14 +244,19 @@ const controller = {
...
+        ConnectionController.state.status = 'connected'
+      } catch {
+        state.wcError = true
+        state.wcFetchingUri = false
+        state.status = 'disconnected'
+        wcConnectionPromise = undefined
+      }
     } else {
       await ConnectionController._getClient()?.connectWalletConnect?.()
     }
```
0xmkh: do we want to also add a `catch` here as well?

PR #5549, opened 2026-02-18
fix: fix TRON balance fetching and skip identity lookup for non-EVM chains

enesozturk:
## Summary

- **Balance fix**: Switched from the broken JSON-RPC `getAddressBalance` method to the correct REST endpoint `/v1/account/{address}/balance` (same endpoint used by all other chains). Also removes the incorrect manual `÷ 10^6` conversion since `quantity.numeric` is already human-readable.
- **Identity fix**: Added a guard in `syncIdentity` to skip ENS/identity lookup for non-EVM namespaces (`tron`, `solana`, `bip122`, etc.). The `/v1/identity/` endpoint is EVM-only — calling it with TRON addresses produced 3 failing network requests on every page load.

## Root cause

The TRON adapter was using `BlockchainApiController.getAddressBalance()` which posts a JSON-RPC request to `/v1?chainId=...`. The working endpoint is `/v1/account/{address}/balance?chainId=tron:0x2b6653dc&currency=usd&...` which is what `BlockchainApiController.getBalance()` calls.

## Test plan

- [ ] Connect TronLink wallet on TRON mainnet — balance should display correctly
- [ ] Check network requests — no failing `/v1/identity/` requests for TRON addresses
- [ ] Connect EVM wallet — ENS identity lookup should still work normally
- [ ] Connect Solana wallet — no identity requests should be made

### DIFF
--- a/.changeset/quiet-hands-raise.md
+++ b/.changeset/quiet-hands-raise.md
@@ -1,0 +1,6 @@
+---
+'@reown/appkit-controllers': patch
+'@reown/appkit': patch
+---
+
+Fixes issue where mobile connections failing would not set wcError
--- a/packages/controllers/src/controllers/ConnectionController.ts
+++ b/packages/controllers/src/controllers/ConnectionController.ts
@@ -230,30 +230,37 @@ const controller = {
     const isInTelegramOrSafariIos =
       CoreHelperUtil.isTelegram() || (CoreHelperUtil.isSafari() && CoreHelperUtil.isIos())
 
-    if (cache === 'always' || (cache === 'auto' && isInTelegramOrSafariIos)) {
-      if (wcConnectionPromise) {
+    try {
+      if (cache === 'always' || (cache === 'auto' && isInTelegramOrSafariIos)) {
+        if (wcConnectionPromise) {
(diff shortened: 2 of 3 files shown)


packages/appkit/src/client/appkit.ts
```diff
@@ -532,6 +532,11 @@ export class AppKit extends AppKitBaseClient {
   }: Pick<AdapterBlueprint.ConnectResult, 'address' | 'chainId'> & {
     chainNamespace: ChainNamespace
   }) {
+    // Identity/ENS lookup is only supported for EVM chains
+    if (chainNamespace !== ConstantsUtil.CHAIN.EVM) {
```
enesozturk, line 536: Guys I think we had a reason not to do this but I couldn't remember about it - normally identity check only supported for EVM no?

0xmkh: yeah, but maybe we can keep it as is? Not sure if that's going to break something

enesozturk: It's not breaking but if I don't do this it't trying to call the identity for TRON (and probably for other namespaces as well, like Solana, Ton etc)

PR #5616, opened 2026-03-31
fix: filter wallets without wc support from useAppKitWallets search results on mobile

tomiir:
## Problem

When searching wallets in the `useAppKitWallets` hook on mobile, wallets without `supports_wc` (e.g. Rabby) could appear in results. This happened because the search path in `ConnectUtil.getWalletConnectWallets` returned results directly without applying `filterWalletsByWcSupport`, unlike the browse path and the scaffold-ui `w3m-all-wallets-search` component which both apply this filter.

Wallets like Rabby pass the API-level `_filterWalletsByPlatform` check because they have a `webapp_link` (even without `mobile_link`), but should still be excluded by the WC support filter on mobile.

## Fix

Apply the full filtering pipeline to search results in `ConnectUtil.getWalletConnectWallets`:
`markWalletsAsInstalled` → `filterWalletsByWcSupport` → `markWalletsWithDisplayIndex`

This aligns the search path with both:
- The browse path via `WalletUtil.getWalletConnectWallets`
- The scaffold-ui `w3m-all-wallets-search` component

## Changes

- `packages/controllers/src/utils/ConnectUtil.ts` — apply filtering pipeline to search results
- `packages/controllers/tests/utils/ConnectUtil.test.ts` — new tests covering mobile filtering, desktop passthrough, and result ordering

### DIFF
--- a/packages/controllers/tests/utils/WalletUtil.test.ts
+++ b/packages/controllers/tests/utils/WalletUtil.test.ts
@@ -1,0 +1,75 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+
+import {
+  ConnectionController,
+  ConnectorController,
+  OptionsController
+} from '../../exports/index.js'
+import { CoreHelperUtil } from '../../src/utils/CoreHelperUtil.js'
+import type { WcWallet } from '../../src/utils/TypeUtil.js'
+import { WalletUtil } from '../../src/utils/WalletUtil.js'
+
+// -- Helpers ------------------------------------------------------------------
+function createMockWcWallet(overrides: Partial<WcWallet> = {}): WcWallet {
+  return {
+    id: 'wallet-id',
+    name: 'Test Wallet',
+    supports_wc: true,
+    mobile_link: 'testwalletapp://',
+    ...overrides
+  }
+}
+
+// -- Tests --------------------------------------------------------------------
+describe('WalletUtil', () => {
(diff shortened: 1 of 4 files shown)


packages/controllers/tests/utils/WalletUtil.test.ts
```diff
@@ -0,0 +1,75 @@
...
+        createMockWcWallet({ id: 'wc-wallet', supports_wc: true }),
+        createMockWcWallet({ id: 'no-wc-wallet', supports_wc: false })
+      ]
+
+      const result = WalletUtil.filterAndFlagWallets(wallets)
+
+      expect(result.map(w => w.id)).toEqual(['wc-wallet'])
+    })
+
+    it('should pass through all wallets on desktop', () => {
```
enesozturk, line 46: On desktop let's filter out the wallets which `webapp_link` is not valid? When they don't support WC and no webapp_link don't make sense to render them at all

enesozturk (approved):

0xmkh (approved):

packages/controllers/tests/utils/WalletUtil.test.ts
tomiir: hm but on desktop we can have injected wallets on those spots, and clicking on them should trigger those no?

enesozturk (approved):

PR #5618, opened 2026-04-02
fix(send): use string type for sendTokenAmount to preserve precision

Khizr97:
## Summary

Fixes Send/Swap max amount of token throwing transaction errors due to JavaScript floating-point precision loss (REOWN-3617).

### Problem
When sending the max amount of a token with 18 decimals, the transaction fails because the amount exceeds the actual on-chain balance. The sent amount differs from the intended amount due to JS number precision limits.

### Root Cause Analysis
sendTokenAmount was typed as number throughout the send flow (SendController, UI components, event tracking). JavaScript numbers (IEEE 754 double-precision) can only represent ~15-16 significant digits accurately. Tokens with 18 decimals can require up to 18+ significant digits. For example, a balance of 1.123456789012345678 ETH would be truncated/rounded by JS number arithmetic, and the resulting amount could exceed the actual balance by a tiny margin — causing the transaction to revert.

The critical point was w3m-input-token's onMaxClick handler: maxValue.toFixed(20) produced the correct string, but wrapping it in Number(...) immediately lost precision before it reached SendController.

### Approach & Reasoning

**Changed sendTokenAmount from number to string** across the entire flow:

1. **Types** — TxParams, ContractWriteParams, SendControllerState interfaces updated
2. **SendController** — setTokenAmount accepts string, state stores string
3. **w3m-input-token** — @property changed to { type: String }. onMaxClick now passes maxValue.toFixed(20) directly as string (the key fix — no more Number() roundtrip). Input changes wrapped with String(event.detail)
4. **w3m-wallet-send-view** — Balance comparison changed from this.sendTokenAmount > Number(quantity) to NumberUtil.bigNumber(this.sendTokenAmount).gt(quantity) — uses Big.js arbitrary-precision math instead of lossy JS number comparison
5. **w3m-wallet-send-preview-view** — Display values use Number() only for UiHelperUtil.roundNumber (display-only, precision loss acceptable)

**Number() conversion only at boundaries where it's safe:**
- Event tracking amount properties (analytics only — precision loss acceptable)
- Display values via UiHelperUtil.roundNumber (human-readable rounding)
- Solana sendTransaction value (SOL has 9 decimals, fits within JS number safely)

**Critical EVM path preserved full precision:**
- ConnectionController.parseUnits(params.sendTokenAmount.toString(), decimals) — parseUnits takes a string and converts to BigInt, so full precision flows to the contract

**Why string and not BigInt:** The value needs to flow through UI input fields, Lit property bindings, and valtio state — all of which handle strings naturally. BigInt would require constant conversion at every UI boundary.

### Files Changed
- packages/controllers/src/controllers/SendController.ts — types + state
- packages/scaffold-ui/src/partials/w3m-input-token/index.ts — input handling
- packages/scaffold-ui/src/views/w3m-wallet-send-view/index.ts — balance comparison
- packages/scaffold-ui/src/views/w3m-wallet-send-preview-view/index.ts — display
- 5 test files updated to use string literals

### Verification
- 60/60 tests pass across 5 test files (SendController, input-token, send-view, send-preview-view, send-view-params)
- Type check clean on controllers and scaffold-ui packages
- TxParams and ContractWriteParams are internal to SendController — no external consumers affected

## Test plan

- [ ] Send max amount of a token with 18 decimals — transaction should succeed
- [ ] Send partial amounts — verify correct precision
- [ ] Verify balance comparison works (insufficient funds warning)
- [ ] Verify Solana send still works

### DIFF
--- a/packages/controllers/src/controllers/SendController.ts
+++ b/packages/controllers/src/controllers/SendController.ts
@@ -33,7 +33,7 @@ import { SnackController } from './SnackController.js'
 
 export interface TxParams {
   receiverAddress: string
-  sendTokenAmount: number
+  sendTokenAmount: string
   decimals: string
 }
 export interface SendInputArguments {
@@ -47,14 +47,14 @@ export interface SendInputArguments {
 export interface ContractWriteParams {
   receiverAddress: string
   tokenAddress: string
-  sendTokenAmount: number
+  sendTokenAmount: string
   decimals: string
 }
 export interface SendControllerState {
   tokenBalances: Balance[]
   token?: Balance
   hash?: string
-  sendTokenAmount?: number
+  sendTokenAmount?: string
   receiverAddress?: string
   receiverProfileName?: string
   receiverProfileImageUrl?: string
@@ -122,7 +122,7 @@ const controller = {
         getPreferredAccountType(ChainController.state.activeChain) ===
         W3mFrameRpcConstants.ACCOUNT_TYPES.SMART_ACCOUNT,
       token: state.token?.symbol || '',
-      amount: state.sendTokenAmount ?? 0,
+      amount: Number(state.sendTokenAmount ?? '0'),
       network: ChainController.state.activeCaipNetwork?.caipNetworkId || ''
     }
   },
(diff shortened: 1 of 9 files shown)


svenvoskamp (approved):

Khizr97:

PR #5629, opened 2026-04-09
fix: resolve Lit dev warnings from AppKit web components

Khizr97:
## Summary

Fixes 4 Lit development warnings emitted by AppKit internal web components during normal modal usage (REOWN-4559).

### Problem
Four Lit dev warnings were firing in the browser console during normal AppKit modal usage, creating noise for developers integrating AppKit.

### Root Cause Analysis
Each warning had a distinct cause in a different component:

1. **w3m-wallet-send-view** (`change-in-update`): `render()` called `getMessage()` which mutated `this.message` — a `@state()` reactive property. Setting reactive state during render triggers a new update cycle.
2. **w3m-input-address** (`expression-in-textarea`): Two `<textarea>` elements had both `.value=\${...}` binding AND child expression `\${this.value ?? ''}`. Lit doesn't support expressions inside `<textarea>`.
3. **w3m-router-container** (`change-in-update`): `updated()` lifecycle called `onViewChange()` which synchronously set `this.viewDirection` (reactive state), triggering another update during the current cycle.
4. **w3m-all-wallets-list** (`change-in-update`): `render()` called `this.setAttribute('data-mobile-fullscreen', 'true')` — DOM mutation during render.

### Approach & Reasoning

**Fix 1 — Pure function (chosen over computed property):**
Converted `getMessage()` from a state-mutating method to a pure function that returns the message. The return value is passed as a local variable through `render()` and `buttonTemplate(message)`. Removed the `@state() private message` property entirely since it was only set and consumed within the same render cycle. The early-return priority order was carefully reversed to match the original last-assignment-wins semantics: `SELECT_TOKEN > INCORRECT_VALUE > ADD_AMOUNT > INSUFFICIENT_FUNDS > ADD_ADDRESS > INVALID_ADDRESS > PREVIEW_SEND`.

**Fix 2 — Remove child expressions (only option):**
Removed `\${this.value ?? ''}` child text from both textarea elements. The `.value` property binding is the correct and sufficient mechanism for controlling textarea content in Lit. The child text was redundant — it sets `defaultValue` while `.value` sets the current value.

**Fix 3 — queueMicrotask (chosen over setTimeout/requestAnimationFrame):**
Wrapped `this.viewDirection = ...` in `queueMicrotask()`. This defers the state mutation past the current Lit update cycle but runs before the next browser paint, so there's no visual gap in the CSS transition animation. `setTimeout(0)` would run after paint (causing a flash), and `requestAnimationFrame` could miss the current frame.

**Fix 4 — Move to firstUpdated (chosen over willUpdate/updated):**
Moved `setAttribute` to `firstUpdated()` since `mobileFullScreen` is initialized from `OptionsController.state.enableMobileFullScreen` with no subscription — it never changes after init. If it could change dynamically, `updated()` with a `changedProperties` check would be needed instead.

### Verification
- 761 scaffold-ui tests pass (1 skipped)
- All pre-existing type errors only — zero new errors introduced
- getMessage() priority order verified to match original exactly

## Test plan

- [ ] Open AppKit modal in dev mode — no Lit warnings in console
- [ ] Navigate wallet list, send flow, modal views
- [ ] Verify wallet send flow works (address input, amount, preview)
- [ ] Verify modal view transitions animate correctly

### DIFF
--- a/packages/scaffold-ui/src/partials/w3m-all-wallets-list/index.ts
+++ b/packages/scaffold-ui/src/partials/w3m-all-wallets-list/index.ts
@@ -43,6 +43,9 @@ export class W3mAllWalletsList extends LitElement {
   }
 
   public override firstUpdated() {
+    if (this.mobileFullScreen) {
+      this.setAttribute('data-mobile-fullscreen', 'true')
+    }
     this.initialFetch()
     this.createPaginationObserver()
   }
@@ -54,10 +57,6 @@ export class W3mAllWalletsList extends LitElement {
 
   // -- Render -------------------------------------------- //
   public override render() {
-    if (this.mobileFullScreen) {
-      this.setAttribute('data-mobile-fullscreen', 'true')
-    }
-
     return html`
       <wui-grid
         data-scroll=${!this.loading}
--- a/packages/scaffold-ui/src/partials/w3m-input-address/index.ts
+++ b/packages/scaffold-ui/src/partials/w3m-input-address/index.ts
@@ -56,9 +56,7 @@ export class W3mInputAddress extends LitElement {
           ?disabled=${true}
           autocomplete="off"
           .value=${this.value ?? ''}
-        >
-           ${this.value ?? ''}</textarea
-        >
+        ></textarea>
       </wui-flex>`
     }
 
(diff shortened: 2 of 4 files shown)


svenvoskamp (approved):

Khizr97:

PR #5631, opened 2026-04-09
fix: include projectId in API requests when value is empty string

Khizr97:
## Summary

Fixes projectId being silently dropped from api.web3modal.org requests (REOWN-4529).

### Problem
API requests to getAnalyticsConfig and getWallets were missing the projectId query parameter, causing useAppKitWallet to not work with @reown/appkit-wallet-button.

### Root Cause Analysis
FetchUtil.createUrl() used if (value) to filter query params before appending to the URL. This check filters out all falsy values — including empty strings. OptionsController defaults projectId to '' (empty string), so any API call made before OptionsController.setProjectId() runs would have projectId filtered out.

The _getSdkProperties() method reads OptionsController.state.projectId and returns it as a param. When it's still the default empty string, if (value) evaluates to false and the param is silently dropped.

### Approach & Reasoning

**Changed to explicit null/undefined check** — if (value !== undefined && value !== null). This preserves the intended behavior of filtering out undefined params while allowing empty strings through.

**Considered alternatives:**
- Fixing the default value of projectId — would only fix projectId, not other params with the same issue
- Adding a guard in _getSdkProperties() — would be a point fix, not addressing the general FetchUtil problem
- The chosen approach fixes the root cause for all params at once

**Note on empty string params:** This change means params like exclude='' will now be sent as ?exclude= instead of being omitted. Reown's own API should treat empty string params the same as absent params (standard REST behavior). Low risk.

### Verification
- 10/10 FetchUtil tests pass
- 52/52 ApiController tests pass (1 skipped — pre-existing)
- Type check clean

## Test plan

- [ ] Verify getAnalyticsConfig request includes projectId param
- [ ] Verify getWallets request includes projectId param
- [ ] Verify useAppKitWallet works with @reown/appkit-wallet-button

### DIFF
--- a/packages/controllers/src/utils/FetchUtil.ts
+++ b/packages/controllers/src/utils/FetchUtil.ts
@@ -93,7 +93,7 @@ export class FetchUtil {
     const url = new URL(path, this.baseUrl)
     if (params) {
       Object.entries(params).forEach(([key, value]) => {
-        if (value) {
+        if (value !== undefined && value !== null) {
           url.searchParams.append(key, value)
         }
       })
--- a/packages/controllers/tests/utils/FetchUtil.test.ts
+++ b/packages/controllers/tests/utils/FetchUtil.test.ts
@@ -182,6 +182,17 @@ describe('FetchUtil', () => {
       expect(url.toString()).toBe('https://api.example.com/test?key=value&clientId=test-client-id')
     })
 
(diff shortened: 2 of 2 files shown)


svenvoskamp (approved):

Khizr97:

PR #5662, opened 2026-05-11
fix(solana): sync walletProvider after switchNetwork for non-AUTH providers

Khizr97:
## Summary

- Solana adapter now emits `switchNetwork` for WalletConnect and standard wallet providers (previously only AUTH providers triggered this event)
- Base-client `switchNetwork` handler now calls `syncProvider()` after `syncAccount()`, keeping `useAppKitProvider` reactive after network switches

## Root Cause

`useAppKitProvider` reads from `ProviderController.state.providers[namespace]`, which is only updated by `syncProvider()`. The Solana adapter's `switchNetwork` method only emitted the event for AUTH connectors (via `super.switchNetwork()`), leaving WC and standard wallet providers without a re-sync. The base-client handler also never called `syncProvider()` in the `switchNetwork` event path.

## Test plan

- [x] New unit tests: WC provider emits `switchNetwork` event after network change
- [x] New unit tests: Standard wallet provider emits `switchNetwork` event after network change  
- [x] New unit tests: No address → event not emitted
- [x] All 32 Solana adapter tests pass

Fixes: REOWN-4377

### DIFF
--- a/packages/appkit/src/client/appkit-base-client.ts
+++ b/packages/appkit/src/client/appkit-base-client.ts
@@ -1225,6 +1225,19 @@ export abstract class AppKitBaseClient {
       } else {
         this.setUnsupportedNetwork(chainId)
       }
+
+      const currentProvider = ProviderController.getProvider(chainNamespace)
+      const currentProviderId = ProviderController.getProviderId(chainNamespace)
+      const currentConnectorId = ConnectorController.getConnectorId(chainNamespace)
+
+      if (currentProvider && currentProviderId && currentConnectorId) {
+        this.syncProvider({
+          id: currentConnectorId,
+          type: currentProviderId,
+          provider: currentProvider,
+          chainNamespace
+        })
+      }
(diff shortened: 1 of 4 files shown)


enesozturk (approved):

packages/appkit/src/client/appkit-base-client.ts
```diff
@@ -1225,6 +1225,19 @@ export abstract class AppKitBaseClient {
...
+      const currentProviderId = ProviderController.getProviderId(chainNamespace)
+      const currentConnectorId = ConnectorController.getConnectorId(chainNamespace)
+
+      if (currentProvider && currentProviderId && currentConnectorId) {
+        this.syncProvider({
+          id: currentConnectorId,
+          type: currentProviderId,
+          provider: currentProvider,
+          chainNamespace
+        })
```
0xmkh, line 1239: There could be some race condition here

PR #5663, opened 2026-05-11
fix(ethers,ethers5): resolve walletProvider after account switch in modal

Khizr97:
## Summary

- In `connect()`'s early-return path (used when switching accounts in the modal), `connector.provider` was never initialised for connectors backed by `ethersProviders` (injected, coinbaseWallet, baseAccount, safe)
- This caused the base-client's `accountChanged` handler to skip `syncProvider()` — leaving `useAppKitProvider` returning a stale provider
- Fix resolves `connector.provider` from `ethersProviders` before emitting `accountChanged`, when it is not already set
- Applies to both `@reown/appkit-adapter-ethers` and `@reown/appkit-adapter-ethers5`

## Root Cause

`appkit-base-client.ts` only calls `syncProvider()` in the `accountChanged` handler when `connector?.provider` is truthy. The early-return path in `connect()` emits the event before initialising the provider, so the guard is never satisfied.

## Test plan

- [x] New unit test: provider resolved from `ethersProviders` when `connector.provider` is `undefined` in early-return path
- [x] New unit test: existing `connector.provider` is preserved (no re-initialisation) when already set
- [x] All 72 Ethers adapter tests pass
- [x] All 46 Ethers5 adapter tests pass

Fixes: REOWN-4657


packages/adapters/ethers5/src/client.ts
```diff
@@ -454,6 +454,20 @@ export class Ethers5Adapter extends AdapterBlueprint {
...
+           * Resolve the provider before emitting so the base-client's accountChanged
+           * handler can call syncProvider() — keeping useAppKitProvider reactive
+           * when the user switches accounts inside the modal.
+           */
+          if (!connector.provider) {
+            const ethersProvider =
+              this.ethersProviders[connector.id as keyof Omit<ProviderType, 'metadata' | 'EIP6963'>]
+            if (ethersProvider) {
+              await ethersProvider.initialize()
+              connector.provider = (await ethersProvider.getProvider()) as Provider | undefined
```
enesozturk: Isn't the `initialize` returns the provider? If so it could be like:

```suggestion
            if (ethersProvider) {
              connector.provider = await ethersProvider.initialize()
```

enesozturk (approved):

PR #5664, opened 2026-05-12
fix(wallet): bundle @walletconnect/logger to fix SSR named export error

Khizr97:
## Summary

- `@reown/appkit-wallet` previously used `tsc --build` only, which left `@walletconnect/logger` as a raw runtime import in `dist/esm/`
- Vite SSR (SvelteKit, Next.js) externalizes the dependency; Node.js then resolves the CJS version, causing `SyntaxError: Named export 'generateChildLogger' not found` on Vercel and similar SSR environments
- This was a regression introduced in **1.8.10** when `@walletconnect/logger` was bumped from `2.1.2` → `3.0.0` (which added an `exports` map that changed how Vite handles SSR bundling)
- Fix adds a Vite library build step that inlines `@walletconnect/logger` into the dist output — same pattern as `@reown/appkit-cdn`

## Test plan

- [x] `@walletconnect/logger` import no longer present anywhere in `dist/esm/` after build
- [x] Node.js ESM direct import of `dist/esm/exports/index.js` succeeds without CJS/ESM error (simulates Vercel SSR runtime)
- [x] All 27 existing unit tests pass
- [x] Changeset included

Closes REOWN-4396

### DIFF
--- a/packages/wallet/package.json
+++ b/packages/wallet/package.json
@@ -24,7 +24,7 @@
   },
   "scripts": {
     "build:clean": "rm -rf dist",
-    "build": "tsc --build",
+    "build": "tsc --build && vite build",
     "watch": "tsc --watch",
     "typecheck": "tsc --noEmit",
     "lint": "eslint . --ext .js,.jsx,.ts,.tsx",
@@ -49,6 +49,7 @@
   "devDependencies": {
     "@vitest/coverage-v8": "2.1.9",
     "jsdom": "24.1.0",
+    "vite": "5.4.20",
     "vitest": "3.1.3"
   },
   "keywords": [
--- a/packages/wallet/vite.config.ts
+++ b/packages/wallet/vite.config.ts
@@ -1,0 +1,20 @@
+import { resolve } from 'path'
+import { defineConfig } from 'vite'
+
+export default defineConfig({
+  build: {
+    lib: {
+      entry: {
+        'exports/index': resolve(__dirname, 'exports/index.ts'),
+        'exports/utils': resolve(__dirname, 'exports/utils.ts')
+      },
+      formats: ['es'],
+      fileName: (_format, entryName) => `${entryName}.js`
+    },
+    outDir: 'dist/esm',
+    emptyOutDir: true,
+    rollupOptions: {
+      external: ['@reown/appkit-common', '@reown/appkit-polyfills', 'zod']
+    }
+  }
+})
--- a/.changeset/fix-wallet-esm-ssr-bundle.md
+++ b/.changeset/fix-wallet-esm-ssr-bundle.md
@@ -1,0 +1,5 @@
+---
+"@reown/appkit-wallet": patch
+---
+
+Bundle @walletconnect/logger into the wallet package dist to fix SSR environments (SvelteKit, Next.js on Vercel) where Vite externalizes the dependency and Node.js resolves it as CJS, causing a "Named export not found" error at runtime.


packages/wallet/package.json
```diff
@@ -24,7 +24,7 @@
   },
   "scripts": {
     "build:clean": "rm -rf dist",
-    "build": "tsc --build",
+    "build": "tsc --build && vite build",
```
tomiir, line 27: why build with vite herE? tsc is already building, this seems weird

packages/wallet/package.json
```diff
@@ -49,6 +49,7 @@
   "devDependencies": {
     "@vitest/coverage-v8": "2.1.9",
     "jsdom": "24.1.0",
+    "vite": "5.4.20",
```
enesozturk, line 52: Lockfile not updated for this. `pnpm install --frozen-lockfile` will fail in CI.

packages/wallet/vite.config.ts
```diff
@@ -0,0 +1,20 @@
...
+        'exports/index': resolve(__dirname, 'exports/index.ts'),
+        'exports/utils': resolve(__dirname, 'exports/utils.ts')
+      },
+      formats: ['es'],
+      fileName: (_format, entryName) => `${entryName}.js`
+    },
+    outDir: 'dist/esm',
+    emptyOutDir: true,
+    rollupOptions: {
+      external: ['@reown/appkit-common', '@reown/appkit-polyfills', 'zod']
```
enesozturk, line 17: Also add `sourcemap: true, minify: false` to `build`. Lib mode mangles identifiers by default and drops the sourcemaps tsc was emitting.

packages/wallet/vite.config.ts
```diff
@@ -0,0 +1,20 @@
+import { resolve } from 'path'
+import { defineConfig } from 'vite'
+
+export default defineConfig({
+  build: {
+    lib: {
+      entry: {
+        'exports/index': resolve(__dirname, 'exports/index.ts'),
```
enesozturk, line 8: `appkit-utils/src/LoggerUtil.ts` has the same `@walletconnect/logger` named import and still ships unbundled. `@reown/appkit` imports both, so the SSR error survives via that path.

enesozturk (changes requested):
Two blockers, rest is minor.

**1. Lockfile not updated.** `vite` was added to `packages/wallet` devDependencies but `pnpm-lock.yaml` has no entry for it. `pnpm install --frozen-lockfile` in CI will fail.

**2. Fix is incomplete.** `packages/appkit-utils/src/LoggerUtil.ts` does the same named import from `@walletconnect/logger` and `appkit-utils` still builds with plain `tsc`. `@reown/appkit` depends on both packages, so the SSR crash described in the PR still happens through that path. Either apply the same treatment to `appkit-utils`, or fix the root cause in `@walletconnect/logger@3.x` (CJS export shape / `exports` map) so no package has to bundle it. I'd prefer the upstream fix. Bundling a runtime dep per package is a workaround.

Minor:
- Vite lib mode minifies identifiers and emits no sourcemaps by default. Set `minify: false, sourcemap: true` so this package matches the rest of the repo.
- If bundling stays, move `@walletconnect/logger` to devDependencies. Consumers install it for nothing right now.
- `watch` is still `tsc --watch`, so local dev writes unbundled output and never exercises this fix.
- Inlining pino (MIT) into a `SEE LICENSE IN LICENSE.md` artifact drops its license notice. Needs a third-party notices file if this approach lands. Same gap already exists in `packages/cdn`.

PR #5695, opened 2026-06-26
feat: expose headless wallet list imperatively on the AppKit client

enesozturk:
## What

Exposes the **headless wallet list imperatively on the AppKit client**, so a non-React host can list / search / connect wallets without the `useAppKitWallets` React hook.

## Why

The headless wallet flow — list/search/paginate WalletGuide wallets, fetch the WalletConnect URI, and programmatically connect injected / API / mobile-deeplink wallets — lived **only inside the `useAppKitWallets` React hook** (`packages/controllers/exports/react.ts`), reaching into the controllers and re-implementing the connect logic. A non-React consumer (e.g. `@walletconnect/pay-appkit`, which drives a fully custom checkout) has no way to reach it.

This lifts the imperative core onto the client so React and non-React hosts share **one tested code path**.

## Changes

- **`HeadlessWalletUtil`** (new, `@reown/appkit-controllers`) — the framework-neutral core: `fetchWallets` / `getWalletList` / `subscribeWalletList` / `getWalletConnectUri` / `connect` / `resetWcUri` / `resetConnectingWallet`. Reads/writes the global controllers directly (no AppKit instance), exactly like `ConnectionControllerUtil`. The connect logic (injected → API-wallet fallback → mobile deeplink → WalletConnect) is lifted verbatim from the hook.
- **AppKit client methods** (`appkit-base-client.ts`): `appKit.fetchWallets(options?)`, `appKit.getWalletList()`, `appKit.subscribeWalletList(cb)`, `appKit.getWalletConnectUri(options?)`, `appKit.connectWallet(wallet, namespace?, options?)` — each delegating to `HeadlessWalletUtil`.
- `ConnectOptions` / `FetchWalletsOptions` now have a single home in `HeadlessWalletUtil`; the hook re-exports them (resolves the previous duplicate-definition).

`getWalletInfo` / `getApprovedCaipNetworkIds` / `getAccount` / `subscribeAccount` / `subscribeConnections` (which carry `wcUri` / `wcError`) already exist on the client, so no new surface was needed for connected-wallet info or the WC URI state.

## Follow-up (intentionally out of scope)

Refactor `useAppKitWallets` to delegate its `connect` / `fetchWallets` / `getWcUri` bodies to `HeadlessWalletUtil` (dedup the logic). Kept out to keep this diff additive; the hook already shares the option types and is unaffected behaviorally.

## Validation

- `@reown/appkit-controllers` build ✅ · `@reown/appkit` build ✅
[...]

PR #5697, opened 2026-06-29
feat(controllers): expose WalletConnect URI on public state for headless QR

enesozturk:
## What

Exposes the WalletConnect URI on the **public AppKit state** so a headless host can render a QR **without** the `useAppKitWallets` React hook.

`getState()` / `subscribeState()` now include:
- `wcUri: string | undefined`
- `wcError: boolean`
- `wcFetchingUri: boolean`

These are the symmetric **read** for the existing `getWalletConnectUri()` **trigger** added in #5695 — that PR exposed the imperative wallet-list + URI *trigger* on the client, but the only way to *read the generated URI back* was `subscribeConnections`, which is gated behind the `multiWallet` remote feature (no-ops when off) and so can't serve the URI for a single-wallet QR.

## How

A single one-way valtio subscription in `ConnectionController` mirrors `wcUri` / `wcError` / `wcFetchingUri` onto `PublicStateController` whenever any of them changes — so the mirror stays correct regardless of which setter (`setUri`, `setWcError`, `resetUri`, `resetWcConnection`, `connectWalletConnect`, …) mutated it. PublicState never writes back, so there's no feedback loop. This matches the existing precedent where `ConnectionController` already writes `connectingWallet` onto the public state.

## Why

A framework-neutral host (e.g. a plain-JS integration, or `@walletconnect/pay-appkit`) drives everything else through the AppKit **instance** (`getWalletList`, `getAccount`, `getState`, `subscribeState`, …). The WC URI was the one piece with no ungated instance read — forcing consumers to import `@reown/appkit-controllers` directly, which breaks when the package resolves to a different physical instance than the host's `@reown/appkit` (duplicate valtio singleton → the QR reads a copy that's never written). Reading `wcUri` off the public state removes that footgun entirely.

## Tests

- `ConnectionController.test.ts`: asserts the mirror reflects `setUri` / `setWcError` / `resetUri` onto the public state.
- `PublicStateController.test.ts`: updated default/`set()` shape assertions for the new fields.
- `pnpm --filter @reown/appkit-controllers typecheck` clean; controllers tests green; prettier + eslint clean on changed files.

## Follow-up

Once this lands in a canary, `@walletconnect/pay-appkit` will read `appKit.getState().wcUri` / `subscribeState()` instead of the direct `ConnectionController` import.


packages/controllers/src/controllers/PublicStateController.ts
```diff
@@ -38,6 +38,24 @@ export interface PublicStateControllerState {
...
    */
   connectingWallet: WalletItem | undefined
+
+  /**
+   * @description The active WalletConnect pairing URI (for rendering a QR / deeplink), or
+   * `undefined` when none is active. Mirrored from the connection layer so headless hosts can
+   * read it ungated — the symmetric read for `getWalletConnectUri()`.
+   * @type {string | undefined}
+   */
+  wcUri: string | undefined
```
enesozturk: Why do we use PublicStateController here? Like why not ConnectionController or others? Can we double check if this is the best place for this?

packages/controllers/src/controllers/ConnectionController.ts
```diff
@@ -700,3 +700,19 @@ const controller = {
...
+
+/*
+ * Mirror the WalletConnect URI / connection signals onto the public state so headless hosts can
+ * read them ungated via `appKit.getState()` / `appKit.subscribeState()` — the symmetric read for
+ * `getWalletConnectUri()`. (`subscribeConnections` is gated behind the `multiWallet` remote
+ * feature, so it can't serve the URI for a single-wallet QR.) Driving every wcUri/wcError/
+ * wcFetchingUri mutation through one subscription keeps the mirror in sync regardless of which
+ * setter changed it; it's one-way (PublicState never writes back here), so there's no loop.
+ */
+sub(state, () => {
```
enesozturk: Is it ok to call sub here not in a function?

0xmkh (approved):

PR #5706, opened 2026-07-07
fix: honor experimental_preferUniversalLinks on session-request re-opens

ignaciosantise:
## Problem

With `experimental_preferUniversalLinks: true`, AppKit opens a wallet via its **universal link** on connect (pairing), but every subsequent **session request** (signing / confirming a tx) re-opens the wallet via its **native custom scheme** instead, e.g. `rn-web3wallet-internal://wc?requestId=...&sessionTopic=...`.

**Root cause:** at connect time `ConnectionControllerUtil.onConnectMobile` *opens* the universal link but *persists the native `href`* (`safeAppUrl`) as the WalletConnect deeplink choice (`WALLETCONNECT_DEEPLINK_CHOICE`). `@walletconnect/universal-provider` replays that stored `href` on every `request()` (`handleDeeplinkRedirect` builds `${href}wc?requestId=...&sessionTopic=...`). Since the stored `href` is the native base, all request re-opens use the native scheme.

This breaks hosts that embed a dapp in an iOS webview and rely on universal links so they don't have to list every wallet's custom scheme in `LSApplicationQueriesSchemes`. Connect works, but the signing step falls back to a native scheme the host can't open. Reproduced headless (`appKit.connectWallet(...)`), the path WalletConnect Pay uses.

`onConnectMobile` is the single unified path — the headful `w3m-connecting-wc-mobile` partial delegates to it — so one fix covers both headful and headless. No wallet-side or universal-provider change required.

## Fix

- `formatNativeUrl` now also returns `universalHref`, the normalized universal-link base (reuses the same `safeUniversalLink` that already produces `redirectUniversalLink`).
- `onConnectMobile` persists `universalHref` as the deeplink choice when the flag is on **and** a universal link exists, so the stored choice matches what we open with.
- Behavior is unchanged when the flag is off or the wallet has no `link_mode` (native scheme, no regression).

## Tests

- `CoreHelperUtil.test.ts`: `universalHref` equals the normalized universal base, and is `undefined` when no universal link is passed.
- `w3m-connecting-wc-mobile.test.ts`: `setWcLinking` receives the universal base (flag on + `link_mode`), the native base (no `link_mode`), and the native base (flag off).

## Manual verification (recommended)

On mobile / iOS webview: connect a wallet advertising both `mobile_link` and `link_mode`, confirm `localStorage.WALLETCONNECT_DEEPLINK_CHOICE` holds the `https://…/` universal base, then trigger `personal_sign` / `eth_sendTransaction` and confirm the re-open uses the universal link.

### DIFF
--- a/.changeset/prefer-universal-links-deeplink-choice.md
+++ b/.changeset/prefer-universal-links-deeplink-choice.md
@@ -1,6 +1,30 @@
 ---
-'@reown/appkit-controllers': patch
+'@reown/appkit-utils': patch
+'@reown/appkit': patch
+'@reown/appkit-cdn': patch
+'@reown/appkit-cli': patch
+'@reown/appkit-codemod': patch
+'@reown/appkit-common': patch
+'@reown/appkit-core': patch
+'@reown/appkit-experimental': patch
+'@reown/appkit-pay': patch
+'@reown/appkit-polyfills': patch
 '@reown/appkit-scaffold-ui': patch
+'@reown/appkit-siwe': patch
+'@reown/appkit-siwx': patch
+'@reown/appkit-testing': patch
+'@reown/appkit-ui': patch
+'@reown/appkit-universal-connector': patch
+'@reown/appkit-wallet-button': patch
+'@reown/appkit-wallet': patch
+'@reown/appkit-controllers': patch
+'@reown/appkit-adapter-bitcoin': patch
+'@reown/appkit-adapter-ethers': patch
+'@reown/appkit-adapter-ethers5': patch
+'@reown/appkit-adapter-solana': patch
+'@reown/appkit-adapter-ton': patch
+'@reown/appkit-adapter-tron': patch
+'@reown/appkit-adapter-wagmi': patch
 ---
 
 fix: persist the universal-link base as the WalletConnect deeplink choice when `experimental_preferUniversalLinks` is enabled, so session-request re-opens (handled by universal-provider) use the wallet's universal link instead of falling back to its native custom scheme
--- a/.changeset/prefer-universal-links-deeplink-choice.md
+++ b/.changeset/prefer-universal-links-deeplink-choice.md
@@ -1,0 +1,6 @@
+---
+'@reown/appkit-controllers': patch
+'@reown/appkit-scaffold-ui': patch
+---
+
+fix: persist the universal-link base as the WalletConnect deeplink choice when `experimental_preferUniversalLinks` is enabled, so session-request re-opens (handled by universal-provider) use the wallet's universal link instead of falling back to its native custom scheme
--- a/packages/controllers/src/utils/ConnectionControllerUtil.ts
+++ b/packages/controllers/src/utils/ConnectionControllerUtil.ts
@@ -108,21 +108,29 @@ export const ConnectionControllerUtil = {
         ConnectionController.setWcError(false)
         const { mobile_link, link_mode, name } = wallet
         const uriWithPay = CoreHelperUtil.appendPayToUri(wcUri, wcPayUrl)
-        const { redirect, redirectUniversalLink, href } = CoreHelperUtil.formatNativeUrl(
-          mobile_link,
-          uriWithPay,
-          link_mode
-        )
+        const { redirect, redirectUniversalLink, href, universalHref } =
+          CoreHelperUtil.formatNativeUrl(mobile_link, uriWithPay, link_mode)
 
         const deepLink = redirect
         const universalLink = redirectUniversalLink
         const target = CoreHelperUtil.isIframe() ? '_top' : '_self'
 
-        ConnectionController.setWcLinking({ name, href })
+        /*
+         * The href persisted as the WC deeplink choice must match what we open with, so
+         * session-request re-opens (handled by universal-provider) use the same link.
+         */
+        const shouldPreferUniversal =
+          Boolean(OptionsController.state.experimental_preferUniversalLinks) &&
+          Boolean(universalLink)
+
+        ConnectionController.setWcLinking({
+          name,
+          href: shouldPreferUniversal ? (universalHref as string) : href
+        })
         ConnectionController.setRecentWallet(wallet)
 
-        if (OptionsController.state.experimental_preferUniversalLinks && universalLink) {
-          CoreHelperUtil.openHref(universalLink, target)
+        if (shouldPreferUniversal) {
+          CoreHelperUtil.openHref(universalLink as string, target)
         } else {
           CoreHelperUtil.openHref(deepLink, target)
         }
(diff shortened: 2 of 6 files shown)


packages/controllers/src/utils/ConnectionControllerUtil.ts
```diff
@@ -108,21 +108,29 @@ export const ConnectionControllerUtil = {
...
         const universalLink = redirectUniversalLink
         const target = CoreHelperUtil.isIframe() ? '_top' : '_self'
 
-        ConnectionController.setWcLinking({ name, href })
+        /*
+         * The href persisted as the WC deeplink choice must match what we open with, so
+         * session-request re-opens (handled by universal-provider) use the same link.
+         */
+        const shouldPreferUniversal =
+          Boolean(OptionsController.state.experimental_preferUniversalLinks) &&
```
enesozturk: On the client iirc we map the public `experimental_preferUniversalLinks` as `preferUniversalLinks` in state. I mean `experimental_*` prefix only for public facing API, worth double checking

.changeset/prefer-universal-links-deeplink-choice.md
```diff
@@ -0,0 +1,6 @@
+---
+'@reown/appkit-controllers': patch
```
enesozturk: We need to add all of the pckages, not only the ones changed

packages/controllers/src/utils/ConnectionControllerUtil.ts
enesozturk, line 103: Is this function the only place that requires changes really? Might the signature or other request have single source of truth? That's the only part I'm not completely sure, worth double checking

packages/controllers/src/utils/ConnectionControllerUtil.ts
ignaciosantise: this value is taken from the OptionsController, i dont see any accesible flag without the `experimental_`

packages/controllers/src/utils/ConnectionControllerUtil.ts
ignaciosantise: `onConnectMobile` is the single mobile-connect path. Request re-opens are driven entirely by the provider using the persisted WALLETCONNECT_DEEPLINK_CHOICE, which only `onConnectMobile` writes for the mobile universal-link case

.changeset/prefer-universal-links-deeplink-choice.md
ignaciosantise: fixed

packages/controllers/src/utils/CoreHelperUtil.ts
```diff
@@ -191,7 +191,8 @@ export const CoreHelperUtil = {
       redirectUniversalLink: safeUniversalLink
         ? `${safeUniversalLink}wc?uri=${encodedWcUrl}`
         : undefined,
-      href: safeAppUrl
+      href: safeAppUrl,
+      universalHref: safeUniversalLink ?? undefined
```
tomiir, line 195: how is universalHref diferent from redirectUniversalLink? sounds redundant

enesozturk (approved):

packages/controllers/src/utils/CoreHelperUtil.ts
ignaciosantise: `redirectUniversalLink` contains the connection URI (maybe its not the clearest name). `universalHref` is just the app URL

tomiir: im a bit confused, we have `href` , `universalHref` and `redirectUniversalLink` what is each, how do they differ. 

Also, `redirect` and `redirectUniversalLink` are derived from `href` and `universalHref` 
1) they should not be extra return values if they can be derived honestly.
2) do we need a `redirectUniversalHref` as  well?

packages/controllers/src/utils/ConnectionControllerUtil.ts
ignaciosantise: Not changing this — the flag only exists on `OptionsController` state as `experimental_preferUniversalLinks` (interface + default at `OptionsController.ts:144`/`:231`, setter `setPreferUniversalLinks`). There is no non-`experimental_` alias in state, and this read is identical to the line it replaced. So no mapping to reconcile here.

packages/controllers/src/utils/ConnectionControllerUtil.ts
ignaciosantise: Confirmed as the single source of truth, so no other changes needed. `onConnectMobile` is the only mobile-connect entry point — the headful `w3m-connecting-wc-mobile` partial delegates to it and the headless `HeadlessWalletUtil.connect` calls it directly. Session-request re-opens are driven entirely by `@walletconnect/universal-provider` reading the persisted `WALLETCONNECT_DEEPLINK_CHOICE`, which only `onConnectMobile` writes for the mobile universal-link case. Desktop/QR/web paths never pass `link_mode`, so they are out of scope.

.changeset/prefer-universal-links-deeplink-choice.md
ignaciosantise: Done — expanded the changeset to list all publishable `@reown/appkit-*` packages (26 total; `@reown/appkit-adapter-polkadot` stays out per the `ignore` list in `.changeset/config.json`), matching the convention used by prior changesets.

packages/controllers/src/utils/CoreHelperUtil.ts
ignaciosantise: Not redundant — the three serve different roles:

- `href` / `universalHref`: the wallet **base** URLs (native `somewallet://` vs universal `https://wallet.com/`). These are what we persist as the `WALLETCONNECT_DEEPLINK_CHOICE`, because `@walletconnect/universal-provider` appends `wc?requestId=…&sessionTopic=…` to the stored base on every session request.
- `redirect` / `redirectUniversalLink`: those same bases but with the **pairing** `wc?uri=…` already appended — used for the immediate connect-open only.

So it is base-for-persistence vs full-URL-for-connect. This PR only needs `universalHref` (the universal counterpart of the existing `href`) to persist the right base. A `redirectUniversalHref` isn't needed — the base + provider-appended suffix is exactly the mechanism. I left `redirect`/`redirectUniversalLink` as returned values to avoid changing existing callers; deriving them at call sites would be a larger refactor out of scope for this fix.

packages/controllers/src/utils/ConnectionControllerUtil.ts
```diff
@@ -108,21 +108,29 @@ export const ConnectionControllerUtil = {
...
+          Boolean(universalLink)
+
+        ConnectionController.setWcLinking({
+          name,
+          href: shouldPreferUniversal ? (universalHref as string) : href
```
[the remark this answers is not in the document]
ignaciosantise: Good call — applied in 5ff2f976b. Removed both `as string` casts by narrowing inside the selection ternaries (`shouldPreferUniversal && universalLink ? universalLink : deepLink`, same for the persisted href) and collapsed the two `openHref` branches into a single call. Behavior is unchanged; typecheck/eslint/tests pass.

tomiir (approved):

PR #5709, opened 2026-07-10
fix(laboratory): emit missing canary metrics for Verify + Receive Session Proposal alerts

svenvoskamp:
## Problem

Two groups of Grafana canary alerts have been firing continuously — both on **No Data**, not real regressions. Verified against the `monitoring` AWS account (namespace `prod_Canary_Web3Modal`, eu-central-1).

### 1. AppKit Verify (Happy + Unhappy Scam) — 4 alerts
`HappyPath.verify.*` / `UnhappyPath.verify-scam.*` have **0 datapoints over 90 days**. The verify tests in `verify.spec.ts` are `@canary`-tagged with metric annotations, but the file never registered an `afterEach` calling `afterEachCanary()`, so results were never uploaded (only `canary.spec.ts` / `siwe-email.spec.ts` had it).

### 2. AppKit Receive Session Proposal Timing (> 7000ms) — 4 alerts
The alert queries `HappyPath.sign.timing.pairingReceiveSessionProposal`, which also has **0 datapoints** — while `receiveSessionProposal` is healthy (~1.8s, never >7s). `pairingReceiveSessionProposal` was only recorded when `pairingCreatedTime` was set, captured by scraping a WalletConnect `set … core/pairing/pairing` **console log that no longer fires**. So it was never published and the alert fired on No-Data.

## Fixes

1. **`verify.spec.ts`** — add `afterEach` → `afterEachCanary` for the two `@canary` verify fixtures, mirroring `canary.spec.ts`. Non-canary tests sharing the fixtures are skipped by the tag guard.
2. **`w3m-wallet-fixture.ts`** — capture `pairingCreatedTime` deterministically when the wallet starts processing the URI, and push `pairingReceiveSessionProposal` unconditionally. Metric name unchanged, so the existing Grafana alert gets data with no dashboard change.

## Result
Once deployed and the canary runs, all 8 alerts get real data and evaluate genuine pass/fail / latency instead of No-Data. No Grafana provisioning change required.


svenvoskamp:
## Why these alerts fired, and what this PR changes

Both alert groups were firing on **No Data** — not on a real Verify failure or a real latency regression. Verified directly against the `monitoring` AWS account (namespace `prod_Canary_Web3Modal`, eu-central-1).

### Summary

| Alert group | Metric queried | Data in CloudWatch | Real cause | Alert age |
|---|---|---|---|---|
| Verify Happy / Unhappy Scam (4) | `HappyPath.verify.*`, `UnhappyPath.verify-scam.*` | **0 datapoints, ever** (90d, all regions) | Metric never uploaded (missing `afterEachCanary` hook) | New (~1 day) |
| Receive Session Proposal Timing (4) | `HappyPath.sign.timing.pairingReceiveSessionProposal` | **0 datapoints** | Metric stopped emitting (broken console-log dependency) | Old (~15 days) |

Health check for contrast — metrics that *are* emitted and healthy:

| Metric | Status |
|---|---|
| `HappyPath.sign.success` | ✅ ~96/day, healthy |
| `HappyPath.sign.timing.receiveSessionProposal` | ✅ ~1.8s avg, max ~2.7s (never near 7s) |
| `HappyPath.verify.*` / `UnhappyPath.verify-scam.*` | ❌ absent |
| `HappyPath.sign.timing.pairingReceiveSessionProposal` | ❌ absent |

---

### 1. AppKit Verify (Happy + Unhappy Scam)

- **Issue:** the 4 alerts query `HappyPath.verify.*` / `UnhappyPath.verify-scam.*`, which have **zero datapoints ever**. They detect No-Data, not a failing Verify feature.
- **Why it happened:** canary results are only uploaded by an `afterEach` → `afterEachCanary()` hook. `verify.spec.ts` tagged its tests `@canary` and gave them metric names, but never registered that hook (only `canary.spec.ts` / `siwe-email.spec.ts` did). Results were silently discarded.
- **Why it wasn't firing before:** the alert rules are **new**. The metric was always missing, but a No-Data alert can't fire until it exists. These rules were provisioned ~1 day ago (matches recent Grafana provisioning activity) and went straight to Firing.
- **Fix (`verify.spec.ts`):** add `afterEach` → `afterEachCanary` for the two `@canary` verify fixtures, mirroring `canary.spec.ts`. Non-canary tests sharing the fixtures are skipped by the `@canary`-tag guard.

### 2. AppKit Receive Session Proposal Timing (> 7000ms)

- **Issue:** the alert queries `HappyPath.sign.timing.pairing`​`ReceiveSessionProposal`, which returns **0 rows**, while the similarly-named `receiveSessionProposal` is healthy (~1.8s). The 7000ms threshold was never actually being evaluated.
- **Why it happened:** in `w3m-wallet-fixture.ts`, `pairingReceiveSessionProposal` was only recorded when `pairingCreatedTime` was set — and that was captured by **scraping the wallet console** for a WalletConnect `set … core/pairing/pairing` log. That log stopped firing (SDK log change), so `pairingCreatedTime` stayed `null` and the metric was never pushed.

```ts
// before — only recorded if a console log happened to fire
walletPage.page.on('console', msg => {
  if (msg.text().includes('set') && msg.text().includes('core/pairing/pairing')) {
    pairingCreatedTime = new Date()
  }
})
...
if (pairingCreatedTime) {            // never true anymore
  timingRecords.push({ item: 'pairingReceiveSessionProposal', ... })
}
```

- **Why it wasn't firing before:** opposite of the Verify case — the **alert is old and the metric broke**. Before ~15 days ago the log fired, the metric published healthy (< 7000ms) values, and the alert was green. When the log dependency broke, the metric went silent → No-Data → Firing (~15 days, matching the metric's disappearance).
- **Fix (`w3m-wallet-fixture.ts`):** capture `pairingCreatedTime` **deterministically** when the wallet starts processing the URI, and push the metric unconditionally — no SDK log dependency, so it can't silently break again. Metric name unchanged, so the existing Grafana alert gets data with **no dashboard change**.

```ts
// after — deterministic, always recorded
const pairingCreatedTime = new Date()
...
timingRecords.push({
  item: 'pairingReceiveSessionProposal',
  timeMs: proposalReceived.getTime() - pairingCreatedTime.getTime()
})
```

---

### Common thread
Both are No-Data firings from opposite directions: **Verify** = metric never emitted + new alert; **Receive Session Proposal** = metric used to emit but broke + old alert. Neither is a real product/latency issue. After this deploys and the canary runs, all 8 alerts get real data and evaluate genuine pass/fail / latency.

_Caveat: the Verify side is fully confirmed from CloudWatch (0 datapoints vs. healthy comparators). The "broke ~15 days ago" timing for `pairingReceiveSessionProposal` is inferred from the 15-day firing streak + current absence; I didn't pull a 90-day history for that specific metric._

enesozturk (approved):

PR #5710, opened 2026-07-14
fix(testing): deterministic canary session-proposal timing

tomiir:
## Why

The `AppKit Receive Session Proposal Timing > 7000ms` Grafana canary alert paged as a **false positive** (see the burst of P3 escalations on 2026-07-14). Root cause: the queried metric, `HappyPath.sign.timing.pairingReceiveSessionProposal`, was captured by **matching WalletConnect SDK `console.log` strings** on the external wallet page (`core/pairing/pairing`, `resolving attestation`, `verifyContext`). When those internal logs changed format/level the metric silently stopped emitting → the panel went blank → the gap was evaluated as a `>7000ms` breach.

The fix removes the brittle path and replaces it with a deterministic signal.

## What changed

**`packages/testing` (`@reown/appkit-testing`)**
- Add `WalletPage.waitForSessionProposal()` — resolves when the wallet renders the request UI (approve button visible). It reuses the **same owned `session-approve-button` testid** as `performRequestAction` (extracted into `sessionRequestButton()`), so a selector change breaks the approve flow *and* this wait together → the run fails and pages via the success/failure canary, rather than silently dropping just the timing metric.

**`apps/laboratory` canary fixture (`w3m-wallet-fixture.ts`)**
- Delete the `page.on('console')` block (the only console-log-gated timing capture in the suite).
- Replace `receiveSessionProposal` / `pairingReceiveSessionProposal` / `sessionProposalVerification` with two deterministic timings:
  - **`sessionProposalReceived`** — `connectWithUri` resolved → proposal rendered in wallet (the metric the alert should track)
  - **`sessionProposalApproved`** — proposal rendered → session approved (approve round-trip; the boundary split)
- Capture interval timestamps immediately on `await` resolution (no work in between) to avoid skew.

## Required monitoring-repo follow-up

Repoint the Grafana query (`cloudwatch.libsonnet`) from `timing.pairingReceiveSessionProposal` → `timing.sessionProposalReceived`. Until then the panel stays blank (timing alerts already resolve gaps to `OK`, so no false page in the interim). New baseline reads lower — it excludes the approve round-trip — so retune thresholds against real data once it flows.

## Notes / known limitations
- `sessionProposalReceived` includes the test wallet's render time; the pure relay/SDK delivery leg needs a wallet-side signal (owned markers in `react-wallet-v2`) — tracked as a separate follow-up.
- Dropped attestation-latency (`sessionProposalVerification`); Verify correctness is still covered by the pass/fail verify canaries.

## Verification
- [ ] `pnpm build && (cd apps/laboratory && pnpm typecheck)` — not run in the authoring workspace (deps not installed); relying on CI.
- [ ] Canary run: `sessionProposalReceived` + `sessionProposalApproved` present in `timingRecords`; old items absent.

### DIFF
--- a/.changeset/canary-session-proposal-timing.md
+++ b/.changeset/canary-session-proposal-timing.md
@@ -1,0 +1,5 @@
+---
+'@reown/appkit-testing': patch
+---
+
+Add `WalletPage.waitForSessionProposal()` — a deterministic signal (approve-button visibility) for canary session-proposal timing that replaces brittle WalletConnect console-log parsing in the laboratory canary fixture.
--- a/apps/laboratory/tests/shared/fixtures/w3m-wallet-fixture.ts
+++ b/apps/laboratory/tests/shared/fixtures/w3m-wallet-fixture.ts
@@ -32,10 +32,12 @@ export const testConnectedMW = base.extend<ModalWalletFixture>({
     const connectionInitiated = new Date()
     timeEnd('walletPage.connectWithUri')
 
-    // The session proposal is received once the wallet renders the request UI (approve button).
-    // Deterministic, owned-testid signal instead of parsing SDK console logs. Also guards delivery
-    // failures: a proposal that never arrives times out here and fails the run (pages via the
-    // success/failure canary). Interval includes the wallet's render time.
+    /*
+     * The session proposal is received once the wallet renders the request UI (approve button).
+     * Deterministic, owned-testid signal instead of parsing SDK console logs. Also guards delivery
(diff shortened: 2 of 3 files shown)


.changeset/canary-session-proposal-timing.md
```diff
@@ -0,0 +1,5 @@
+---
+'@reown/appkit-testing': patch
```
enesozturk, line 2: Nit: include all packages

enesozturk (approved):

svenvoskamp (approved):

PR #5726, opened 2026-07-24
fix(ci): retry playwright:install-deps on transient apt mirror failures

santgr11:
# Description

The CI runner intermittently can't reach `security.ubuntu.com` while running `pnpm playwright:install-deps` (installing Playwright's system dependencies), which fails the whole shard with 0 tests run - unrelated to any test or app code. Confirmed in real CI logs across two separate runs (different shards each time), with `apt-get` retrying for ~8 minutes before giving up with `Connection failed`.

Wraps both occurrences of the `playwright:install-deps` step in `ui_tests.yml` (the `ui_tests` job and the `merge_reports` job) in a 3-attempt retry loop with a 15s backoff. Still fails the step (not silently) if all 3 attempts fail, so a genuinely broken install isn't masked.

**Caveat:** this hasn't been verified against a live occurrence yet - the flake didn't recur in the two CI runs checked since this landed (every shard succeeded on the first attempt), so the retry path itself is still unexercised. It's a fix for a confirmed, real failure mode, just not yet confirmed to work in practice.

## Type of change

# Associated Issues

For GH issues: closes #...

# Showcase (Optional)

# Checklist

### DIFF
--- a/.github/workflows/ui_tests.yml
+++ b/.github/workflows/ui_tests.yml
@@ -300,7 +300,17 @@ jobs:
       - name: Install Playwright system dependencies
         if: steps.playwright-cache.outputs.cache-hit == 'true' # <-- Re-enable condition
         working-directory: ./apps/laboratory/
-        run: pnpm playwright:install-deps
+        run: |
+          for attempt in 1 2 3; do
+            echo "playwright:install-deps attempt $attempt/3"
+            if pnpm playwright:install-deps; then
+              exit 0
+            fi
+            echo "Attempt $attempt failed (likely a transient apt mirror timeout) - retrying in 15s..."
+            sleep 15
+          done
+          echo "playwright:install-deps failed after 3 attempts"
+          exit 1
 
       - name: Install Playwright Browsers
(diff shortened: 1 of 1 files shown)


.github/workflows/ui_tests.yml
```diff
@@ -300,7 +300,17 @@ jobs:
       - name: Install Playwright system dependencies
         if: steps.playwright-cache.outputs.cache-hit == 'true' # <-- Re-enable condition
         working-directory: ./apps/laboratory/
-        run: pnpm playwright:install-deps
+        run: |
+          for attempt in 1 2 3; do
+            echo "playwright:install-deps attempt $attempt/3"
+            if pnpm playwright:install-deps; then
```
0xmkh, line 306: We used to do this before, but then realised retries can actually bypass certain bugs. 

Would it be possible to retry on test level rather than on playwright level?

santgr11: This retry is for the dep install, not the tests themselves. If this step fails, then there is no test to run

PR #5728, opened 2026-07-29
fix: recover headless WalletConnect connects after a disconnect

enesozturk:
## Problem

On a headless host (AppKit with no scaffold-ui — e.g. WalletConnect Pay's Buyer Experience), connecting a wallet after disconnecting a previous one leaves the page unusable until a reload: taps open nothing, and every subsequent attempt surfaces a connection error. Reported as [WCPBX-1014](https://linear.app/reown/issue/WCPBX-1014/switching-wallets-after-disconnect-leaves-wallet-connection-broken).

Two gaps, both unreachable in headful mode (`w3m-connecting-wc-mobile` only fires `onConnect` once it has a URI, and its "try again" clears the error):

1. **`wcError` is sticky.** It is set in `connectWalletConnect`'s catch, but cleared only by a successful mobile deeplink (`onConnectMobile`) or the headful "try again" button. A headless host that hit one error reads as failed for the rest of the page's life.
2. **A mobile deeplink with no URI fails silently.** `ConnectionControllerUtil.onConnectMobile` is a no-op when `wcUri` is missing — which is exactly the state a disconnect leaves behind (`resetWcConnection` clears the URI). `connectWallet()` resolved as though a wallet had been opened, so the host sat in a connecting state waiting on a redirect that never fires.

## Change

- `ConnectionController.connectWalletConnect` clears `wcError` when a new attempt starts (same thing headful's "try again" does manually).
- New `ConnectionControllerUtil.assertWcUriForDeeplink()`, called by both headless connect paths (`HeadlessWalletUtil.connect` and `useAppKitWallets`' `connect`) before the mobile deeplink branch. The attempt now rejects — which also clears `connectingWallet` via the existing catch — so the host can pre-fetch a URI and retry. It cannot be fetched here: the redirect has to happen inside the tap (iOS Safari blocks one issued after an await).

Deliberately not changed: `resetWcConnection` does **not** clear `wcError`, because `w3m-connecting-wc-view` sets the error and then calls `resetWcConnection` — clearing it there would wipe the headful error UI.

## Tests

- `connectWalletConnect` clears a previous `wcError` on a new attempt.
- `HeadlessWalletUtil.connect` rejects and clears `connectingWallet` when the URI is gone, without calling `onConnectMobile`.
- Existing mobile-deeplink test now sets a URI (it relied on the silent no-op).

`packages/controllers` + `packages/scaffold-ui` suites pass; the 5 failing files on this branch (`w3m-connecting-wc-mobile` universal-link href, `ChainController`, and three localStorage/OTP view files) fail identically on `main`.

### DIFF
--- a/packages/controllers/src/controllers/ConnectionController.ts
+++ b/packages/controllers/src/controllers/ConnectionController.ts
@@ -227,11 +227,6 @@ const controller = {
 
   async connectWalletConnect({ cache = 'auto' }: ConnectWalletConnectParameters = {}) {
     state.wcFetchingUri = true
-    /*
-     * A new attempt starts from a clean error state — `wcError` is otherwise only cleared by a
-     * successful mobile deeplink or the headful "try again" button, so a headless host (no
-     * scaffold-ui) that hit one error stayed in an error state for the rest of the page's life.
-     */
     state.wcError = false
     const isInTelegramOrSafariIos =
       CoreHelperUtil.isTelegram() || (CoreHelperUtil.isSafari() && CoreHelperUtil.isIos())
--- a/packages/controllers/src/controllers/ConnectionController.ts
+++ b/packages/controllers/src/controllers/ConnectionController.ts
@@ -227,6 +227,12 @@ const controller = {
 
   async connectWalletConnect({ cache = 'auto' }: ConnectWalletConnectParameters = {}) {
     state.wcFetchingUri = true
+    /*
+     * A new attempt starts from a clean error state — `wcError` is otherwise only cleared by a
+     * successful mobile deeplink or the headful "try again" button, so a headless host (no
+     * scaffold-ui) that hit one error stayed in an error state for the rest of the page's life.
+     */
+    state.wcError = false
     const isInTelegramOrSafariIos =
       CoreHelperUtil.isTelegram() || (CoreHelperUtil.isSafari() && CoreHelperUtil.isIos())
 
--- a/packages/controllers/src/utils/ConnectionControllerUtil.ts
+++ b/packages/controllers/src/utils/ConnectionControllerUtil.ts
@@ -100,6 +100,25 @@ export const ConnectionControllerUtil = {
       recentConnections: dedupedRecentConnections
     }
   },
+  /**
+   * Throw when there is no WalletConnect pairing URI to deeplink into.
+   *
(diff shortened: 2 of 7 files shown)


packages/controllers/src/controllers/ConnectionController.ts
```diff
@@ -227,6 +227,12 @@ const controller = {
 
   async connectWalletConnect({ cache = 'auto' }: ConnectWalletConnectParameters = {}) {
     state.wcFetchingUri = true
+    /*
+     * A new attempt starts from a clean error state — `wcError` is otherwise only cleared by a
+     * successful mobile deeplink or the headful "try again" button, so a headless host (no
+     * scaffold-ui) that hit one error stayed in an error state for the rest of the page's life.
+     */
```
tomiir: would remove these

packages/controllers/src/utils/ConnectionControllerUtil.ts
```diff
@@ -100,6 +100,25 @@ export const ConnectionControllerUtil = {
...
+   *
+   * {@link ConnectionControllerUtil.onConnectMobile} is a no-op without a URI, and it cannot
+   * fetch one itself: the redirect has to happen inside the tap that triggered it (iOS Safari
+   * blocks a redirect issued after an await). A headless host is the layer that pre-fetches the
+   * URI, so it needs the failure as a rejection — otherwise a connect() call that opened nothing
+   * looks like a connect() call still in flight, and the buyer waits on a wallet that never opens.
+   *
+   * The URI is missing whenever the last one was consumed or cleared — most visibly after a
+   * disconnect (`resetWcConnection`), which is when reconnecting used to silently do nothing.
+   */
```
tomiir: error is already self descriptive imo

tomiir (approved):

0xmkh (approved):

packages/controllers/src/controllers/ConnectionController.ts
enesozturk: Dropped in aa38b39.

packages/controllers/src/utils/ConnectionControllerUtil.ts
enesozturk: Dropped the JSDoc in aa38b39 — kept one line for the part the error message can't say, that not fetching here is deliberate (the redirect has to fire inside the tap).

PR #5736, opened 2026-08-01
fix(siwx): preserve signing context in multichain wallet sessions

0xmkh:
## Summary

- Fixes race condition where active `namespace` changes between message creation and signing
- `DefaultSigner` now calls adapter directly using captured chainId
- Added SIWX test configurations for QA


svenvoskamp:
Approved, do we need to fix this on ethers level as well?

svenvoskamp (approved):

0xmkh:
> Approved, do we need to fix this on ethers level as well?

Nope since we don't set `universalProvider.setDefaultChain("...")` like we do on wagmi

PR #5746, opened 2026-08-12
fix: stop wiping stored connections when a connector registers late

santgr11:
# Description

`useAppKitAccount` could stay permanently disconnected after a reload if the namespace's adapter registers its connectors asynchronously after boot (e.g. an injected wallet extension whose `readyState` resolves later, such as
TronLink). Solana, Bitcoin, and Wagmi adapters share the same async-registration shape, so this affects them too, not just TRON.

Root cause: `syncAdapterConnection` couldn't tell "connector hasn't registered with the adapter yet" apart from "the user disconnected," so it wiped the stored connection (`onDisconnectNamespace`) the moment boot-time
sync ran ahead of the connector showing up. Nothing ever retried once it did.

Fix, entirely in `packages/appkit/src/client/appkit-base-client.ts` (no adapter package touched, so every adapter with this shape benefits, not just TRON):
- `syncAdapterConnection` no longer disconnects when a connector hasn't registered yet. It leaves storage intact and schedules a bounded (10s) cleanup in case the connector genuinely never shows up (e.g. the wallet extension was uninstalled).
- `setConnectors` now retries the namespace sync when a connector matching the stored connector id registers late, while explicitly respecting `enableReconnect: false` and not racing the normal boot-time sync (a real
  gap the whole-branch review caught: every adapter mock in the test suite no-ops `syncConnectors`, so this exact race had no prior test coverage).

## Type of change

- [ x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)

# Associated Issues

Closes REOWN-4797

# Checklist


packages/appkit/src/client/appkit-base-client.ts
```diff
@@ -1460,6 +1491,20 @@ export abstract class AppKitBaseClient {
...
+       * A connector was connected in a previous session but hasn't registered with the
+       * adapter yet (e.g. an injected wallet extension whose readyState resolves
+       * asynchronously). This is not a user disconnect: leave storage untouched so a
+       * late connector registration (see `setConnectors`) can retry this sync. If the
+       * connector never registers, `schedulePendingConnectorCleanup` eventually clears
+       * the stale stored connection instead of leaving it dangling forever.
+       */
+      this.setStatus('disconnected', namespace)
+      this.schedulePendingConnectorCleanup(namespace, connectorId)
+      return
```
0xmkh: I think we should avoid doing this on the appkit base client. I think 99% of connectors should register very quickly when appkit gets initialised. 

Could we maybe look into adding the logic in tron adapter? E.g i've done something very similar on bitcoin adapter for Leather wallet:
https://github.com/reown-com/appkit/blob/77a0dbeb6e59adc15b9fbfdab9fcd9b932e01ddf/packages/adapters/bitcoin/src/connectors/SatsConnectConnector.ts#L141-L150

0xmkh: Maybe we can do something like this? Just an example

```ts
async syncConnectors() {
  const chains = ChainController.getCaipNetworks()

  const readyAdapters = await TronConnectUtil.waitForAdapters(this.walletAdapters)

  for (const adapter of readyAdapters) {
    this.addConnector(new TronConnectConnector({ adapter, chains }))
  }
}
```

santgr11: I moved the sync fix from the base client to tron adapter

packages/adapters/tron/src/utils/TronConnectUtil.ts
```diff
@@ -50,7 +50,60 @@ function listenForReady(
...
+          if (state !== WalletReadyState.Loading) {
+            settle()
+          }
+        }
+        adapter.on('readyStateChanged', handler)
+
+        return { adapter, handler }
+      })
+
+      const timer = setTimeout(finish, timeoutMs)
```
0xmkh: Do we know if every tron wallet emits the `readyStateChanged` event?

Instead of a fixed 3-second wait, could we implement a polling interval that checks every ~200ms? This way users wouldn't have to wait the full duration if the wallet doesn't respond.

We might be able to reuse this existing function:

https://github.com/reown-com/appkit/blob/d6cb98d47eccd976ef47b016264708399bb0a175/packages/adapters/bitcoin/src/connectors/SatsConnectConnector.ts#L146-L150

santgr11: done

0xmkh (approved):

PR #5747, opened 2026-08-13
Fix image for broken image urls in token list

santgr11:
# Description

Tokens with a dead or unreachable `iconUrl` were rendering the browser's native broken image icon instead of falling back to the placeholder icon. This only affected the "URL present but fails to load" case; a missing/empty `iconUrl` already fell back correctly.

`wui-list-token` renders a raw `wui-image` whenever `tokenImageUrl` is set, but never reacted to the image actually failing to load. Other composites in the same package (`wui-token-list-item`, `wui-transaction-visual`) already listen for `wui-image`'s `onLoadError` event and swap to a placeholder icon; `wui-list-token` was the one place missing that handling.

Added an `imageError` state to `wui-list-token` that gets set on `onLoadError` and falls back to the existing `coinPlaceholder` icon, matching the pattern already used elsewhere. The state resets whenever `tokenImageUrl` changes so a recycled list item does not get stuck showing the placeholder for a different token whose icon loads fine.

## Type of change

# Associated Issues

Closes REOWN-4538

# Showcase (Optional)

2026-08-13 Token with broken image url

# Checklist

- [ ] Code in this PR is covered by automated tests (Unit tests

### DIFF
--- a/packages/ui/src/composites/wui-list-token/index.ts
+++ b/packages/ui/src/composites/wui-list-token/index.ts
@@ -1,5 +1,5 @@
 import { LitElement, html } from 'lit'
-import { property } from 'lit/decorators.js'
+import { property, state } from 'lit/decorators.js'
 
 import { NumberUtil } from '@reown/appkit-common'
 
@@ -27,6 +27,8 @@ export class WuiListToken extends LitElement {
   @property() public tokenCurrency = ''
 
   @property({ type: Boolean }) public clickable = false
+
+  @state() private imageError = false
 
   // -- Render -------------------------------------------- //
   public override render() {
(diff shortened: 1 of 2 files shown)


packages/ui/src/composites/wui-list-token/index.ts
```diff
@@ -67,13 +69,27 @@ export class WuiListToken extends LitElement {
...
+  }
+
   public visualTemplate() {
-    if (this.tokenName && this.tokenImageUrl) {
-      return html`<wui-image alt=${this.tokenName} src=${this.tokenImageUrl}></wui-image>`
+    if (this.tokenName && this.tokenImageUrl && !this.imageError) {
+      return html`<wui-image
+        alt=${this.tokenName}
+        src=${this.tokenImageUrl}
+        @onLoadError=${this.handleImageError}
```
0xmkh, line 83: Can we add `.bind(this)`?

```ts
@onLoadError=${this.handleImageError.bind(this)}
```

0xmkh: Or arrow fn

santgr11: .bind(this) isn't needed here, [Lit automatically binds declarative @event listeners to the component instance](https://lit.dev/docs/components/events/#understanding-this-in-event-listeners).
Not using bind also matches the existing pattern in this package, like in packages/ui/src/wui-token-list-item/index.ts:108 `(@onLoadError=${this.imageLoadError})` and packages/ui/src/wui-image/index.ts:72 `(@error=${this.handleImageError})`

0xmkh: Yeah i think it's still nice to add it in case people want to add more code to the fn later on, e.g it's hard to catch bugs so having it in case would be nice, but i can merge PR as is

0xmkh (approved):

PR #5336, opened 2025-11-10
chore: lazy load coinbase

tomiir:
# Description

- **Introduces shared provider abstractions** under `@reown/appkit-utils/ethers` and migrates adapters to use them.
- **Lazy-loads Coinbase Base Account SDK** so it only initializes on user connection, avoiding eager network calls and improving startup performance.
- **Unifies Safe provider handling** via a shared wrapper that normalizes `eth_requestAccounts`.
- **Aligns ethers and ethers5 adapters** to the same provider wiring and initialization semantics.

Reference: [`fix/lazy-load-coinbase` comparison](https://github.com/reown-com/appkit/compare/fix/lazy-load-coinbase?expand=1)

### Motivation
- Eager initialization of the Coinbase Base Account SDK triggers unnecessary network activity and delays cold starts.
- Safe provider behavior differs from standard EIP-1193 wallets (no `eth_requestAccounts`), leading to adapter-specific workarounds.
- Centralizing provider wrappers reduces duplication, improves ergonomics, and makes future maintenance/testing easier.

### What changed
- **appkit-utils**
  - Added provider wrappers:
    - `EthersProvider` base class (with `initialize()` and `getProvider()`).
    - `InjectedProvider` for `window.ethereum`.
    - `BaseProvider` (Coinbase Base Account) with deferred init.
    - `SafeProvider` with `eth_requestAccounts` normalization.
  - Exported new wrappers via `@reown/appkit-utils/ethers`.
  - Updated `EthersTypesUtil` to include wrapper-driven `ProviderType`.
- **adapters/ethers**
  - Switched to provider wrappers.
  - Do not initialize Coinbase provider at boot; initialize on connection.
  - Initialize Safe provider eagerly only when inside a Safe App.
  - When creating connectors, call `provider.getProvider()`; in `connect()`, initialize wrapper on-demand.
- **adapters/ethers5**
  - Mirrored the same provider abstraction and lazy-loading pattern used in `ethers`.
  - Updated imports order to align with shared exports.
- **Tests and housekeeping**
  - Removed adapter-local `SafeProvider.ts` in `ethers` adapter (now shared).
  - Adjusted tests where provider construction/init semantics changed.
  - Updated workspace lockfile and package exports.

### Behavioral details
- **Coinbase/Base Account**
  - No SDK initialization until user selects Coinbase/Base connector.
  - On connect, the wrapper initializes and returns an EIP-1193-compatible provider.
- **Injected/EIP-6963**
[...]

PR #5750, opened 2026-08-18
fix(ethers,ethers5): restore Coinbase Wallet / Base Account session on page reload

### LINKED ISSUE #5749: [bug] ethers/ethers5: Coinbase Smart Wallet and Base Account disconnect on page refresh (wagmi is fine)

https://appkit-lab.reown.com/appkit?name=ethers

### Steps to Reproduce

I also tested with the latest published version and the bug is still there.

Do this in a browser WITHOUT the Coinbase Wallet extension installed (or with it disabled).

1. Open https://appkit-lab.reown.com/appkit?name=ethers
2. Click "Connect" and choose "Coinbase Wallet"
3. A popup opens on keys.coinbase.com. Sign in with passkey and approve.
4. Wallet is connected, address is shown in UI.
5. Refresh the page.
6. Wallet is disconnected.

https://github.com/user-attachments/assets/9e7d08ff-565b-4df9-82af-ab3ef06977ab

### Summary

Expected: after connecting with Coinbase Smart Wallet or Base Account, a page refresh should keep the wallet connected, the same way it does for MetaMask.

Actual: on refresh the wallet is disconnected.

Only the ethers and ethers5 adapters are affected. wagmi works fine. The Coinbase browser extension also works fine. It looks like this started in 1.8.15 (after the change that lazy loads the Coinbase SDK, PR #5336) and is still there in the latest version. 

I have looked into the cause and have a small PR ready for ethers and ethers5 with tests. I will link it here!

```
    "@reown/appkit": "^1.8.23",
    "@reown/appkit-adapter-ethers": "^1.8.23",
```

### Node.js Version

v20.19.6

### Package Manager

yarn@1.22.19

kamal9494:
# Description

Fixes Coinbase Smart Wallet / Base Account sessions being dropped on page reload with the ethers and ethers5 adapters.

### Root cause

Regression from #5336 (lazy load coinbase, shipped in 1.8.15).

- `createEthersConfig()` creates `CoinbaseWalletProvider` and `BaseProvider` but intentionally does not call `initialize()`, so the SDK is only loaded when needed.
- `syncConnectors()` then registers each connector with `provider: await provider.getProvider()`, which is `undefined` for those two because nothing has built the SDK yet.
- `connect()` was updated to call `initialize()` before use, so the first connection works.
- The reload path was not. On page load the core calls `syncConnection()`, which reads `connector.provider` directly:

```ts
const selectedProvider = connector?.provider as Provider // undefined for coinbaseWallet / baseAccount

if (!selectedProvider) {
  throw new Error('Provider not found')
}
```

- The core catches that error and calls `onDisconnectNamespace()`, so the user is silently logged out. The SDK is never created on that page load.

### Fix

Keep the lazy load, but treat a returning user as the one exception. In `createEthersConfig()`, right where each lazy provider is created:

```ts
this.ethersProviders.coinbaseWallet = new CoinbaseWalletProvider()
if (this.hasStoredConnection(CommonConstantsUtil.CONNECTOR_ID.COINBASE)) {
  await this.ethersProviders.coinbaseWallet.initialize()
}
```

`hasStoredConnection()` is a small private helper around `HelpersUtil.getConnectorStorageInfo()`, the same check `syncConnections()` already uses (`hasConnected && !hasDisconnected`).

### Tests

Four cases added to the existing `createEthersConfig` suite in each adapter:

- previously connected with coinbase: `coinbaseWallet.initialize()` called once, base not touched
- previously connected with base account: the mirror
- never connected: neither initialized
- explicitly disconnected: neither initialized

The first two fail on the current `main` and pass with this change.

## Type of change

# Associated Issues

closes #5749

# Checklist

### DIFF
--- a/.changeset/wide-plums-yell.md
+++ b/.changeset/wide-plums-yell.md
@@ -1,0 +1,30 @@
+---
+'@reown/appkit-adapter-ethers5': patch
+'@reown/appkit-adapter-ethers': patch
+'@reown/appkit': patch
+'@reown/appkit-adapter-bitcoin': patch
+'@reown/appkit-adapter-solana': patch
+'@reown/appkit-adapter-ton': patch
+'@reown/appkit-adapter-tron': patch
+'@reown/appkit-adapter-wagmi': patch
+'@reown/appkit-utils': patch
+'@reown/appkit-cdn': patch
+'@reown/appkit-cli': patch
+'@reown/appkit-codemod': patch
+'@reown/appkit-common': patch
+'@reown/appkit-controllers': patch
+'@reown/appkit-core': patch
+'@reown/appkit-experimental': patch
+'@reown/appkit-pay': patch
+'@reown/appkit-polyfills': patch
+'@reown/appkit-scaffold-ui': patch
+'@reown/appkit-siwe': patch
+'@reown/appkit-siwx': patch
+'@reown/appkit-testing': patch
+'@reown/appkit-ui': patch
+'@reown/appkit-universal-connector': patch
+'@reown/appkit-wallet': patch
+'@reown/appkit-wallet-button': patch
+---
+
+Fix Coinbase Wallet and Base Account not reconnecting on page reload with ethers and ethers5 adapters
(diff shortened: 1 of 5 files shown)


kamal9494:

PR #5755, opened 2026-08-31
fix(ui): enlarge default wallet icon fallback in visual thumbnail

santgr11:
# Description

Make default wallet icon bigget in Sign In message request

## Type of change

# Associated Issues

Closes FS-153

# Showcase (Optional)

## Before
image

## After
Screenshot_5

# Checklist

### DIFF
--- a/packages/ui/src/composites/wui-visual-thumbnail/styles.ts
+++ b/packages/ui/src/composites/wui-visual-thumbnail/styles.ts
@@ -17,7 +17,6 @@ export default css`
   }
 
   wui-icon {
-    width: 32px;
-    height: 32px;
+    font-size: 48px;
   }
 `
--- a/.changeset/chubby-grapes-retire.md
+++ b/.changeset/chubby-grapes-retire.md
@@ -1,0 +1,32 @@
+---
+'@reown/appkit-ui': patch
+'pay-test-exchange': patch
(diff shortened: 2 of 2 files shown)


packages/ui/src/composites/wui-visual-thumbnail/styles.ts
```diff
@@ -17,7 +17,6 @@ export default css`
   }
 
   wui-icon {
-    width: 32px;
-    height: 32px;
+    font-size: 48px;
```
0xmkh, line 20: Can we use `width` and `height` instead like we did before just increase it?

santgr11: The icon sits two shadow roots deep (wui-icon → ph-wallet → the actual <svg>), and Phosphor draws that SVG with width="1em" height="1em", sized off its own font-size, not off any container's width. On top of that, ph-wallet itself uses display: contents, so it has no box at all to apply a width to. That's why setting width/height from outside never reaches it, no matter the value. font-size works because, unlike width, it's inherited: it passes down through shadow boundaries automatically, so setting it on wui-icon flows all the way to the SVG's 1em and actually resizes it, the same way color already does for its fill.

0xmkh (approved):

PR #5758, opened 2026-09-01
fix(ui): keep sign modal buttons inside the modal boundary

santgr11:
# Description

On the Sign modal (SIWE/SIWX "sign this message" screen), the Cancel and Sign buttons could
render outside the modal's boundary when both showed their longer loading labels
("Cancelling..." / "Signing..."), reachable when a sign flow is disrupted by a
network problem, since neither button is disabled while the other is loading.

Root cause: `wui-button`'s CSS used `white-space: nowrap` with no way for the button or its
label to shrink, so once the combined content (label + spinner) needed more width than the
button's flex-shrunk share of the row, it spilled outside the button and outside the modal
instead of adapting.

Fix (`packages/ui/src/composites/wui-button/styles.ts`, CSS-only, no template/behavior changes):

- `:host { min-width: 0; }`: lets `wui-button` shrink below its natural content width as a flex
  item, instead of refusing to shrink at all.
- `wui-text { min-width: 0; overflow: hidden; text-overflow: ellipsis; }`: truncates the label
  with `…` once the button is compressed, instead of overflowing.
- `wui-loading-spinner, slot[name='iconLeft'], slot[name='iconRight'] { flex-shrink: 0; }`:
  keeps the spinner/icons from being squeezed or distorted; only the text truncates.

This guarantees both buttons stay within the modal boundary regardless of label length or how
many buttons are simultaneously in a loading state. Normal short labels ("Cancel"/"Sign") are
unaffected: no visual change when there's enough room, since these rules only engage once space
is genuinely insufficient. See `.claude/docs/fs-160-sign-modal-buttons-out-of-bounds/findings.md`
for the full investigation, including why an initial `wui-text`-only fix (matching the pattern
already used in `wui-cta-button`) was insufficient here and the `:host` change was also required.

## Type of change

# Associated Issues

Closes FS-160

# Showcase (Optional)

## Before
image

## After
2026-09-01 SignIn and Cancel

# Checklist

### DIFF
--- a/packages/ui/src/composites/wui-button/styles.ts
+++ b/packages/ui/src/composites/wui-button/styles.ts
@@ -3,6 +3,7 @@ import { css } from '../../utils/ThemeHelperUtil.js'
 export default css`
   :host {
     width: var(--local-width);
+    min-width: 0;
   }
 
   button {
@@ -17,6 +18,18 @@ export default css`
         ${({ easings }) => easings['ease-out-power-1']};
     will-change: scale, background-color, border-radius;
     cursor: pointer;
+  }
+
+  wui-text {
+    min-width: 0;
+    overflow: hidden;
+    text-overflow: ellipsis;
+  }
+
+  wui-loading-spinner,
+  slot[name='iconLeft'],
+  slot[name='iconRight'] {
+    flex-shrink: 0;
   }
 
   /* -- Sizes --------------------------------------------------- */
--- a/.changeset/open-hairs-fly.md
+++ b/.changeset/open-hairs-fly.md
@@ -1,0 +1,32 @@
+---
+'@reown/appkit-ui': patch
+'pay-test-exchange': patch
+'@reown/appkit-adapter-bitcoin': patch
+'@reown/appkit-adapter-ethers': patch
+'@reown/appkit-adapter-ethers5': patch
+'@reown/appkit-adapter-solana': patch
+'@reown/appkit-adapter-stellar': patch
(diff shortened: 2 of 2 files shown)


packages/ui/src/composites/wui-button/styles.ts
```diff
@@ -3,6 +3,7 @@ import { css } from '../../utils/ThemeHelperUtil.js'
 export default css`
   :host {
     width: var(--local-width);
+    min-width: 0;
```
0xmkh, line 6: Can we try editing the styles on modal/components level? I think this change would affect all buttons everywhere in the modal

santgr11: This CSS only has an effect when wui-button is a flex item and gets squeezed narrower than its label, since wui-button's host has no display set, so it falls back to display: inline (where min-width and text-overflow don't apply) everywhere except inside a flex row. For every other button in the modal, including non-fullWidth ones and fullWidth ones with enough room, this is a no-op. The only case where it changes anything is exactly the one we're fixing: a fullWidth button squeezed below its content width, causing overflow. So it just turns "overflow past the modal" into "shrink and ellipsis". I also checked, and w3m-swap-preview-view has the same two-fullWidth-buttons-in-a-row setup, so this probably fixes the same latent bug there too. I can still scope it to the view if you'd prefer, but I wanted to flag that it's effectively already scoped by how the CSS resolves.

0xmkh (approved):

PR #5760, opened 2026-09-02
fix(scaffold-ui): show SIWX sign-message prompt after connecting without a refresh

santgr11:
# Description

With SIWX configured on the Solana adapter, signing in by email did not show the SIWXSignMessage popup after the OTP was verified. The modal just closed, and the popup only appeared after a manual page refresh.

Root cause: three connect flows (w3m-email-verify-otp-view, w3m-connecting-farcaster-view, w3m-connecting-wc-browser) closed the modal unconditionally right after connecting, racing the address-change subscription that opens the SIWXSignMessage view. The close was synchronous, and the open was async, so the close consistently won and discarded the pending sign-message screen.

Fix: each of the three now awaits SIWXUtil.isAuthenticated() before closing, mirroring the guard w3m-network-switch-view already uses for the same problem after a network switch. The modal only closes once a session already exists (or SIWX is not configured); otherwise, it defers to the SIWX flow to open the sign-message prompt.

## Type of change

# Associated Issues

Closes FS-166

# Showcase (Optional)

## Before
2026-09-02 Sign with email needs refresh

## After
2026-09-02 Sign with email working ok

# Checklist

### DIFF
--- a/packages/scaffold-ui/src/partials/w3m-connecting-wc-browser/index.ts
+++ b/packages/scaffold-ui/src/partials/w3m-connecting-wc-browser/index.ts
@@ -6,7 +6,8 @@ import {
   ConnectorController,
   EventsController,
   ModalController,
-  RouterController
+  RouterController,
+  SIWXUtil
 } from '@reown/appkit-controllers'
 import { customElement } from '@reown/appkit-ui'
 
@@ -53,7 +54,9 @@ export class W3mConnectingWcBrowser extends W3mConnectingWidget {
         throw new Error('w3m-connecting-wc-browser: No connector found')
       }
 
-      ModalController.close()
+      if (await SIWXUtil.isAuthenticated()) {
+        ModalController.close()
+      }
     } catch (error) {
       const isUserRejectedRequestError =
         error instanceof AppKitError &&
--- a/packages/scaffold-ui/src/partials/w3m-connecting-wc-browser/index.ts
+++ b/packages/scaffold-ui/src/partials/w3m-connecting-wc-browser/index.ts
@@ -6,6 +6,7 @@ import {
   ConnectorController,
   EventsController,
   ModalController,
+  OptionsController,
   RouterController,
   SIWXUtil
 } from '@reown/appkit-controllers'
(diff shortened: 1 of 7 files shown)


packages/scaffold-ui/src/views/w3m-connecting-farcaster-view/index.ts
```diff
@@ -209,11 +210,13 @@ export class W3mConnectingFarcasterView extends LitElement {
           })
         }
         this.loading = false
-        if (hasConnections && isMultiWalletEnabled) {
-          RouterController.replace('ProfileWallets')
-          SnackController.showSuccess('New Wallet Added')
-        } else {
-          ModalController.close()
+        if (await SIWXUtil.isAuthenticated()) {
```
0xmkh: Can we check if SIWX is configured first before doing `await` call?

packages/scaffold-ui/src/partials/w3m-connecting-wc-browser/index.ts
```diff
@@ -53,7 +54,9 @@ export class W3mConnectingWcBrowser extends W3mConnectingWidget {
         throw new Error('w3m-connecting-wc-browser: No connector found')
       }
 
-      ModalController.close()
+      if (await SIWXUtil.isAuthenticated()) {
```
0xmkh: Same here

packages/scaffold-ui/src/views/w3m-connecting-farcaster-view/index.ts
santgr11: SIWXUtil.isAuthenticated() already returns true internally when SIWX isn't configured, but that still means unwrapping a promise/microtask for nothing in the non-SIWX case. Added an explicit !OptionsController.state.siwx || (await SIWXUtil.isAuthenticated()) guard here so we skip the await entirely when SIWX isn't set up, same pattern as w3m-email-verify-otp-view. Updated the affected tests to mock siwx as configured where they were asserting the "not yet authenticated" case.

0xmkh (approved):

PR #5766, opened 2026-09-06
fix(onramp): include countryCode and source params in Meld URL

### LINKED ISSUE #5752: [bug] Onramp: Meld URL omits countryCode, showing a fabricated "You receive" amount on load

Open the Meld widget with the exact parameter set AppKit builds in `OnRampController.setSelectedProvider()`:

```
https://meldcrypto.com/?publicKey=<MELD_PUBLIC_KEY>&destinationCurrencyCode=USDC&walletAddress=<ANY_EVM_ADDRESS>&externalCustomerId=<PROJECT_ID>
```

Substitute `MELD_PUBLIC_KEY` from `packages/controllers/src/utils/ConstantsUtil.ts`, any EVM address, and any project id. Or reach it in-app: connect a wallet on mainnet and open the `OnRampProviders` view, then pick Meld.

### Steps to Reproduce

1. Open the on-ramp and select Meld.
2. Look at "You receive" before touching anything.

**Expected:** a real quote (~93-95 USDC for $100).
**Actual:** `0.001201 USDC`, attributed to Binance Connect.

Entering or changing the fiat amount by hand replaces it with a correct quote (`237.77 USDC` for $250). So the wrong figure is only ever the *initial* render.

### Summary

Two separate problems, one of which is fixable in AppKit.

**1. The number is a static placeholder, not a mispriced quote.** Loading the widget with different parameters returns the identical value every time:

| URL params | "You receive" on load |
|---|---|
| `USDC`, `sourceAmount=100` | `0.001202 USDC` |
| `USDC`, `sourceAmount=500` | `0.001202 USDC` |
| `BTC`, `sourceAmount=100` | `0.001202 BTC` |
| no destination currency | `0.001202 BTC` |

Same number regardless of amount or asset — a skeleton value that is never replaced, because no quote request fires on load. This part is Meld's widget behaviour.

**2. AppKit makes it worse by omitting `countryCode`.** `setSelectedProvider()` appends only:

```
publicKey · destinationCurrencyCode · walletAddress · externalCustomerId
```

With no `countryCode`, Meld selects **Binance Connect**, which cannot serve US customers at all — Meld's own quote API returns `NO_VALID_QUOTES` for `BINANCECONNECT`+`US`, while `US` returns Stripe and Swapped. Adding `countryCode=US` to the same URL switches the provider and the fabricated figure never appears:

```
no countryCode  ->  "Binance Connect · 0.001202 USDC"
countryCode=US  ->  "Swapped · See more quotes"        (no bogus figure)
```

`sourceAmount` is also never sent. Meld documents it as defaulting to 100, but supplying it explicitly does not by itself trigger a quote — it populates the display only.

### Suggested fix

Append `countryCode` (and ideally `sourceAmount` / `sourceCurrencyCode`) to the Meld URL in `setSelectedProvider()`. Country could come from the existing geo/locale signal, or be exposed as an on-ramp option so integrators can set it.

This is the same underlying complaint as #5394, which was closed as a feature request. The repro above shows it isn't cosmetic: users are shown a concrete "you receive" amount that is wrong by five orders of magnitude, next to an enabled "Buy" button.

`@reown/appkit` 1.8.17, `@reown/appkit-adapter-wagmi` 1.8.17 (Nuxt 3 / Vue)

### Node.js Version

18.19.0

### Package Manager

yarn

devtechedge:
## Summary

Selecting Meld in the on-ramp builds a widget URL with only `publicKey`, `destinationCurrencyCode`, `walletAddress`, and `externalCustomerId`. Without `countryCode`, Meld routes to Binance Connect and paints a fabricated "You receive" figure (around `0.0012`) on first load. Adding `countryCode=US` to that same URL switches providers and removes the bogus amount.

This change makes AppKit send `countryCode` (plus `sourceCurrencyCode` / `sourceAmount` when known) so the widget opens on a provider that can actually quote.

## What changed

- Resolve a country code in this order: explicit `OnRampController.setCountryCode()`, browser locale region via `Intl` / `navigator.language`, then `US` when the payment currency is `USD`.
- Append `countryCode`, `sourceCurrencyCode` (from `paymentCurrency.id`), and `sourceAmount` (from `paymentAmount` when set) to the Meld URL inside `setSelectedProvider()`.
- Add `countryCode` to on-ramp state, clear it in `resetState()`, and cover the URL builder with unit tests including an explicit GB override.

## Why

Meld's initial render is sensitive to country. Omitting it leaves US users on Binance Connect with a skeleton receive amount next to an enabled Buy button. Passing country (and source currency/amount) is the AppKit-side fix called out in #5752; the remaining placeholder behavior is on Meld's widget.

## How tested

- Updated `OnRampController` unit tests for the Meld URL shape with `countryCode`, `sourceCurrencyCode`, and `sourceAmount`.
- Added a case that prefers an explicit `setCountryCode('GB')` over locale defaults.
- Could not run the full package suite in this API-only workflow. Changes are localized to the controller and its existing vitest file.

Fixes #5752

### DIFF
--- a/.changeset/onramp-meld-country-code.md
+++ b/.changeset/onramp-meld-country-code.md
@@ -1,0 +1,5 @@
+---
+"@reown/appkit-controllers": patch
+---
+
+Fix Meld on-ramp widget URL missing `countryCode`, `sourceCurrencyCode`, and `sourceAmount`, which left users on Binance Connect with a fabricated receive amount.
--- a/packages/controllers/src/controllers/OnRampController.ts
+++ b/packages/controllers/src/controllers/OnRampController.ts
@@ -6,6 +6,7 @@ import { ConstantsUtil } from '@reown/appkit-common'
 
 import { MELD_PUBLIC_KEY, ONRAMP_PROVIDERS } from '../utils/ConstantsUtil.js'
 import type { PaymentCurrency, PurchaseCurrency } from '../utils/TypeUtil.js'
+import { CoreHelperUtil } from '../utils/CoreHelperUtil.js'
 import { withErrorBoundary } from '../utils/withErrorBoundary.js'
 import { ApiController } from './ApiController.js'
 import { BlockchainApiController } from './BlockchainApiController.js'
@@ -34,6 +35,11 @@ export interface OnRampControllerState {
   providers: OnRampProvider[]
   error: string | null
   quotesLoading: boolean
+  /**
+   * ISO 3166-1 alpha-2 country code forwarded to Meld.
+   * When unset, AppKit falls back to the browser locale region, then US for USD.
+   */
+  countryCode?: string
 }
 
 type StateKey = keyof OnRampControllerState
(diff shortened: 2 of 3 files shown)


devtechedge:
Hi there, checking whether this one is still on your radar.

It passes countryCode and source through to the Meld onramp URL.

Open since 6 September with green checks 🔧 and no conflicts. Happy to revise as needed.

PR #5767, opened 2026-09-07
fix: use caipNetworkId in createNamespaces

### LINKED ISSUE #5552: [bug] `WCNamespaceUtil.createNamespaces` uses the `id` property to build the CAIP network ID instead of using the network’s `caipNetworkId`

https://stackblitz.com/edit/github-erhvie8k

### Steps to Reproduce

1 - Initialize AppKit using a custom network where the `id` differs from the reference used in `caipNetworkId`.

2 - Open the `connect` modal.

3 - Observe that the generated optional namespace contains a chain built using the network `id`, instead of the expected `caipNetworkId`.

### Summary

Just for clarification, I am adding support for the Neo3 and Stellar blockchains by building custom AppKit adapters and network definitions.

Both blockchains follow the same network ID conventions already used by their respective WalletConnect SDK implementations.

References:

- Stellar WalletConnect SDK: https://github.com/Creit-Tech/Stellar-Wallets-Kit/blob/main/src/sdk/modules/wallet-connect.module.ts

 - Neo3 WalletConnect SDK: https://github.com/CityOfZion/wallet-connect-sdk/blob/main/packages/wallet-connect-sdk-core/src/index.ts

Currently, `WCNamespaceUtil.createNamespaces` uses the `id` to build the CAIP network ID which could cause some inconsistencies when you need to use two networks with the same network id.

For example, in the syncAccounts method of AppKitBaseClient, the find operation relies on the id to locate the corresponding caipNetwork. When two networks share the same id, this lookup may return the wrong network, resulting in incorrect account synchronization and other inconsistencies.

Reference: 

https://github.com/reown-com/appkit/blob/1c5eb503dfdceafa08c71937e9a8984f2ffb3c91/packages/appkit/src/client/appkit-base-client.ts#L1606

I cannot assign a unique id to each network as a workaround, because WCNamespaceUtil.createNamespaces uses the network id to construct the caipNetworkId.

Proposed Solutions:

Refactor all locations where the id property is used to construct the caipNetworkId, ensuring that the existing caipNetworkId property is used directly instead of being derived from id.

Refactor all lookup operations that rely on the id property (e.g., find operations) to use caipNetworkId as the primary identifier for network resolution.

@reown/appkit
@reown/appkit-controllers

### Node.js Version

v22.14.0

### Package Manager

npm@11.8.0

devtechedge:
# Description

`WCNamespaceUtil.createNamespaces` rebuilt each WalletConnect chain id as `${chainNamespace}:${id}`. Custom adapters (Stellar, Neo3) set `id` to a local alias that is not the CAIP reference, so the optional namespace advertised the wrong chain (`stellar:stellar` instead of `stellar:pubnet`). The same happens for a Solana network defined with `id: 'solana'`.

This uses `chain.caipNetworkId` for `namespace.chains`. `rpcMap` still keys on `id` (WalletConnect EVM expects `1` / `137`). The deprecated-Solana switch and `syncAccounts` `find()`-by-`id` are left alone.

## Type of change

# Associated Issues

Fixes #5552

# Checklist


devtechedge:
Hi there - gentle check-in on this one whenever you have a spare moment.

It fixes `createNamespaces` so custom adapters advertise the real CAIP chain id (for example `stellar:pubnet` instead of `stellar:stellar`) by reading `caipNetworkId`, while leaving `rpcMap` keyed on `id` for WalletConnect EVM.

The change is small: `WCNamespaceUtil` plus unit coverage and a changeset, targeting `Fixes #5552`, with CI looking clean on my side.

Happy to rebase, retarget, or tweak the Solana / Stellar edge cases if that would make review easier.

No rush at all - just wanted to keep it on the radar since it has been open for a couple of weeks and is still ready from my end.

PR #5773, opened 2026-09-11
fix(ui): keep sign modal buttons inside the modal boundary

santgr11:
# Description

On the Sign modal (SIWE/SIWX "sign this message" screen), the Cancel and Sign buttons could
render outside the modal's boundary when both showed their longer loading labels
("Cancelling..." / "Signing...") at once, reachable when a network disrupts a sign flow
problem, since neither button is disabled while the other is loading.

## Root cause

`wui-button` has two independently sized boxes: its own custom-element `:host`, and the native
`<button>` inside its shadow DOM. Nothing let either one shrink below its natural content size as
a flex item, and nothing clipped or truncated the label if it didn't fit. Once the combined
content (label + spinner) of both buttons needed more width than the row had, the excess spilled
past the button and past the modal instead of adapting.

## Fix

`packages/ui/src/composites/wui-button/styles.ts`, CSS-only, no template or behavior changes:

- `:host { min-width: 0; }`: lets `wui-button`'s outer box shrink below its natural content width
  as a flex item, instead of refusing to shrink at all.
- `button { width: 100%; }`: keeps the inner native `<button>` always sized to match its own host,
  whatever that host's width ends up being. Without this, a button whose host had shrunk (via the
  rule above) would render its inner `<button>` at its own independent, unshrunk content size,
  overflowing past the now-smaller host, which is exactly what happened in the Swap modal's
  "Select token" button in an earlier version of this fix (see "Why this lives in `wui-button`,
  not in the Sign modal's views" below).
- `wui-text { min-width: 0; overflow: hidden; text-overflow: ellipsis; }`: truncates the label
  with `…` once the button is compressed, instead of overflowing.
- `wui-loading-spinner, slot[name='iconLeft'], slot[name='iconRight'] { flex-shrink: 0; }`: keeps
  the spinner/icons from being squeezed or distorted; only the text truncates.

Together, these guarantee any `wui-button` stays within its container's boundary regardless of
label length or how many buttons in a row are squeezed at once. Buttons with enough room are
unaffected: no visual change, since every rule above only engages once space is genuinely
insufficient. `fullWidth` buttons (like the Sign modal's Cancel/Sign) are also unaffected by the
`button { width: 100% }` line specifically, since `--local-width` was already `100%` for them
before this fix.

## Why this lives in `wui-button`, not in the Sign modal's two views

This fix could instead live entirely in `w3m-siwx-sign-message-view` and `w3m-connecting-siwe-view`
(the only two views that actually have the bug from the ticket), touching zero shared code. That
was deliberately rejected, for two reasons:

1. **Half of it can't be scoped there even if we wanted to.** `wui-button`'s `:host { min-width: 0 }`
   *can* be set from a caller's own stylesheet (`wui-button` is a normal light-DOM node from the
   caller's side), but the label-truncation piece (`wui-text { overflow: hidden; ... }`) lives
   inside `wui-button`'s own shadow DOM and is unreachable from outside without `wui-button`
   exposing a CSS Shadow Part. We built and verified a fully working version of that (`part="button"`
   on the inner `<button>`, `::part(button)` overrides in each view's own styles) specifically to
   check whether "modal-only" was actually achievable: it is, but at the cost of touching 4 files
   instead of 1, adding two new `styles.ts` files to views that had none, and duplicating the same
   mechanism per view instead of fixing it once. Any future modal with this same shape (a
   non-`fullWidth` button squeezed in a flex row) would need to remember to opt in by hand, rather
   than getting the fix automatically.
2. **This is already the established pattern in this codebase**, not a new architecture: `wui-cta-button`
   and `wui-card-select` already do the identical "ancestor sets `overflow`/`text-overflow`,
   `wui-text`'s slot forwards it via `inherit`" trick from inside their own component styles, and
   the original version of this fix was already reviewed and merged with the equivalent reasoning
   (see PR #5758's review discussion).

The tradeoff we picked: a two-line, always-safe default in the one shared component that already
owns this exact "does my content fit?" concern, over a wider-but-scoped, opt-in mechanism spread
across every caller that happens to need it.

## Type of change

# Associated Issues

Closes FS-160

# Checklist


0xmkh:
@santgr11 just a small thing. Can we have "Select token" text not crop out of swap modal?

**Before**
image

**After**
image

santgr11:
> @santgr11 just a small thing. Can we have "Select token" text not crop out of swap modal?
> 
> **Before** image
> 
> **After** image

Fixed. Can you take another look?

